PR #14, opened 2018-01-09
5 fix average combiner

Maximophone:
The engine base docker image goes from openjdk (alpine) to deeplearning4j/dl4j-devkit (centOS)

**RUN apk update && apk add curl** has been removed


cliveseldon:
Looks good

PR #51, opened 2018-01-18
Dockerize the entier wrapping process of building sklearn_iris example

errordeveloper:


errordeveloper:
There are probably more docs changes that would be needed, but I'm just trying to demonstrate how simple this becomes.

ahousley:
Thanks @errordeveloper! @gsunner is reviewing.

PR #60, opened 2018-01-22
change validation to handle no method check

### LINKED ISSUE #59: Remove the cluster manager functionality that checks for "type" presence in predictive units graph
The actual behaviour should be:

```
If implementation is specified, ignore the rest
if not, implementation defaults to UNKNOWN_IMPLEMENTATION and
   if type is specified ignore the rest
   if not, type defaults to UNKNOWN_TYPE and
      if methods is not specified, raise an error (we are in the case when none of implementation, type, methods has been specified)
```

cliveseldon:
fixes #59


Maximophone:
Logic is correct, tests pass, merging

PR #68, opened 2018-01-24
57 wrapping auto docs

Maximophone:
Related to issue #57 

To test: 
- build core-python-wrapper
- use core-python-wrapper to wrap an example microservice
- build the wrapped microservice
- The build folder should contain a readme.md file and docker inspect <image_name> should contain a list of labels reflecting the argument that were used when wrapping


Maximophone:
Fixed a bug where image name would default to None (should default to model name in lowercase)

cliveseldon:
should we add --rm to docker command in docs to ensure exited container is removed after wrapping , e.g.
```
docker run --rm -v /home/clive/work/seldon/seldon2.0/seldon-core-fork/examples/models/keras_mnist:/my_model seldonio/core-python-wrapper:0.5 /my_model MnistClassifier 0.1 seldonio
```

PR #72, opened 2018-01-24
add engine resources to proto and modify cluster manager

cliveseldon:
fixes #72


cluster-manager/src/main/java/io/seldon/clustermanager/k8s/SeldonDeploymentOperatorImpl.java
```diff
@@ -125,7 +125,11 @@ private static String getEngineEnvVarJson(Message protoMessage) throws SeldonDep
 						.addCommand("-c")
 						.addCommand("curl 127.0.0.1:"+clusterManagerProperites.getEngineContainerPort()+"/pause && /bin/sleep 5"))));
 
-			
+		// Add engine resources if specified
+		if (predictorDef.hasEngineResources())
+		    cBuilder.setResources(predictorDef.getEngineResources());
+		else // set default resource requests for cpu
+		    cBuilder.setResources(V1.ResourceRequirements.newBuilder().putRequests("cpu", Quantity.newBuilder().setString("0.1").build()));
```
Maximophone: the 0.1 default should be a variable in a config file somewhere else

Maximophone:
Tested on minikube, will merge

PR #125, opened 2018-03-26
Anonymous usage metrics collection

gsunner:


cliveseldon:
We also need a doc explaining the usgae stats, spartakus etc

PR #138, opened 2018-04-11
Java wrappers

### LINKED ISSUE #137: Create wrappers for Java based models
To cover
 * H2O
 * DL4j
 * Spark (enterprise standalone exported models)

cliveseldon:
Fixes #137 
 Docs in ./docs/wrappers/java.md
 H2O MOJO example ./examples/models/h20_mojo


gsunner:
The proto dir for java wrappers needs to exist before compile. Can be done like the other core components with an empty file.
seldon-core/wrappers/s2i/java/wrapper/src/main/proto/.keep
(Or create the proto dir when its needed)

PR #177, opened 2018-06-28
fix crd.libsonnet error

fisache:
Error when apply crd.libonnet of seldon-core through ksonnet

```
ks registry add seldon-core github.com/SeldonIO/seldon-core/tree/master/seldon-core && \
ks pkg install seldon-core/seldon-core@master && \
ks generate seldon-core seldon-core --withApife=false --namespace=seldon --withRbac=true
ks apply default
```

> ERROR find objects:
vendor/seldon-core/seldon-core/crd.libsonnet:258:2-3 Did not expect: (",", ",")
},

### DIFF
--- a/seldon-core/seldon-core/crd.libsonnet
+++ b/seldon-core/seldon-core/crd.libsonnet
@@ -255,4 +255,4 @@ local podTemplateValidation = import 'json/pod-template-spec-validation.json';
     },
 },
 
-},
+}


cliveseldon:
Which version of ksonnet are you using? Maybe this is a change in a newer version that does not allow this?

fisache:
I'm using v0.11.0 and currently latest version.

```
$ ks version
ksonnet version: 0.11.0
jsonnet version: v0.10.0
```

cliveseldon:
Thanks for the fix!

Issue #216, opened 2018-09-08
Nodejs wrapper for javascript models


SachinVarghese:
With the release of Tensorflow js with nodeJS bindings, the number of developers using javascript for creating and deploying models has certainly increased. Creating a nodeJS wrapper for these models with a REST and gRPC endpoints would certainly attract a lot of new model developers to the Seldon project's language-agnostic way of deploying models.

cliveseldon:
This sounds great. We would be happy to help you work on it.
Would an [S2I](https://github.com/openshift/source-to-image) wrapper make sense?

cliveseldon:
There already exists some example of [S2I Node JS Builder images](https://github.com/sclorg/s2i-nodejs-container) which could be a base.
The key would be to create the microservice API server, which for example is Flask in python.

SachinVarghese:
Yes, sure I can start creating an S2I wrapper for the same. Can someone share more details on how can I contribute to the project? I'll start by creating a builder image for a model component.

cliveseldon:
I suggest you look at the python S2I wrapper code which can be found at: https://github.com/SeldonIO/seldon-core/tree/master/wrappers/s2i/python
Create something similar with a similar structure, testing and image name.

SachinVarghese:
I have forked the Seldon-core repository and created an S2I wrapper code similar to python. I have tried to keep the idea similar as much as possible. Although it is limited to REST service only as of now, it is tested with an example tensorflow model. Please find the link at [seldon-core-s2i-nodejs ](https://github.com/SachinVarghese/seldon-core/tree/issue-216/wrappers/s2i/nodejs). 

@cliveseldon can you review and please let me know if I shall proceed with a pull request.

cliveseldon:
This looks great. Feel free to do a Pull Request.
It would be great to add some docs in the docs folder following the pattern of the other wrappers available. Also, to add one example in example/models folder that can be referenced from the top level readme.

SachinVarghese:
As per your suggestion, I have added a [nodejs tensorflow model](https://github.com/SachinVarghese/seldon-core/tree/issue-216/examples/models/nodejs_tensorflow) in the examples section and also added a doc for the [nodejs wrapper](https://github.com/SachinVarghese/seldon-core/blob/issue-216/docs/wrappers/nodejs.md).

[...]

PR #224, opened 2018-09-17
GRPC API for javascript models with Nodejs s2i wrapper

SachinVarghese:
Creating a pull request for the implementation of GRPC API in the Nodejs s2i wrapper for JavaScript models as per issue #216.

- Updated the node js microservice and its requirements for the grpc api

-  Updated the test model-template-app encironment for the API_TYPE

- Created  a grpc_client.js to test the model-template-app API


cliveseldon:
Suggest you create a 0.2 of the nodejs wrapper that has the grpc js components.

SachinVarghese:
> Suggest you create a 0.2 of the nodejs wrapper that has the grpc js components.

Are you suggesting to create a separate wrapper at `wrappers/s2i/` folder or upgrade the current one as nodejs:0.2 because it supports both REST and GRPC now?

Also the user doesn't need to create grpc components, only a grpc client to connect with the server created by the wrapper. I have tested it with the current `test/model-template-app` using the example grpc_client that is created. Let me know your thoughts.

cliveseldon:
Sorry my mistake.

 Ok, so we should still eventually have this as 0.2, as we can't change 0.1 now its officially published.
If you could update the version to 0.2-SNAPSHOT while we test.

If would be great to have update the mnist example to test using gRPC. You can look at https://github.com/SeldonIO/seldon-core/blob/master/examples/models/sklearn_iris/sklearn_iris.ipynb as an example that tests both REST and gRPC.

SachinVarghese:
@cliveseldon  Also can you point me a model example in which the send_feedback API is implemented? I saw the python implementation but need more understanding about how the feedback ties back to other Seldon components like router and Seldon analytics. Any kind of documentation should be apposite.

cliveseldon:
A router example would the the [e-greedy MAB](https://github.com/SeldonIO/seldon-core/blob/master/notebooks/epsilon_greedy_gcp.ipynb). The general internal API is defined [here](https://github.com/SeldonIO/seldon-core/blob/master/docs/reference/internal-api.md).

Yes, we do need docs for helping people building routers. @jklaise

SachinVarghese:
@cliveseldon Thanks for sharing the links. My doubt was regarding the [send feedback API](https://github.com/SeldonIO/seldon-core/blob/master/wrappers/python/model_microservice.py#L74) implemented as a part of the model microservice in the python flask app. It seems that this is another function definition that a model developer could have in order to tap into the feedback cycle of the deployment using a router because it is not mentioned in the [internal API doc](https://github.com/SeldonIO/seldon-core/blob/master/docs/reference/internal-api.md#model) for the model component.

I think we can review the models' GRPC API implementation for now and next I shall take up the send feedback implementation for models as a separate request along with a router implementation. Let me know your thoughts.

cliveseldon:
I think the "model" wrapper should implement the sendFeedback method as it is an optional method for a Model. We need documentation for this but the core code is:
https://github.com/SeldonIO/seldon-core/blob/0d589a9dd4ed1b355ae208d9f61d709b61566cb5/engine/src/main/java/io/seldon/engine/predictors/PredictorConfigBean.java#L42-L71

So by default a model just implements transform-input which is converted to a /predict call for MODELs. But it can optionally implement send-feedback if it wishes. This would be key to reinforcement learning. 

If its easy to add I would. More importantly a full Wrapper needs to cover all Predictive Unit types so more importantly the NodeJS code needs to handle Routers, Transformers, Combiners etc to be a fully fledged Wrapper.

cliveseldon:
The sendFeedback return is actually ignored at present so its only informational for the components:

https://github.com/SeldonIO/seldon-core/blob/0d589a9dd4ed1b355ae208d9f61d709b61566cb5/engine/src/main/java/io/seldon/engine/predictors/PredictiveUnitBean.java#L126-L128

SachinVarghese:
Hi @cliveseldon , I have created an implementation of the GRPC SendFeedback API for the model component as per the [python implementation](https://github.com/SeldonIO/seldon-core/blob/master/wrappers/python/model_microservice.py#L112) which adds the GRPC service using the [Model Service](https://github.com/SeldonIO/seldon-core/blob/master/wrappers/python/model_microservice.py#L135) but to accomodate the Send Feedback API it should have used the [Seldon Service](https://github.com/SeldonIO/seldon-core/blob/master/proto/prediction.proto#L106) as mentioned in the prediction.proto, Please clarify.

Also, I have created a Transformer Predictive Unit for the nodejs wrapper and have added it to the wrapper test. Please review this implementation.

cliveseldon:
Looks good. Small thing. The notebooks should use 0.2-SNAPSHOT throughout to ensure everything still works.

SachinVarghese:
> Looks good. Small thing. The notebooks should use 0.2-SNAPSHOT throughout to ensure everything still works.

This is done. The notebook examples are using the new 0.2-SNAPSHOT image now.

SachinVarghese:
@cliveseldon Please let me know if any more changes are required. Also, you may review and close this pull request.

cliveseldon:
One thing I saw was that the Docker container running the model does not respond to SIGTERM signals. If possible it would be good to change this so it shutdowns on receiving a signal.

PR #250, opened 2018-10-08
Engine java opts annotations and ambassador timeout annotation

cliveseldon:
* Allow engine java_opts via annotation ```seldon.io/engine-java-opts```
 * Ensure timeout annotations are added to Ambassador svc configuration


codebreach:
are the builds autodeployed somewhere?

do we have a `DOCKER_IMAGE_TAG:COMMIT_HASH` type situation?

PR #366, opened 2019-01-06
Mistyped check causing NULL Pointer Exceptions with getNamespace function

ro7m:
Mistyped check causing NULL Pointer Exceptions with getNamespace function in watcher


ro7m:
@cliveseldon

cliveseldon:
Thanks for this. Good catch. Will merge and create new images. We should add some test coverage in a new PR.

PR #373, opened 2019-01-08
Adding test case for SeldonDeploymentWatcher

ro7m:
Adding test case for SeldonDeploymentWatcher.
@cliveseldon


ro7m:
@cliveseldon  i choose something which does not match the SeldonDeployment to mimic a failure scenario

cliveseldon:
I think the issue is actually this test is failing: testModifiedSeldonMLDeploymentsForMultiNamespaceCluster

ro7m:
fixed that

cliveseldon:
Great. Thanks for the contribution!

PR #395, opened 2019-01-18
Outlier update

arnaudvl:
Update VAE outlier detector in line with Seq2Seq detector and add outlier detection README.


arnaudvl:
> Looks good to me. Are we keeping the binary model weight files in version control because they take long to train for the case studies (same with seq2-seq but not with isolation forest)?

Training is fast for the VAE, but a bit slower for seq2seq. I kept the binary model weight files there so the demo can be run without having to train the model first.

Issue #258, opened 2018-10-17
Update Ambassdor Helm or remove and use Ambassador's helm chart


cliveseldon:
We should remove Ambassador from our helm chart and use the official Ambassaor Helm chart unless there is a blocker for things we need. If there is a blocker we should do a PR to Ambassador to update their Helm chart if possible.

We also need to update Ambassador version as there is a possible [security vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2018-0500) in their use of the curl library which is fixed in 0.40.0

ryandawsonuk:
We are now using ambassador chart. We hit a slight hitch with new ambassador's grpc support that led to https://github.com/SeldonIO/seldon-core/pull/510 but we are ported to the helm chart

PR #445, opened 2019-02-12
Make ambassador a dependency

naseemkullah:
Tackles  #258 

Please review!
Thanks


naseemkullah:
Good catch, this would indeed be redundant since https://github.com/helm/charts/blob/master/stable/ambassador/templates/rbac.yaml would be generated with rbac enabled.

Mind you, it is cluster wide and not configurable to be namespace scoped. (e.g. ClusterRole not Role)

naseemkullah:
Same thing for Datawire's chart FWIW https://github.com/datawire/ambassador/blob/master/helm/ambassador/templates/rbac.yaml

cliveseldon:
> Good catch, this would indeed be redundant since https://github.com/helm/charts/blob/master/stable/ambassador/templates/rbac.yaml would be generated with rbac enabled.
> 
> Mind you, it is cluster wide and not configurable to be namespace scoped. (e.g. ClusterRole not Role)

Ahh.. that is an issue. We need to allow DevOps to deploy with just Roles if namespaced and ClusterRoles if clusterwide. This allows fine grained permissions for users. Maybe this was the main blocker for not using the Helm/Datawire chart... ?

naseemkullah:
I suppose it is!
We should create a proposal upstream for this.

We can take the grafana chart as an example where they do this:

https://github.com/helm/charts/blob/4df52d5628e776d943ba59cabf975907302317a8/stable/grafana/values.yaml#L5

cliveseldon:
Yes - we should get a PR for the upstream change. Are you willing to do that and link to this PR?

naseemkullah:
Sure!

naseemkullah:
https://github.com/helm/charts/pull/11354 created.

naseemkullah:
I've made the required changed pre-emptively assuming the upstream PR will go through (removed ambassador rbac related objects, added `rbac.namespaced` value)

naseemkullah:
Alright we are good to with regards to `rbac`. The PR was merged upstream and it is now configurable with the `ambassador.rbac.namespaced` value.

Something worth noting is that we still have the `single_namespace` value as well.

naseemkullah:
bump

ryandawsonuk:
One knock-on effect of this is that the labels on ambassador become different. This is used in [notebooks which match using "service=ambassador"](https://github.com/SeldonIO/seldon-core/search?l=Jupyter+Notebook&q=%22service%3Dambassador%22). Using the [ambassador chart we get a standard `app.kubernetes.io` label](https://github.com/helm/charts/blob/master/stable/ambassador/templates/deployment.yaml#L10) so it becomes "app.kubernetes.io/name=ambassador"

ryandawsonuk:
Am also noticing that when I run through the helm_examples notebook with this change the grpc call for the a/b test fails with

```
_Rendezvous: <_Rendezvous of RPC that terminated with:
    status = StatusCode.UNIMPLEMENTED
    details = ""
    debug_error_string = "{"created":"@1553618148.143629760","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1039,"grpc_message":"","grpc_status":12}"
```
The REST call works. Both calls work when using master branch.

cliveseldon:
worth checking : https://docs.seldon.io/projects/seldon-core/en/latest/workflow/troubleshooting.html#i-get-404s-when-calling-the-ambassador-endpoint

ryandawsonuk:
Turns out the "StatusCode.UNIMPLEMENTED" is a known issue not related to this PR and happens intermittently on GRPC requests, especially the first one. Will raise that separately.

ryandawsonuk:
I'm working on getting the E2E tests to pass for these changes in a [seldon-core-ambassador-pr445](https://github.com/SeldonIO/seldon-core/tree/seldon-core-ambassador-pr445). Currently hit a problem with multi-armed bandit test where the 'mymab-abtest' pod never full comes up:

mymab-abtest-never-becomes-ready

Not seeing this problem on master and have merged master into that branch. Not sure how it can be related to ambassador though.

ryandawsonuk:
When stepping through the notebook the `mymab-abtest-eg-router-seldonio-mab-epsilon-greedy-1-1` service for eg-router does get created and the seldon-container-engine pod then does become ready.

ryandawsonuk:
So the above gist shows 13 failures and [now down to 7](https://gist.github.com/ryandawsonuk/94e17c0d28ce241d5b2b94f87949b580). A lot of grpc call failures resolved in part through adding resource to minikube and specify resources on ambassador (this dealt with pod restarts I was seeing) and also by adding retries to deal with https://github.com/SeldonIO/seldon-core/issues/473 But now it seems there are http-related ones. This is strange as don't seem to see those when running particular tests in isolation.

Also need to map the current use of `--set single_namespace=false` to [what the official chart supports](https://github.com/helm/charts/blob/master/stable/ambassador/values.yaml#L24).

ryandawsonuk:
I think the problem might be that the labels now differ for a helm vs ksonnet deployment and the code to port-forward is common and matches on labels

ryandawsonuk:
Have created a branch with the above changes and changing the default for ambassador to cluster wide as a workaround for the namespace var issue. The workaround can be reversed if the helm/charts PR gets merged.

PR #446, opened 2019-02-12
Update redis

naseemkullah:


ryandawsonuk:
Will merge this but will then move to redis helm chart shortly after

PR #447, opened 2019-02-12
Standardize values

naseemkullah:
This builds on top of PR #445 but did not want to put too much in 1 PR.

After it gets approved, please consider this PR.

It follows more standardized helm practices which makes the chart more familiar to users of stable helm chart.


cliveseldon:
We need to do some further checks on this as a few dependent projects for our Marketplace release depend on the core chart for [GCP marketplace](https://github.com/SeldonIO/seldon-core-launcher) and [AWS Marketplace](https://github.com/SeldonIO/seldon-core-aws)

naseemkullah:
Sounds good, thanks!

ryandawsonuk:
We also need to look at all the [references in the seldon-core repo to these properties via `--set`](https://github.com/SeldonIO/seldon-core/search?q=%22--set%22&unscoped_q=%22--set%22)

cliveseldon:
I am closing this as we have changed our helm install so seldon-core chart is no longer used and will be removed.

PR #457, opened 2019-02-25
WIP: Python wrappers rewrite

cliveseldon:
Rewrite of python wrapper

 * Code is for python 3.X only
 * Adds type annotations
 * Consolidates REST and gRPC processing so all requests are parsed for correctness
 * Creates a SeldonClient reference library for call the Seldon API
 * Updates tester programs to use new SeldonClient
 * Extends tests


python/tests/resources/contract.json
```diff
@@ -0,0 +1,39 @@
+{
+    "features":[
+	{
+	    "name":"sepal_length",
+	    "dtype":"FLOAT",
+	    "ftype":"continuous",
```
oshev: would be good to have a test for categorical variables as well

python/seldon_core/microservice_tester.py
```diff
@@ -0,0 +1,243 @@
...
+
+        if args.prnt:
+            print("RECEIVED RESPONSE:")
+            print(response)
+            print()
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("contract", type=str,
+                        help="File that contains the data contract")
```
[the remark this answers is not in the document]
jklaise: That would be very useful!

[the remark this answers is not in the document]
oshev: Great! How it's better proceed with it? Should I create a PR or just pass the code to you? I need to clean up a bit. Will be ready on Monday.

python/seldon_core/metrics.py
```diff
@@ -1,28 +1,79 @@
...
-from seldon_core.microservice import SeldonMicroserviceException
+from typing import List, Dict
 
 COUNTER = "COUNTER"
 GAUGE = "GAUGE"
 TIMER = "TIMER"
 
 
 def create_counter(key, value):
+    """
```
oshev: missing type annotation

python/seldon_core/microservice.py
```diff
@@ -17,10 +20,18 @@
 DEBUG_PARAMETER = "SELDON_DEBUG"
 DEBUG = False
 
-ANNOTATIONS_FILE = "/etc/podinfo/annotations"
-ANNOTATION_GRPC_MAX_MSG_SIZE = 'seldon.io/grpc-max-message-size'
 
 def startServers(target1, target2):
+    """
```
oshev: missing type annotation

python/seldon_core/microservice.py
```diff
@@ -30,24 +41,17 @@ def startServers(target1, target2):
...
-        if status_code is not None:
-            self.status_code = status_code
-        self.payload = payload
-        self.reason = reason
-
-    def to_dict(self):
-        rv = {"status": {"status": 1, "info": self.message,
-                         "code": -1, "reason": self.reason}}
-        return rv
+def parse_parameters(parameters):
```
oshev: type annotation

python/seldon_core/persistence.py
```diff
@@ -35,15 +44,24 @@ def restore(user_class, parameters):
         return pickle.loads(saved_state_binary)
 
 
-def persist(user_object, push_frequency=None, debug=False):
+def persist(user_object, push_frequency=None):
+    """
+    Start a thread to oersist a user class to Redis
```
oshev: typo

python/seldon_core/persistence.py
```diff
@@ -35,15 +44,24 @@ def restore(user_class, parameters):
         return pickle.loads(saved_state_binary)
 
 
-def persist(user_object, push_frequency=None, debug=False):
+def persist(user_object, push_frequency=None):
```
oshev: type annotation

oshev:
I can't see this branch when I do `git branch --all` so I can't pull it and observe it locally (I did `git fetch` and even tried to checkout the repo from scratch). Any ideas why does this happen?

oshev:
Sorry, I didn't realise Clive was merging from his fork of seldon-core. Never mind my question. Clear now.

python/seldon_core/microservice_tester.py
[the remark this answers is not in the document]
cliveseldon: A PR would be great.

oshev:
Shouldn't all API testing functionality go into a separate package inside `seldon-core` package? Otherwise, you're mixing it with production functionality.

python/seldon_core/microservice_tester.py
```diff
@@ -0,0 +1,252 @@
...
+            payload_type = "tensor"
+        else:
+            payload_type = "ndarray"
+
+        response_predict = sc.microservice(data=batch, transport="rest",payload_type=payload_type,method="predict")
+        response_feedback = sc.microservice_feedback(prediction_request=response_predict.request,prediction_response=response_predict.response,reward=1.0,transport=transport)
+        if args.prnt:
+            print("RECEIVED RESPONSE:")
+            print(response_feedback)
+            print()
```
oshev: or just:
`print(f"RECEIVED RESPONSE:\n{response_feedback}\n")`

python/seldon_core/microservice_tester.py
```diff
@@ -0,0 +1,252 @@
...
+    contract = json.load(open(args.contract, 'r'))
+    contract = unfold_contract(contract)
+    endpoint = args.host + ":" + str(args.port)
+    sc = SeldonClient(microservice_endpoint=endpoint)
+
+    for i in range(args.n_requests):
+        batch = generate_batch(contract, args.batch_size, 'features')
+        if args.prnt:
+            print('-' * 40)
+            print("SENDING NEW REQUEST:")
```
oshev: or just:
`print(f"{'-' * 40}\nSENDING NEW REQUEST:")`

python/seldon_core/seldon_client.py
```diff
@@ -0,0 +1,1394 @@
...
+class SeldonClient(object):
+    """
+    A reference Seldon API Client
+    """
+
+    def __init__(self, gateway: str = "ambassador", transport: str = "rest", namespace: str = None, deployment_name: str = None,
+                 payload_type: str = "tensor", oauth_key: str = None, oauth_secret: str = None,
+                 seldon_rest_endpoint: str = "localhost:8002", seldon_grpc_endpoint: str = "localhost:8004",
+                 ambassador_endpoint: str = "localhost:8003", microservice_endpoint: str = "localhost:5000",
+                 grpc_max_send_message_length: int =4 * 1024 * 1024, grpc_max_receive_message_length: int =4 * 1024 * 1024):
```
oshev: Not for this PR but maybe it's worth to refactor the client to inject in it 1) authentication method (instead of addressing each type of authentication with "ifs"), 2) endpoint type (REST or GRPC).

This might help to standardise functionality or different types, drastically reduce the number of "ifs" and number of parameters.

oshev: Another option could be to create a parent SeldonClient and inherit 2 classes: SeldonClientGRPC and SeldonClientRest and put them in 3 separate modules. Currently, `seldon_client.py` is huge with lots of similar bits.

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+
+    """
+    if hasattr(user_model, "tags"):
+        return user_model.tags()
+    else:
+        return {}
+
+
+def client_class_names(user_model: object, predictions: np.ndarray) -> List[str]:
+    '''
```
oshev: should be """ as above

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+       Optional payload
+
+    """
+    if hasattr(user_model, "send_feedback"):
+        return user_model.send_feedback(features, feature_names, reward, truth, routing=routing)
+
+
+def client_route(user_model: object, features: Union[np.ndarray,str,bytes], feature_names: List) -> int:
+    """
+    Gte routing from user model
```
oshev: typo

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+    ----------
+    user_model
+       A Seldon user model
+    features_list
+       A list of payloads
+    feature_names_list
+       Column names for payloads
+    Returns
+    -------
+       An aggrehated payload
```
oshev: typo

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+    Returns
+    -------
+       Class names
+    '''
+    if len(predictions.shape) > 1:
+        if hasattr(user_model, "class_names"):
+            return user_model.class_names
+        else:
+            n_targets = predictions.shape[1]
+            return ["t:{}".format(i) for i in range(n_targets)]
```
oshev: "t" for target in the output is not quite obvious unless you read the code, especially when the name of the function is `client_class_names`

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+    Parameters
+    ----------
+    user_model
+       A Seldon user model
+    features
+       Data payload
+    feature_names
+       Data payload couln names
+    kwargs
+       Optional keyowrd args
```
oshev: typo

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    Transform output
+    Parameters
+    ----------
+    user_model
+       A Seldon user model
+    features
+       Data payload
+    feature_names
+       Data payload couln names
```
oshev: typo

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+    feature_names_list
+       Column names for payloads
+    Returns
+    -------
+       An aggrehated payload
+    """
+    if hasattr(user_model, "aggregate"):
+        return user_model.aggregate(features_list, feature_names_list)
+    else:
+        return features_list[0]
```
oshev: Are you sure this is correct that you only use first elements in features list and ignore feature_names_list altogether in this case? Looks suspicious.

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+        if not validate_metrics(metrics):
+            jStr = json.dumps(metrics)
+            raise SeldonMicroserviceException(
+                "Bad metric created during request: " + jStr, reason="MICROSERVICE_BAD_METRIC")
+        return metrics
+    else:
+        return []
+
+
+def client_feature_names(user_model: object, original: List) -> List:
```
oshev: original: List[str]?

python/seldon_core/user_model.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Returns
+    -------
+       A list of custom metrics
+
+    """
+    if hasattr(user_model, "metrics"):
+        metrics = user_model.metrics()
+        if not validate_metrics(metrics):
+            jStr = json.dumps(metrics)
```
oshev: `j_str`

python/seldon_core/utils.py
```diff
@@ -0,0 +1,377 @@
...
+        return grpc_datadef_to_array(datadef)
+    elif data_type == "binData":
+        return request.binData
+    elif data_type == "strData":
+        return request.strData
+    else:
+        raise SeldonMicroserviceException("Unknown data in SeldonMessage")
+
+
+def grpc_datadef_to_array(datadef: prediction_pb2.DefaultData) -> np.ndarray:
```
oshev: Maybe worth to extract GRPC-related methods to a separate module `utils_grpc.py`.

python/seldon_core/microservice.py
oshev: also, the name according to PEP-8 should be `start_servers`

python/seldon_core/microservice.py
```diff
@@ -291,7 +117,7 @@ def main():
     parser.add_argument("--log-level", type=str, default="INFO")
     parser.add_argument("--tracing", nargs='?',
                         default=int(os.environ.get("TRACING", "0")), const=1, type=int)
-
+
```
oshev: Above, on lines 89-90:
```
89: value = parts[1]
90: value = parts[1][1:-1]
```
line 89 does nothing

python/seldon_core/microservice.py
```diff
@@ -316,21 +137,11 @@ def main():
 
     if args.persistence:
         logger.info('Restoring persisted component')
-        user_object = persistence.restore(user_class, parameters, debug=DEBUG)
+        user_object = persistence.restore(user_class, parameters)
```
oshev: Please, consider refactoring `main()` as currently it's 136 lines and contians specific bits which could be separate methods.

python/seldon_core/microservice_tester.py
```diff
@@ -0,0 +1,243 @@
...
+    args = parser.parse_args()
+
+    LOG_FORMAT = '%(asctime)s - %(name)s:%(funcName)s:%(lineno)s - %(levelname)s:  %(message)s'
+    if args.log_level == "DEBUG":
+        log_level = logging.DEBUG
+    elif args.log_level == "INFO":
+        log_level = logging.INFO
+    else:
+        log_level = logging.ERROR
+    logging.basicConfig(level=log_level, format=LOG_FORMAT)
```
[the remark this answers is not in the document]
cliveseldon: There is a command line arg to print requests and responses.

python/seldon_core/seldon_client.py
cliveseldon: Yes, agree. Maybe leave for a future PR.

python/seldon_core/utils.py
cliveseldon: future PR

python/seldon_core/seldon_methods.py
```diff
@@ -193,19 +197,20 @@ def aggregate(user_model: object, request: prediction_pb2.SeldonMessageList) ->
         request_json = json_format.MessageToJson(request)
         response_json = user_model.aggregate_rest(request_json)
         return json_to_seldon_message(response_json)
-    if hasattr(user_model, "aggregate_grpc"):
+    elif hasattr(user_model, "aggregate_grpc"):
```
oshev: are you sure this kind of checks will still work if you inherit from SeldonComponent? all attributes will be in the child class

cliveseldon: Was trying to get round some type check failures. Solved by allowing Any at certain parts  as the user is not required to sub class SeldonComponent at present. Will push an update.

oshev: I liked the idea of having to instantiate SeldonComponent but then - yes - you can't rely on `hasattr` methods and should use a different mechanism. I reckon you could make it an Interface with no methods, and then branch from it child interfaces with specific set of methods (my guess is that not every method would form a separate interface but you could group some). Users would then inherit several interfaces they need and you would check if their class is an instance of specific interfaces instead of using `hasattr`. This way you could keep strong typing. It sounds more Java, I know, but with type annotations that's where Python seems to be shifting a bit. :-) 

I realise that even if you agree, this a big change, probably requires a bit more thinking, and shouldn't go in this PR. I think using Any is a good solution for now. Not very elegant but works.

cliveseldon: Yes I think that all makes a lot of sense. We should add an issue when this is merged.

Issue #473, opened 2019-03-26
intermittent UNIMPLEMENTED error from ambassador


ryandawsonuk:
Spotted this in the helm_examples notebook but have also seen in the E2E tests. Was seeing:

```
_Rendezvous: <_Rendezvous of RPC that terminated with:
    status = StatusCode.UNIMPLEMENTED
    details = ""
    debug_error_string = "{"created":"@1553618148.143629760","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1039,"grpc_message":"","grpc_status":12}"
```
Might especially affect the A/B test scenario but not sure. Was encountered when testing https://github.com/SeldonIO/seldon-core/pull/445 but it isn't related to that change as has been seen separately.

ryandawsonuk:
Seems to especially affect first request so could be related to https://github.com/datawire/ambassador/issues/504
Also related to https://github.com/grpc/grpc/issues/16515 as 503 is unavailable and it seems this is the behaviour for a network connectivity failure

ryandawsonuk:
Seem to be able to resolve this for tests using https://github.com/SeldonIO/seldon-core/commit/656ac5b07193a54ce182d480c1750e4f16e89cd7#diff-3119fc203a07a67876b8d74d342faceeR44 - have only applied that in a particular branch for now

ryandawsonuk:
I think upgrading to latest ambassador (through https://github.com/SeldonIO/seldon-core/pull/480) also resolves this for the notebooks. I was seeing this intermittently in notebooks and haven't seen it in that branch with the latest ambassador

ryandawsonuk:
Turns out the issue is present in 0.53.1 but can be worked around by setting ambassador to run as root.

Will check again in 0.60.0 - https://github.com/datawire/ambassador/issues/504

ryandawsonuk:
Actually issue still present even when running as root. I managed to recreate the ambassador grpc problem from the helm_examples.ipynb. This time though I didn't uninstall and reinstall ambassador. I just followed the notebook and for the AB test scenario (at which point ambassador has been there for a while) I saw was the grpc calls reporting as failed for a good minute while REST requests worked. It seems the issue actually affects the initial period when a new instance comes online, specifically for A/B tests. Oddly, ambassador logs were reporting that the requests were going through successfully and there were no errors in other logs but the response was clearly a failure:

[...]

PR #480, opened 2019-03-29
make ambassador a dependency

ryandawsonuk:
Replaces https://github.com/SeldonIO/seldon-core/pull/445

The core of this is that change.

As discussed in that thread, there are also changes to improve the reliability of the E2E tests.

The chart doesn't currently allow us to [default ambassador to a single namespace](https://github.com/SeldonIO/seldon-core/pull/445/files#diff-b37aa1d151528f0e08d013ead8427f31L59) and override so have had to change a lot of notebooks. The last commit does this. Will reverse that if the PR to the stable/ambassador chart is merged. We could create a ticket dedicated for keeping an eye on that and reversing these changes.


.gitignore
```diff
@@ -174,3 +174,7 @@ doc/source/_static/cluster-manager
 python/build/
 python/dist/
 testing/scripts/proto
+testing/scripts/tensorflow
```
ryandawsonuk: Noticed these were generated

components/outlier-detection/isolation-forest/isolation_forest.ipynb
```diff
@@ -273,7 +273,8 @@
    "source": [
     "!helm install ../../../helm-charts/seldon-core --name seldon-core \\\n",
     "        --namespace seldon \\\n",
-    "        --set ambassador.enabled=true"
+    "        --set ambassador.enabled=true \\\n",
+    "        --set ambassador.env.AMBASSADOR_SINGLE_NAMESPACE=true"
```
ryandawsonuk: I think this is not strictly necessary as otherwise it will be cluster-wide and that would work too. But previously it was namespace-only.

testing/scripts/Makefile
```diff
@@ -6,10 +6,17 @@ build_protos:
 	python -m grpc.tools.protoc -I. --python_out=. --grpc_python_out=. ./proto/prediction.proto
 	cd ../../notebooks && make build_protos
 
+#requires mvn, helm, ks, python, s2i and an accessible k8s cluster (if minikube increase mem and cpu)
+#tests will create namespaces, deploy seldon core and create examples
```
ryandawsonuk: Pointers to help getting started with running E2E tests

testing/scripts/Makefile
```diff
@@ -6,10 +6,17 @@ build_protos:
...
 test: build_protos
 	pytest -s -W ignore > run.log 2>&1
 
 .PHONY: clean
 clean:
 	rm -f proto/prediction*
+	rm -rf run.log
+	rm -rf tensorflow
+	rm -rf my-model
+	rm -rf ../../wrappers/s2i/python/_python/
```
ryandawsonuk: These are generated files so I figured they could be included in the clean

testing/scripts/k8s_utils.py
```diff
@@ -3,6 +3,8 @@
 import subprocess
 import os
 import time
+import random
+from retrying import retry
```
ryandawsonuk: Use of retry is to handle sporadic connectivity failures. See https://github.com/SeldonIO/seldon-core/issues/473

components/outlier-detection/isolation-forest/isolation_forest.ipynb
ryandawsonuk: For a little while I wasn't able to see anything in the grafana dashboard but that turned out to be unrelated as I was seeing it with master too and it was addressed by recreating minikube.

Issue #491, opened 2019-04-04
use v1 ambassador api


ryandawsonuk:
v0 is deprecated - https://github.com/datawire/ambassador/commit/87f90e852c09d0ca983a80313235ac04ca0eb138#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR13

This means checking that uses in https://github.com/SeldonIO/seldon-core/blob/master/cluster-manager/src/main/java/io/seldon/clustermanager/k8s/SeldonDeploymentOperatorImpl.java#L528 such as timeout_ms, prefix and service conform to https://www.getambassador.io/reference/mappings/ or updating to the new API

PR #492, opened 2019-04-04
ambassador v1 api

ryandawsonuk:
fixes https://github.com/SeldonIO/seldon-core/issues/491

I checked against the specs referenced in that issue and did a step through of the timeouts notebook to check that feature works.

I also ran the E2E tests - I did [get a failure](https://gist.github.com/ryandawsonuk/b9feaa125cc3e7aca18a6ae65b735db5) but it looks like an instance of https://github.com/SeldonIO/seldon-core/issues/473 Surprised to see that even with the retry library there though


ryandawsonuk:
Actually when I run `pytest -s -W ignore test_helm_charts_single_namespace.py` it really does fail

ryandawsonuk:
With https://github.com/SeldonIO/seldon-core/pull/492/commits/561157c5d2e12aea482a945e7f94dba4217a6085 it now passes for me.

I was able to recreate the behaviour in a notebook, though it does not happen every time so was not easy to reproduce. Seems like it's not just that the first grpc request after deployment fails, it's that sometimes it can take minutes for the grpc to start working, even after rollout reports completion (which is longer than the retries were allowing for)

ryandawsonuk:
After discussion on the ambassador slack and the linked ambassador issue, have found we can get the grpc requests working in a timely fashion by running ambassador as root.

ryandawsonuk:
Turns out grpc requests are better but still intermittent problems - https://github.com/SeldonIO/seldon-core/pull/498

PR #494, opened 2019-04-05
downgrade ambassador version due to grpc reliability

ryandawsonuk:
We upgraded ambassador in https://github.com/SeldonIO/seldon-core/pull/480

But it has caused problems with grpc calls on ambassador startup, recorded in https://github.com/SeldonIO/seldon-core/pull/492

There was perhaps an intermittent problem already but it is much more serious in the new ambassador version (0.52.0) than it was in 0.40.2. To get the new version working we'd have to put long retry timeouts in our tests

### DIFF
--- a/helm-charts/seldon-core/values.yaml
+++ b/helm-charts/seldon-core/values.yaml
@@ -1,5 +1,7 @@
 ambassador:
   enabled: false
+  image:
+    tag: 0.40.2
   replicaCount: 1
   resources:
     limits:


ryandawsonuk:
Alternatively I've found that 0.50.0 works. (But 0.51.0 does not.)

ryandawsonuk:
0.50.3 seems good too and I've checked that against https://github.com/SeldonIO/seldon-core/pull/492

ryandawsonuk:
Now going to run as root to resolve instead of downgrading - https://github.com/SeldonIO/seldon-core/pull/492/commits/976c9e69aea9943112d7ad56f5256eaa1903a774

PR #500, opened 2019-04-09
Fix HPA Nullpointer

### LINKED ISSUE #499: SeldonDeployment keeps hanging
There is a problem with the new 2.7-snapshot version of cluster manager - when a seldondeployment is deployed, it correctly creates a deployment, a replica set and the pods. However then it fails, because it fails to get the namespacedHorizontalPodAutoscalers, it correctly logs the error (KubeCRDHandlerImpl->getOwnedHPAs:324) but then it returns null, which causes a silent NullPointerException in SeldonDeletionHandler:144 which kills the whole task without updating the state of the seldon deployment (whatever it is).

A solution is to catch the nullpointer (heavy handed) or to return an optional instead (imho better)

The symptom is that the seldondeployment forever hangs as Creating

cliveseldon:
Fixes #499


ahousley:
@cliveseldon I guess this will be part of the 0.2.7 release - please can you tag it to the appropriate project to confirm. Thanks!

PR #501, opened 2019-04-10
downgrade ambassador due to grpc unreliability

ryandawsonuk:
Wanted to avoid this downgrade but after more testing it is the safest choice for now -https://github.com/SeldonIO/seldon-core/issues/473#issuecomment-481631117

### DIFF
--- a/testing/scripts/test_bad_graphs.py
+++ b/testing/scripts/test_bad_graphs.py
@@ -7,7 +7,7 @@ from seldon_utils import *
 from k8s_utils import *
 
 def wait_for_status(name):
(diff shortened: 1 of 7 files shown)


testing/scripts/test_bad_graphs.py
```diff
@@ -7,7 +7,7 @@
 from k8s_utils import *
 
 def wait_for_status(name):
-    for attempts in range(3):
+    for attempts in range(5):
```
ryandawsonuk: This change is unrelated. I just noticed that it can occasionally take a bit longer for the status to appear.

Issue #495, opened 2019-04-08
default ambassador chart to single namespace


ryandawsonuk:
We use `--set ambassador.env.AMBASSADOR_SINGLE_NAMESPACE=true` in a lot of our examples. You can see this by searching the repo - it was introduced in https://github.com/SeldonIO/seldon-core/pull/480

At the time we couldn't default the ambassador helm chart to a single namespace. But now we can as https://github.com/helm/charts/pull/12670 has been merged

So we can upgrade to that version of the chart and set single namespace as the default. Then we can remove all those instances of setting the env var to true. Instead we'd need to set the new `scope.singleNamespace` to false for the cluster-wide cases

PR #509, opened 2019-04-11
default ambassador to singleNamespace

ryandawsonuk:
fixes https://github.com/SeldonIO/seldon-core/issues/495


ryandawsonuk:
When running the e2e to check this I hit the A/B test 503 issue again. Seems like it can still happen occasionally on this version and with this timeout. I then ran through the helm notebook again and it was fine. I also re-ran just the tests for which it failed and on the re-run they passed. I don't think it's related to these changes, just noting it for info.

Issue #505, opened 2019-04-11
ci-cd demo uses old jenkins and kubectl fails on newer k8s


ryandawsonuk:
The kubectl binary available to the jenkins image used in https://github.com/SeldonIO/seldon-core/tree/master/examples/cicd-argocd is old now. The result on a newer k8s now is that you get this error when the [job for building images](https://github.com/ryandawsonuk/cicd-demo-model-source-files/blob/master/Makefile.build-model-images-job#L11) tries to use kubectl:

```
error: SchemaError(io.k8s.api.core.v1.ConfigMapNodeConfigSource): invalid object doesn't have additional properties
make: *** [Makefile.build-model-images-job:8: do_build] Error 1
```

I saw this and added a line to print the `kubectl version`, which output:

```
/bin/kubectl version
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.0", GitCommit:"fc32d2f3698e36b93322a3465f63a14e9f0eaead", GitTreeState:"clean", BuildDate:"2018-03-26T16:55:54Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:45:25Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
```
So the issue is https://stackoverflow.com/a/55581050/9705485

I think we need to update the jenkins being used

Might have to make changes like https://github.com/SeldonIO/seldon-core/pull/506/files#diff-b0f64032faabe39ddbcfd9a55b142519R14 for jenkins (that one is argocd)

ryandawsonuk:
We could also try to add the github plugin via the chart to avoid installing it manually:

https://github.com/helm/charts/blob/master/stable/jenkins/values.yaml#L138

ryandawsonuk:
We may also be able to get the jenkins to use local identity store out of the box too. I think that's a [choice of realm](https://jenkins.io/doc/book/managing/security/) in jenkins terminology. See https://github.com/helm/charts/blob/master/stable/jenkins/values.yaml#L21 and https://github.com/jenkins-x/cloud-environments/blob/4010385ead7bc7a1816cb6d0fb540878dd888f78/env-minishift/myvalues.yaml#L56

[...]

PR #517, opened 2019-04-18
update argocd and jenkins in cd demo and script for minikube

ryandawsonuk:
for https://github.com/SeldonIO/seldon-core/issues/505

linked pr - https://github.com/SeldonIO/cicd-demo-model-source-files/pull/5


ryandawsonuk:
This includes changes to use the jenkins identity store straight away. This is needed as crumb doesn't seem to work with the default store.

I think we could simplify further by configuring the jobs from the beginning with https://github.com/helm/charts/tree/master/stable/jenkins#providing-jobs-xml But we could do that as a subsequent step. For the future I'd like to consider whether there's a way to avoid using the xml or avoid end users needing to fully understand it, rather like Jenkins-X has `jx import`.

PR #544, opened 2019-05-07
fix for travis failure

gsunner:


gsunner:
This is not needed, it was an issue with travis

PR #566, opened 2019-05-16
[DNM] Python release

jklaise:
Updated master following Python package release, some files had outdated image references, should all be fixed in this PR, also references to 0.7-SNAPSHOT where changed to 0.7.


jklaise:
Just noticed that references to the python2 s2i images have been updated as well, so that will cause breaking with old examples (in fact that is already the case as seldon-core-s2i-python2:0.6 doesn't exist). In general we should do the following:

- Rework any Python2 examples to Python3 so any references to Python2 can be
   - As a quick fix we can revert the python2 image references to the latest stable (0.5.1)
- Update the script renaming images in the repo to be more precise
   - Need to handle X-SNAPSHOT references differently

jklaise:
Closing in favour of #594 .

seldondev:
* failed to create agent
* failed to load trigger config for repository SeldonIO/seldon-core for ref a18dd5649719e891a58988056f864ac2217a376a
* failed to find any lighthouse configuration files in repo SeldonIO/seldon-core at sha a18dd5649719e891a58988056f864ac2217a376a
* failed to process repo SeldonIO/seldon-core refref a18dd5649719e891a58988056f864ac2217a376a
* no such file or directory

PR #573, opened 2019-05-16
updated openvino version to 2019.1 in python_openvino model wrapper

dtrawins:


dtrawins:
I added also a PR https://github.com/SeldonIO/seldon-core/pull/574 with the exemplary pipeline which includes the new model wrapper version with OpenVINO 2019.1.
@cliveseldon could you publish those 2 new docker images (the wrapper and exemplary model)?

Issue #562, opened 2019-05-14
Docker image name and version messed up


sasvaritoni:
https://github.com/SeldonIO/seldon-core/blob/3d8b3791b343118953757a1e787e5919cc64e697/engine/src/main/java/io/seldon/engine/predictors/PredictiveUnitState.java#L90-L95

This will end up really ugly when the docker registry is running on a non-default port (this is often the case for private docker registries).

E.g.:  "hostname:5000/myimage:v0.0.1"

The split should in fact occur at the last ":", not the first one.

cliveseldon:
Could you create a PR for this?

sasvaritoni:
Added PR #576

PR #576, opened 2019-05-17
Fix PredictiveUnitState image name and version (#562)

sasvaritoni:
Proposal for splitting the docker image name and version


sasvaritoni:
Resolves #562

Issue #580, opened 2019-05-22
Tensorflow MNIST Model example on EKS


gsunner:
Create an example based on the current "deep_mnist" but with instructions on how to use on AWS EKS cluster

axsauze:
This has been resolved with #581, #582 and #587

PR #581, opened 2019-05-22
AWS Elastic Kubernetes/Container Service Deep Mnist Example

axsauze:
# Deep Mnist in AWS EKS

This pull request contains the code for an example of the Deep Mnist tensorflow model, deployed into an EKS cluster. The full example can be found at https://github.com/SeldonIO/seldon-core/blob/aws_eks_deep_mnist/examples/models/aws_eks_deep_mnist/aws_eks_deep_mnist.ipynb

### This is related to #580.


axsauze:
Thanks for the heads up @cliveseldon, I've added the nblink file now

PR #582, opened 2019-05-22
Added missed s2i folder to Scikitlearn SpaCy Text Example

axsauze:
# Added missed s2i folder to Scikitlearn SpaCy Text Example
This PR contains the missing folder required for the configuration of the .s2i process.


axsauze:
Added the nblink file as per @cliveseldon's suggestion in #580

PR #587, opened 2019-05-24
Added missing link to Jupyter notebook

axsauze:


axsauze:
Thanks @cliveseldon you're correct- I've made the amendment so the SKLearn nblink is in that same directory and the link is present in the notebooks.rst

Issue #527, opened 2019-04-25
option to not set runAsUser for engine


ryandawsonuk:
Cluster manager currently [injects a securityContext for the engine container](https://github.com/SeldonIO/seldon-core/blob/master/cluster-manager/src/main/java/io/seldon/clustermanager/k8s/SeldonDeploymentOperatorImpl.java#L132). But this is actually set at pod level and effectively applies a restriction to the model itself. This effectively means the user has to choose for the cluster manager a user that will work for all the models... but the models are accessing different directories so they might need different permissions. 

Openshift can infer acceptable permissions (this was observed as causing awkwardness on openshift) if you don't set runAsUser so it would be good provide an option to not set the securityContext.

ryandawsonuk:
Sorry the above is incorrect. The runAsUser is being applied on to the engine container.

(FWIW we seem to have introduced this capability to support istio integration - https://github.com/SeldonIO/seldon-core/commit/318a66aac426857e504648f5db7a3ea6c15c7aa1)

ryandawsonuk:
Given that runAsUser is only being set by the cluster manager for the engine container, this doesn't seem to be a problem for cluster manager. So closing and will reopen if anyone reports otherwise.

ryandawsonuk:
Am reopening this as it would be convenient to not have to set a user. Otherwise when you install seldon-core you have to know the range of available users.

It doesn't affect the models but it does add complexity to deploying cluster-manager on clusters where the user range is restricted. Since [we only introduced this setting for istio integration](https://github.com/SeldonIO/seldon-core/commit/318a66aac426857e504648f5db7a3ea6c15c7aa1), it should be ok to make it optional.

ryandawsonuk:
Above PRs will provide a way to do this by setting `--set engine.securityContext.enabled=false`

PR #601, opened 2019-05-29
option to not set engine user

ryandawsonuk:
fixes https://github.com/SeldonIO/seldon-core/issues/527
depends on https://github.com/SeldonIO/seldon-operator/pull/8

Tested by deploying following [steps in helm notebook](https://github.com/SeldonIO/seldon-core/blob/master/notebooks/helm_examples.ipynb) and using `--set engine.securityContext.enabled=false` to not set any engine user when deploying seldon-core operator. After deploying model verify by checking podspec of model with `kubectl get pod <pod> -o yaml`.

Without overriding the flag the default is to set engine user as 8888


ryandawsonuk:
@cliveseldon oops, thanks... fixed that now

Issue #545, opened 2019-05-08
Investigate Request Logging


cliveseldon:
* Add to engine optional push of request/responses to external request logger
 * What would be best initial backends to support: InfluxDB, ElasticSearch?

jklaise:
For data science usage we also need to think about breaking down batch requests into single instance logging.

axsaucedo:
I guess we could close this @cliveseldon @jklaise ?

cliveseldon:
Fixed.

PR #610, opened 2019-05-31
WIP: initial EFK setup

ryandawsonuk:
Part of https://github.com/SeldonIO/seldon-core/issues/545

Able to run EFK and find log entries in kibana such as:

image

But need to consider how we'd (optionally) log the full request and how the content of the request would appear in kibana. Clearly we see full log and search by timestamp already. Would we need to be able to search by a feature value?


cliveseldon:
Cool. I think by feature might be difficult but we do have a "names" field in the "default_data" which can give columns names. The core would be to get the request back and be able to load it as a SeldonMessage for retransimisison

ryandawsonuk:
Initial attempt at request-response logging logs first SeldonMessage.toString and then full request body. Comes out like:
```
2019-05-31 15:04:15.151  INFO 1 --- [nio-8000-exec-2] i.s.e.api.rest.RestClientController      : {"data": {"names": ["f0", "f1"], "ndarray": [[0.12, 0.16]]}}
2019-05-31 15:04:15.157  INFO 1 --- [nio-8000-exec-2] i.s.e.api.rest.RestClientController      : {
  "meta": {
    "puid": "cecr3j2pl3mie72sohhbait7ti",
    "tags": {
    },
    "routing": {
    },
    "requestPath": {
      "classifier": "seldonio/mock_classifier:1.0"
    },
    "metrics": []
  },
  "data": {
    "names": ["proba"],
    "ndarray": [[0.05859816662313229]]
  }
}
```
Doesn't look great in kibana presently as it's multi-line and each line of logging is a separate entry. I think I need to figure out how to format it as single line and then figure out how to search for something more specific.

ryandawsonuk:
It seems it is [possible to get the logging to be parsed as structured](https://github.com/fluent/fluent-bit/issues/691), I think for that the log line might need to be entirely json and [single-line](https://github.com/kubernetes/kubernetes/issues/15795#issuecomment-149832661).

We could do the logging in a general way in the java engine using a [contextHandler](https://medium.com/@federicogaule/collecting-access-logs-into-elasticsearch-1a6f05288f8a) but we might be better doing it explicitly as then we can deserialize the grpc.

ryandawsonuk:
It looks better as single-line json but still not actually structured:

image

ryandawsonuk:
Actually looking at it in the tabular view, it seems it has been parsed:

image

ryandawsonuk:
Now able to filter on particular values from the request/response. Next step is to enrich the messages to indicate which is a request and which is a response. Going to do this in the PredictionService as there the request is enriched with puid and if we log that version then we'll be able to search for `meta.puid` to see linked request-response-feedback entries.

ryandawsonuk:
I tried moving the logging code to the PredictionService but then realised that we don't want to log the SeldonMessage itself - we want to log it after it goes through ProtoBufUtils.toJson (without this it doesn't seem to get read as json by elastic, although it does resemble json I think it has too many spaces and so on). So will see if I can get the puid include in the logging from RestClientController as that already does a ProtoBufUtils.toJson

ryandawsonuk:
Now have puid in request and response. 

Not sure about how we could log data without running up against [array querying limitations](https://discuss.elastic.co/t/query-array-by-position/124765).

We could look at doing [some custom parsing/transformation with fluentbit](https://docs.fluentbit.io/manual/parser/json) but this might be more complex than it's intended for. Other options might be if there's a simple way to just [serialize the message differently](https://stackoverflow.com/questions/28545401/java-json-protobuf-back-conversion) or send the request to another agent to transform (e.g. through a message broker).

ryandawsonuk:
Considering alternative idea of transforming data under https://github.com/SeldonIO/seldon-core/pull/616

ryandawsonuk:
Closing this in favour of https://github.com/SeldonIO/seldon-core/pull/616

seldondev:
* failed to create agent
* failed to load trigger config for repository SeldonIO/seldon-core for ref a9c51fbeb09fed3d6fa162b7a93f564ea149cae8
* failed to find any lighthouse configuration files in repo SeldonIO/seldon-core at sha a9c51fbeb09fed3d6fa162b7a93f564ea149cae8
* failed to process repo SeldonIO/seldon-core refref a9c51fbeb09fed3d6fa162b7a93f564ea149cae8
* no such file or directory

PR #616, opened 2019-06-04
EFK with transform for tabular data

ryandawsonuk:
Alternative to https://github.com/SeldonIO/seldon-core/pull/610

In this version the engine [can do a transform to put tabular data into key-value pairs](https://github.com/SeldonIO/seldon-core/pull/616/files#diff-0e71feb26156fe3f8429f348259802f3R123) for searching. We can then do e.g. value-range searches:

image

It checks whether there's a 'data.names' array before transforming but need to do more checks to make sure this won't break other types of requests (e.g. what can we assume about the data value type and array shape). Also need to consider performance.


ryandawsonuk:
For images we may be better to store the image on disk or in a bucket and keep a searchable reference to it. I'd like to investigate computing metadata about the image before storing it so that you could search by metadata (e.g. presence of certain colours) - see https://stackoverflow.com/questions/30440224/possible-to-store-images-in-elasticsearch

We should discuss whether to do this in the engine or in a wrapper library

ryandawsonuk:
Next steps:

- use thread to make async
- consider more examples - e.g. config option for column-primary
- look at binary, tensor and str cases
- Look at how parameterised - what env cars and what defaults and how to apply (think example helm charts)
- If single engine for graph is that ok that we won’t know which component? Will we know anyway? Do we need to?
- run on GCP
- document how it works
- consider KF-serving where engine not present - could it use same logic in python as part of its base image/s? Or should we publish a py lib?

ryandawsonuk:
New idea - look at performing the enrichment in a separate batch job that updates the entries in elasticsearch

ryandawsonuk:
depends on https://github.com/SeldonIO/seldon-operator/pull/13

ryandawsonuk:
Depends on https://github.com/SeldonIO/seldon-operator/pull/17

Issue #563, opened 2019-05-15
node exporter port conflict


csabika7:
Helm chart for seldon-core-analytics includes installing a node-exporter daemonset. As it is directly uses the host network reserving the port 9100.
Deployment of this helm chart would then fail if node-exporeter have been deployed already for another application like **ranch**.

Do you know some way to still be able to install seldon-core-analytics which would use the already deployed node-exporter if it is already present?

cliveseldon:
Would a fix me to make that port configurable in the Helm chart values?  Could you submit a PR?

csabika7:
Can you even do that?
In the Dockerfile of the node-exporter the 9100 port is exposed.
In the helm chart the hostNetwork is set to true.

So the container will use the host network namespace so setting any other port in the chart will not have any effect right?

gsunner:
merged changes in PR#617

PR #617, opened 2019-06-05
node exporter configurable port

csabika7:
Solution for issue: https://github.com/SeldonIO/seldon-core/issues/563


cliveseldon:
Thanks for this. @gsunner review from you and we can include in 0.3.0 release?

gsunner:
lgtm

Issue #383, opened 2019-01-11
Add an WSGI app server to Python REST wrappers


jklaise:
We should use e.g. Gunicorn to replace the Flask built-in server.

ryandawsonuk:
See https://ironboundsoftware.com/blog/2016/06/27/faster-flask-need-gunicorn/

Related is #453

marrrcin:
Have you considered using simpler WSGI server like Cheroot? It can be configured directly from the code.

PR #635, opened 2019-06-19
use gunicorn

ryandawsonuk:
First steps towards addressing https://github.com/SeldonIO/seldon-core/issues/383

See the link I posted on that issue. Seems to be common to use just flask on local machine and add gunicorn within container. See for example https://blog.miguelgrinberg.com/post/the-flask-mega-tutorial-part-xix-deployment-on-docker-containers

Have also done this in https://github.com/SeldonIO/seldon-core/blob/master/seldon-request-logger/Dockerfile

### DIFF
--- a/doc/source/python/README.md
+++ b/doc/source/python/README.md
@@ -7,3 +7,5 @@ To create a seldon Component in Python:
  1. [Wrap the component using S2I](python_wrapping_s2i.md) or [Docker](python_wrapping_docker.md).
 
 You can find various examples in our [example notebooks](../examples/notebooks.html).
+
+It also possible to mount a model into an existing image. This can be seen in the [seldon kubeflow examples](https://github.com/kubeflow/example-seldon)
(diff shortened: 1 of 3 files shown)


axsauze:
Just as a heads up it's worth taking into consideration that the Gunicorn server actually spins up multiple workers by default, which normally is indeed desired), howver given that we (at least currently) offload load balancing and replicas to the gateway components, we may have to avoid having more than one worker running. I do think that the benefits make it a great choice, and it's not too hard to set up 👍 we could even use this as an opportunity to discuss normalising the microservice vs seldonEngine APIs

cliveseldon:
We have considered gUnicorn as a standard way to wrap the Flask server we use for the python wrapper. This is also used by Sagemaker. Another option is to remove Flask and use Tornado.

ryandawsonuk:
My thinking is that gunicorn on k8s will help to get maximum use out of each container. The articles indicate that you can reliably handle a greater throughput of requests in a single container by adding gunicorn. So I’d expect that to carry over to multiple resource-constrained containers. To be sure we could do new benchmarking https://docs.seldon.io/projects/seldon-core/en/latest/reference/benchmarking.html

My impression is it doesn’t seem to negatively affect cold-start. 

We could instead just add a note to the docs saying that gunicorn isn’t included by default and how to add it if desired.

ryandawsonuk:
Closing this in favour of https://github.com/SeldonIO/seldon-core/pull/684

Issue #619, opened 2019-06-05
 Tensorflow GPU Example


axsaucedo:
Several community members and enterprise clients have requested examples using Seldon Core with GPU usecases. We should create an example that shows a step by step example building a tensorflow GPU model, and serving it with Seldon Core.

axsauze:
Once this example is up and running we could try to use #602 if relevant.

axsauze:
This will also address discussions like #590.

cliveseldon:
Done

Issue #602, opened 2019-05-29
tensorflow-gpu


ryandawsonuk:
The seldon-core microservice library has tensorflow (cpu version) [as a dependency](https://github.com/SeldonIO/seldon-core/blob/master/python/setup.py#L22). This leads to either dependency issues if anyone wants to use tensorflow-gpu or being able to build and then getting a 'CPU supports instructions that this TensorFlow binary was not compiled to use' message. This has been[ encountered before](https://github.com/SeldonIO/seldon-core/issues/380) and came up again recently [on slack](https://seldondev.slack.com/archives/C8Y9A8G0Y/p1559139617052000?thread_ts=1559060738.035900&cid=C8Y9A8G0Y).

We should look at ways to make it easier to use tensorflow-gpu with seldon. Some ideas:

- Will it work to install seldon-core, uninstall tensorflow and then install tensorflow-gpu? Slack discussion currently suggests not but if we did get that working would it be adequate to document that?
- Could we just rely on the tensorflow protos (which is what we really need) without bringing in all of tensorflow?
- Could we use another [library to handle detection](https://pypi.org/project/tensorflow-auto-detect/)?
- Provide two versions of seldon-core?

cliveseldon:
The seldon-core python libraries just need to convert a tensorProto to ndarray but so far not found a way to isolate that code easily and maintainably.

cliveseldon:
Still hard to see any easy solution to this without creating separating out the tensorflow part.

cliveseldon:
Fixed - you can specify tensorflow-gpu in python wrapper setup

Issue #590, opened 2019-05-26
GPU support with SERVICE_TYPE Model


muma378:
Hi, I was trying to deploy a `SeldonDeployment` to the cluster, which asks for gpu resource and CUDA. I writes the .yaml as the [official doc](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/) suggests, however the deployment was blocked at "parsing" the CRD  stage, which results in no deployment or service was created. It was totally OK to deploy models without gpu required.

I didn't find any example on using gpus, so, my question is: does Seldon-core support GPU? Or does anyone has succeeded in deploying a model with gpu required?

This is a part from my .yaml:
```
  predictors:
  - annotations:
      predictor_version: v1
    componentSpecs:
    - spec:
        containers:
        - image: xxx
          resources:
            limits:
              alpha.kubernetes.io/nvidia-gpu: 1
          imagePullPolicy: IfNotPresent
          name: xx
          volumeMounts:
          - mountPath: /usr/local/nvidia/bin
            name: bin
          - mountPath: /usr/lib/nvidia
            name: lib
        imagePullSecrets:
        - name: regcred
        terminationGracePeriodSeconds: 1
        volumes:
          - hostPath:
              path: /usr/lib/nvidia-384/bin
            name: bin
          - hostPath:
              path: /usr/lib/nvidia-384
            name: lib
    graph:
      children: []
      endpoint:
        type: GRPC
      name: xx
      type: MODEL
    name: xx
    replicas: 1
```

cliveseldon:
Can you provide more details on the error at "parsing" stage? 
Are you using the "master" branch or an earlier version of Seldon Core?

muma378:
> Can you provide more details on the error at "parsing" stage?
> Are you using the "master" branch or an earlier version of Seldon Core?

No, the version for seldon-core and seldon-core-crd both are 0.2.5. Installed with helm locally.

The seldon-core-apiserver reports an error  below:
```
2019-05-26 15:41:08.623  INFO 1 --- [pool-3-thread-1] io.seldon.apife.k8s.DeploymentWatcher    : The time is now 15:41:08
2019-05-26 15:41:08.623  INFO 1 --- [pool-3-thread-1] io.seldon.apife.k8s.DeploymentWatcher    : Watching with rs 3980232
2019-05-26 15:41:08.682  INFO 1 --- [pool-3-thread-1] io.seldon.apife.k8s.DeploymentWatcher    : ADDED
[...]

PR #638, opened 2019-06-20
Gpu tensorflow example

JoelH96:
Developed GPU Deep Mnist Tensorflow Example to run on GKE. The main sections are as follows:

1. Install of all CUDA and cuDNN drivers required for TF-GPU=1.13.1
2. Training and local testing of Deep Mnist model (this was achieved using a GCE VM with GPU).
3. Pushing the Deep-Mnist-GPU image to Google Container Registry.
4. Deploying the image using Seldon on GKE (with a GPU configured node)

ISSUES involving GPU support: #590 #602 #619


divyadilip91:
@cliveseldon @JoelH96 @ryandawsonuk Hi all ,
I am currently facing an issue of gpu not being found by my docker container when deploying a gpu specific seldon deployment on kubernetes .
I have seldon core 1.1.0 installed on kubernetes. The tensorflow-gpu python package that has been used in my docker image is 1.14 and the base docker image used is ufoym/deepo:pytorch-cu101-ver200129. 
 I receive the following errors
image
image
image
image


Would seldon deployments with gpu work only by using tensorflow-gpu 1.13 ? Will it not work for 1.14?
Could you please guide me on how to set up the docker image using gpu that can be used by seldon deployment yaml for deploying on kubernetes.
According to the above example notebook, its show a lot of dependencies to be installed. Is a base image already present that involves all these dependencies installed? Expecting your reply.

axsaucedo:
@divyadilip91 we've resolved this issue via #2048, it seems the image had a bug so we've depricated this tf GPU image, and we're providing a plain GPU conda image in 1.2.2 which resolves this issue (cc. @RafalSkolasinski)

RafalSkolasinski:
I think Alejandro meant that in 1.3 we will provide plain GPU image, see https://github.com/SeldonIO/seldon-core/issues/1789.
Right now there is still `tensorflow` there but idea is that one will be able to install whichever 
version of `tensorflow` they see fit.

@divyadilip91 It looks like you are not using our image as your base.
Could you try ours image https://hub.docker.com/repository/docker/seldonio/seldon-core-s2i-python3-tf-gpu
especially the one with latest tag `seldon-core-s2i-python3-tf-gpu:1.2.2-dev`.

RafalSkolasinski:
@divyadilip91 If you find an issue with our image, could you open a new issue for it please?

divyadilip91:
@axsaucedo Thanks for your immediate reply. 
@RafalSkolasinski I have seldon core 1.1.0 installed on my kuberenetes and I wrap my model using seldon python class and then create the dockerfile which has the seldon core command.So I havent used this as my base image. So will the seldon-core python package 1.2 create issues if the seldon core installed on kuberenetes is 1.1.0.Moreover would this image cause issue since I see the image mentioned above has 1.2 version and my kuberenetes seldon has 1.1.0.
Kindly help

RafalSkolasinski:
@divyadilip91 this will be only a version python wrapper inside the image. It should work properly with Seldon Core 1.1 in the cluster (this actually describes a version of the seldon core operator that is installed in the cluster).

You can use `seldon-core-s2i-python3-tf-gpu:1.2.2-dev` as the base image in your `Dockerfile` - you are not limited to using the s2i tool.

Also, you can then in your Dockerfile `pip install` any version of `seldon-core` and `tensorflow`.

divyadilip91:
Ok Thankyou @RafalSkolasinski . I ll use this as my base image and will get back in case I face any issues.

PR #639, opened 2019-06-20
Add Kustomize

cliveseldon:
* Scripts to autogenerate CRD yaml with OpenAPI validation
  * update CRD to yaml version
  * remove ksonnet util folder


cliveseldon:
The kubeflow kustomize integration has a few kubeflow specific components : https://github.com/kubeflow/manifests/pull/36
We could us a similar set of steps to create the kustomize from the helm chart.

PR #684, opened 2019-07-12
WIP: Update python wrapper to use gunicorn

### LINKED ISSUE #383: Add an WSGI app server to Python REST wrappers

### LINKED ISSUE #453: Question: Concurrency in Model REST server
I was wondering if you have any tests on how the REST API for `MODEL` type deployments handle concurrency?
I'm guessing this is the part of the code that handles it, so I would expect the Tornado coroutines to do something: https://github.com/SeldonIO/seldon-core/blob/master/python/seldon_core/model_microservice.py#L228-L262

However when testing with `ab` to throw some concurrent traffic at the API I see very linear scaling of the processing-time no matter if the model is extremely simple (a simple one layer model going from 100x1) or if it's more complex (Embeddings, RNNs, DotProducts).

Example:
Note: This is currently while running both the docker server and `ab` on the same machine, so I don't except everything to fly off the moon, but I expected some concurrency on a 4core machine with hyperthreading.
The problem seems to be identical when running the server with the CLI `seldon-core-microservice Model REST`, although with less network overhead.

Concurrency 1
```
ab -p model_input -T "application/x-www-form-urlencoded" -c 1 -n 1000 http://0.0.0.0:5000/predict
This is ApacheBench, Version 2.3 <$Revision: 1826891 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 0.0.0.0 (be patient)
Completed 100 requests
Completed 200 requests
Completed 300 requests
Completed 400 requests
Completed 500 requests
Completed 600 requests
Completed 700 requests
Completed 800 requests
Completed 900 requests
Completed 1000 requests
Finished 1000 requests


Server Software:        Werkzeug/0.14.1
Server Hostname:        0.0.0.0
Server Port:            5000

Document Path:          /predict
Document Length:        167 bytes

Concurrency Level:      1
Time taken for tests:   28.145 seconds
Complete requests:      1000
Failed requests:        0
Total transferred:      346000 bytes
Total body sent:        235000
HTML transferred:       167000 bytes
Requests per second:    35.53 [#/sec] (mean)
Time per request:       28.145 [ms] (mean)
Time per request:       28.145 [ms] (mean, across all concurrent requests)
Transfer rate:          12.01 [Kbytes/sec] received
                        8.15 kb/s sent
                        20.16 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       1
Processing:    17   28  36.0     26    1039
Waiting:       17   28  36.0     26    1038
Total:         17   28  36.0     26    1039

Percentage of the requests served within a certain time (ms)
  50%     26
  66%     28
  75%     29
  80%     29
  90%     31
  95%     32
  98%     36
  99%     39
 100%   1039 (longest request)
```

Concurrency 5
```
ab -p model_input -T "application/x-www-form-urlencoded" -c 5 -n 1000 http://0.0.0.0:5000/predict
This is ApacheBench, Version 2.3 <$Revision: 1826891 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 0.0.0.0 (be patient)
Completed 100 requests
Completed 200 requests
Completed 300 requests
Completed 400 requests
Completed 500 requests
Completed 600 requests
Completed 700 requests
Completed 800 requests
Completed 900 requests
Completed 1000 requests
Finished 1000 requests


Server Software:        Werkzeug/0.14.1
Server Hostname:        0.0.0.0
Server Port:            5000

Document Path:          /predict
Document Length:        167 bytes

Concurrency Level:      5
Time taken for tests:   27.359 seconds
Complete requests:      1000
Failed requests:        0
Total transferred:      346000 bytes
Total body sent:        235000
HTML transferred:       167000 bytes
Requests per second:    36.55 [#/sec] (mean)
Time per request:       136.797 [ms] (mean)
Time per request:       27.359 [ms] (mean, across all concurrent requests)
Transfer rate:          12.35 [Kbytes/sec] received
                        8.39 kb/s sent
                        20.74 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       0
Processing:    30  136  81.6    131    1207
Waiting:       29  122  72.2    118    1156
Total:         30  136  81.6    131    1207

Percentage of the requests served within a certain time (ms)
  50%    131
  66%    141
  75%    155
  80%    162
  90%    182
  95%    199
  98%    233
  99%    255
 100%   1207 (longest request)
```

Am I doing something wrong here, or is the REST api intended to scale on a pod-level, not at the thread level?

Or are you intending for gRPC to be used in this case?

Sorry for flooding your issues! :)

### LINKED ISSUE #674: Liveness probe kills Seldon engine container with long init waiting time (Python wrapper)
The Seldon engine containers get killed if the init function takes more than 20 seconds. This is the default behaviour as the init function of the python wrapper gets triggered after the flask server has been initialised and the pod has been marked as initialised. This is problematic if the container needs to perform init work that depends on parameters passed by the init function (through Seldon deploy params). An example of this is the PyTorch Hub integration which crashes if the model selected is large enough for the download to take longer than 20 seconds (https://github.com/SeldonIO/seldon-core/pull/642).

It is possible to avoid this issue if the init work is done before the wrapping class definition (I.e accessing the parameters directly from the env variables). The liveness probe doesn't get triggered in this case as the container is not marked as ready given the download executes synchronously before even reaching the wrapper definition. 

As we look to use reusable model servers as a more common design pattern, where the model weights or binaries are downloaded from an object store, we will need to think of an standard way to handle longer loading times during initialisation.

This could be tackled by making the parameters accessible as env variables by default, or alternatively by adding a PRE function that would execute before flask server initialisation.

cliveseldon:
* Adds a command line arg `--workers` defaults to 4 (only uses gunicorn if workers>1)
 * Adds an optional `load()` method to user_model prototype. This is called after init in each worker.
    * Needed for Tensorflow where loading your graph needs to be done in each process and not before gunicorn forks
 
 Fixes #453 
 Fixes #383
 Fixes #674


axsaucedo:
It woudl be worth adding to the flag to the s2i wrapper (the `run` file) so that the number of workers can be changed when deploying.

The reason why it is important to be able to change the number of workers is because there would be situations where you would only want to have 1 worker per pod - an example would be tasks that have heavy resource usage, such as memory.

If you have an ML graph that takes 50gb RAM for example, you may want your scaling to be horizontally, as if you request 50gb for the pods, as a user they may be confused that their pods keep going over their requests.

cliveseldon:
@axsaucedo You can set the env vairaible GUNICORN_WORKERS to limit the number of workers so don't think an s2i change is needed.

axsaucedo:
Ohh my bad, I had not seen the `os.environ.get("GUNICORN_WORKERS", "4")` - awesome! Thanks for the clarification @cliveseldon

alexlatchford:
I've been evaluating Seldon this week and this was on my list, nice job as I'm guessing this will land well within our time frame!

Not sure if this is the place to request this but would be great to include an option to choose the [worker type](http://docs.gunicorn.org/en/latest/install.html#async-workers), some of our models at inference time go out to feature store to enrich input requests. We'd be doing at as a part of an `input_transformer` component likely which we could set to just have a single worker and have it feed a `model` component with many times more workers.

PS. Apologies if there is already a different pattern for this!

alexlatchford:
Digging in some more, looking at the server implementations as they load the models in memory on the [first prediction](https://github.com/SeldonIO/seldon-core/blob/master/servers/sklearnserver/sklearnserver/SKLearnServer.py#L26). This would be done post process fork resulting in the model being stored in memory as many times are there are workers.

If you also expose the `preload_app` option for Gunicorn and set it to `True` if you've loaded the model during the initial phase (ie. not on the first network call) then it'll keep only one copy in memory and not duplicate the memory space but be accessible for all child processes. This is something we do here at Zillow to keep the memory footprint down in our current iteration of serving infrastructure but keeping a larger number of worker processes 😄 

This does have some downsides as it's incompatible with the `reload` option which enables hot-code reloading (useful for local development) but we just turn off `preload_app` locally to fix that.

axsaucedo:
Very interesting @alexlatchford, that's one of the challenges I outlined above, the preload_app functionality would certainly allow for the workers to only load one copy of the model reducing memory footprint. Of course there is still a challenge if the features extracted are big but that is a different challenge. I am not sure how much work may be involved to extend the wrapper to support this, definitely worth checking out, but depending on that we may have to open a separate issue to do this once the initial/basic move to gunicorn is landed.

examples/models/deep_mnist/DeepMnist.py
```diff
@@ -1,18 +1,26 @@
 import tensorflow as tf
 import numpy as np
+import os
 
 class DeepMnist(object):
     def __init__(self):
+        self.loaded = False
```
axsaucedo: Should this be set explicitly by the user? The other alternative could be to handle this without the user knowing, and basically assume loaded is true if, for example "load" returns, or even returns True.

examples/models/deep_mnist/DeepMnist.py
```diff
@@ -1,18 +1,26 @@
...
         self.class_names = ["class:{}".format(str(i)) for i in range(10)]
+        
+    def load(self):
+        print("Loading model",os.getpid())
```
axsaucedo: We could use log.debug as convention, not critical tho

doc/source/python/python_component.md
```diff
@@ -221,6 +221,72 @@ class UserCustomException(Exception):
...
+    def predict(self,X,feature_names):
+        if not self.loaded:
+            self.load()
+        predictions = self.sess.run(self.y,feed_dict={self.x:X})
+        return predictions.astype(np.float64)
+```
+
+### Gunicorn workers
+
+By default 4 gunicorn workers will be created for your model. If you wish to change this default then add the environment variable GUNICORN_WORKERS to the container for your model. An example is shown below:
```
lennon310: If you are using gunicorn workers and make this variable overridable,  you probably also need to consider (1) the `preload` option to reduce the memory usage, and (2) `max_requests` (probably along with `max_requests_jitter`) to restart the workers to avoid their memory leak. These are the opts in gunicorn command, but I guess there should be corresponding APIs that provide those options too.

Ref: http://docs.gunicorn.org/en/stable/settings.html

cliveseldon: I think we would need to make peload optional as not sure it will work for Tensorflow graphs that need a separate session in each fork.

PR #687, opened 2019-07-14
Alibaba Cloud Elastic Container Service (AKS) example

axsaucedo:
This PR contains an example to run Seldon in the Alibaba Cloud. It includes the basic requirements to leverage the Alibaba Cloud Elastic Container Service (AKS), together with the Alibaba container registry. There is also a Jupyter notebook in chinese version that was translated using Google translate and will be reviewed with the Alibaba team.

This all can be accessible through the [Readme](https://github.com/axsaucedo/seldon-core/tree/alibaba_example/examples/models/alibaba_aks_deep_mnist).


axsaucedo:
Removed Chinese-version of the page so this can be landed and a new PR can be opened with just this, so the conversation can start going with us supporting already Alibaba cloud from at least providing an example on how to use and deploy Seldon on it.

Issue #694, opened 2019-07-15
Deploying modified TensorRT example on GKE does not deploy anything


damitkwr:
Hi,

I am modifying the TensorRT example by getting rid of the TRTProxy Image and directly including a TRTServer pod in the deployment and a TYPE: MODEL deployment. Here is the relevant slack thread:

https://seldondev.slack.com/archives/C8Y9A8G0Y/p1563207453433900

damitkwr:
Update: The issue seems to be the liveness probe and readiness probe. Seldon was not recognizing this as seen by this: 

`{
 controller:  "seldondeployment-controller"
 error:  "Deployment.apps "trt-sp-sp-nvidia-72d9821" is invalid: [spec.template.spec.containers[1].livenessProbe: Required value: must specify a handler type, spec.template.spec.containers[1].readinessProbe: Required value: must specify a handler type]"
 level:  "error"
 logger:  "kubebuilder.controller"
 msg:  "Reconciler error"
 request:  "kubeflow/nvidia-sp"
 stacktrace:  "github.com/seldonio/seldon-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error
    /go/src/github.com/seldonio/seldon-operator/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/seldonio/seldon-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
    /go/src/github.com/seldonio/seldon-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:217
github.com/seldonio/seldon-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
    /go/src/github.com/seldonio/seldon-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158
github.com/seldonio/seldon-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
    /go/src/github.com/seldonio/seldon-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
github.com/seldonio/seldon-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil
    /go/src/github.com/seldonio/seldon-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134
[...]

Issue #693, opened 2019-07-15
Support JSON format in TFServer Proxy REST API


axsaucedo:
There has been demand to support the standard TF API REST JSON interface, as outlined on https://www.tensorflow.org/tfx/serving/api_rest

For this, the work would include extending the TFServing Proxy (which is now also a prepackaged model server) to ensure the REST endpoint is able to support JSON data type.

axsaucedo:
Functionality has been added via the new TFProxy image (seldonio/tfserving-proxy:0.4). To close this issue, we need to make sure this new image is being used and works properly in the TFProxy Prepackaged Model Server

axsaucedo:
As an update, the functionality is now added using image seldonio/tfserving-proxy:0.5

axsaucedo:
Fixed with #699

PR #699, opened 2019-07-16
Added TFServing JSON Support

axsaucedo:
This pull request contains the code that adds support for raw JSON to be sent to the TF Server, so it allows for the more flexible interface that TF Serving provides. This issue addresses #694

### DIFF
--- a/examples/models/tfserving-mnist/mnist_tfserving_deployment.json.template
+++ b/examples/models/tfserving-mnist/mnist_tfserving_deployment.json.template
@@ -25,7 +25,7 @@
 				    "--model_name=mnist-model",
 				    "--model_base_path=gs://seldon-tfserving-store/mnist-model"
 				],
-				"image": "gcr.io/kubeflow-images-public/tensorflow-serving-1.7:v20180604-0da89b8a",
+				"image": "tensorflow/serving:latest",
 				"name": "mnist-model",
 				"ports": [
 				    {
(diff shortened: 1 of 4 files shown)


axsaucedo:
This has been tested locally and works as expected, but when run in the kubernetes cluster the TFServing image fails to pull from the Google bucket with the following error:
```
2019-07-16 18:22:10.905196: I tensorflow_serving/model_servers/main.cc:154] Building single TensorFlow model file config:  model_name: mnist-model model_base_path: gs://seldon-mnist-model-bucket/mnist-model
2019-07-16 18:22:10.905412: I tensorflow_serving/model_servers/server_core.cc:444] Adding/updating models.
2019-07-16 18:22:10.905441: I tensorflow_serving/model_servers/server_core.cc:499]  (Re-)adding model: mnist-model
2019-07-16 18:22:11.776606: I external/org_tensorflow/tensorflow/core/platform/cloud/retrying_utils.cc:77] The operation failed and will be automatically retried in 1.45458 seconds (attempt 1 out of 10), caused by: Unavailable: Error executing an HTTP request (error code 6, error message 'Couldn't resolve host name')
        Performing request. Detailed error: Couldn't resolve host 'metadata'
```

axsaucedo:
@cliveseldon it seems that the container cannot reach the bucket. It's currently public, and it has been attempted with both the default bucket and a bucket created for this. Is there any pieces that I should be aware in regards to access for buckets? Are they only reachable from a GCloud cluster?

axsaucedo:
The issue was the TFServing image which was outdated and causing issues. It now has been updated to `tensorflow/serving:latest`, and the changes have been reflected in the Notebook JSON + Helm. The example now provides support for JSON.

axsaucedo:
This introduces the functionality required for #693

Issue #703, opened 2019-07-18
All numbers get converted to float when sending JSON resulting in unexpected behaviour in Proxies


axsaucedo:
When sending a JSON request such as `{ "jsonData": { "instances": [{ "length": 9 }] } }`, the number `9` would get converted into `9.0`. This becomes an issue in the TFProxy as the number datatypes would be changed, resulting in unexpected behaviour that differs from sending the request directly to the TFX server. 

This can be reproduced by running the following code:

```
from seldon_core.utils import seldon_message_to_json, json_to_seldon_message
seldon_message_to_json(
    json_to_seldon_message(
         { "jsonData": [ { "length": 9 } ] }
    )
)
```

Which outputs:

```
{'jsonData': [{'length': 9.0}]}
```

If you just run:
```
 json_to_seldon_message(
         { "jsonData": [ { "length": 9 } ] }
    )
```

You are able to see that the GRPC proto message converted is already changed to float:
```
list_value {
  values {
    struct_value {
      fields {
        key: "length"
        value {
          number_value: 9.0
        }
      }
    }
  }
}
```

PR #704, opened 2019-07-18
TFServing Enabled Text Response and Fixed JSON Parse

axsaucedo:
An issue arised when testing with NLP models that leverage more complex interfaces. This PR supports error responses from TFX server, and enhances the print with logs (which can be configured with env variable for log level. This also uncovered an issue with json objects being sent, which now is outlined in #703 - for now TFservers will have to be configured to accept Float instead of int32 for numbers.


axsaucedo:
@cliveseldon are the changes for this PR good to go?

Issue #458, opened 2019-02-27
The output from ambassador converts integer types to float .


nandagopal1992:
For instance the output of my s2i wrapped docker images looked like "{"data": "ndarray" : [{"values" : [0,1,2]"}]}" but when the same output comes from ambassador the response looks like "{"data": "ndarray" : [{"values" : [0.0,1.0,2.0]}]}"

axsaucedo:
This has now been fixed in #707 specifically with REST requests

PR #707, opened 2019-07-18
WIP: Proposed changes to ensure JSON REST Consistency (Fixes  #703)

### LINKED ISSUE #703: All numbers get converted to float when sending JSON resulting in unexpected behaviour in Proxies

axsaucedo:
# WIP: Proposed changes to ensure JSON REST Consistency (Fixes  #703)

This pull request contains the proposed changes to modify the microservice API such that for REST requests, it is ensured that the JSON requests stay intact as per the request in #703, as currently the google proto changes int representations into floats (i.e. 9 to 9.0), which modifies the original JSON in ways that then seem to affect potential proxies such like the TFServing proxy.

Once this is agreed, it would be applied across the rest of the APIs. It would also be necessary to assess that it wouldn't affect other wrappers, etc.

### DIFF
--- a/python/seldon_core/seldon_methods.py
+++ b/python/seldon_core/seldon_methods.py
@@ -3,12 +3,15 @@ from seldon_core.utils import *
 from seldon_core.user_model import *
 from google.protobuf import json_format
 from seldon_core.proto import prediction_pb2
-from typing import Any
+from typing import Any, Union, List, Dict
 
 logger = logging.getLogger(__name__)
 
 
-def predict(user_model: Any, request: prediction_pb2.SeldonMessage) -> prediction_pb2.SeldonMessage:
+def predict(
+            user_model: Any, 
+            request: Union[prediction_pb2.SeldonMessage, List, Dict]
+        ) -> prediction_pb2.SeldonMessage:
     """
     Call the user model to get a prediction and package the response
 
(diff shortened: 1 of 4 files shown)


axsaucedo:
@cliveseldon I've added some new tests on the funcitons added using the other tests as reference.

axsaucedo:
Confirmed all relevant tests run - as soon as @cliveseldon can confirm tests run locally should be possible to land (as local tests on GRPC fail due to configuration)

jklaise:
I can confirm that tests run locally.

axsaucedo:
Perfect, thanks @jklaise

axsaucedo:
Fixes  #458 as well

PR #732, opened 2019-07-27
endpoint /seldon.json not returning the file

klalit:
the file was present in location openapi/seldon.json. it is fixed now.


cliveseldon:
thanks!

Issue #733, opened 2019-07-27
Add the support to send image files directly using multipart/form-data instead of application/x-www-form-urlencoded


klalit:
This avoids the need to read and base64 encode the image prior to REST api call. It will be useful for models that take file/image as input for e.g.  model for  the image classification or ocr.  

A detailed discussion in the context

https://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data

microsoft also uses this MIME type for the images

https://westus.dev.cognitive.microsoft.com/docs/services/56f91f2d778daf23d8ec6739/operations/56f91f2e778daf14a499e1fe

klalit:
functionality implemented and merged

PR #735, opened 2019-07-28
add multipart/form-data functionality

klalit:
Implements #733 

All of the keys in SeldonMessage can be passed using "multipart/form-data". This gives the flexibility to load the data in "binData" or "strData" directly from file. The consumers don't have to read the data, do base64 encoding and pass it to the api. The http client will support this implicitly


gsunner:
This code adds the ability on the seldon python microservice to accept "multipart/form-data" which can be useful.

However there also needs to be work done in the engine to fully support this feature.
The rest controller of the engine at present accepts "application/json" and needs to be updated with ability to accept "multipart/form-data" in addition. Then it would also need to call the model microservice with a "multipart/form-data" call.

klalit:
please give me some pointers where can i start implementing the feature for engine. At present the functionality is implemented at the service level so it is an interface that can accept requests in different form. May be integration with engine can be taken as separate feature request? We are currently using seldon without engine.

klalit:
I guess this is where we want to include "multipart/form-data"
https://github.com/SeldonIO/seldon-core/blob/master/engine/src/main/java/io/seldon/engine/api/rest/RestClientController.java

cliveseldon:
Yes, that would be the place and it would need to somehow pass the request to the internal client predcitions : https://github.com/SeldonIO/seldon-core/blob/master/engine/src/main/java/io/seldon/engine/service/InternalPredictionService.java

klalit:
InternalPredictionService.java itself relies on SeldonMessage so have to follow similar logic as done in python service. Please check

klalit:
I've included the changes for engine. Please review

gsunner:
I saw that the original RestClientController method that handled the "application/json" as been refactored so have been testing this and it looks fine.

To test the new functionality that handles the "multipart/form-data" requests, it would be useful to have a simple example.

If you can take a simple example like "examples/models/sklearn_iris_jsondata" and create a similar one eg.

examples/models/sklearn_iris_multipart_form_data

This example should make use of "multipart/form-data" for the payload to to make the requests

klalit:
Thank you.. let me work on the example

klalit:
the example is added in seldon-core/examples/models/sklearn_iris_multipart_formdata
The "multipart/form-data" request to python service and engine returned the same result. now apis support the MIME type
Please review.

gsunner:
I had some trouble running the example.

I think I have found the problem

In the file:

seldon-core/examples/models/sklearn_iris_multipart_formdata/.s2i/environment

Can you remove the lines that start with the following, these should not be checked in.
HTTPS_PROXY
HTTP_PROXY
NO_PROXY

Also could you remove this file:
seldon-core/examples/models/sklearn_iris_multipart_formdata/IrisClassifier.sav
As it will be created in the example

gsunner:
Also when testing the model locally, the step in the notebook that does the following```

klalit:
Cleanup done. Thank you

PR #751, opened 2019-08-03
add the support for multipart/form-data in python service and engine

klalit:
Implements #733

All of the keys in SeldonMessage can be passed using "multipart/form-data". This gives the flexibility to load the data in "binData" or "strData" directly from file. The consumers don't have to read the data, do base64 encoding and pass it to the api.


gsunner:
this is similar to the previous PR that was tested, so will merge

PR #757, opened 2019-08-06
WIP: Allow predict for sklearn standalone server as wll as predict_proba

cliveseldon:


cliveseldon:
Needs an update to the operator to not create parameters if they are specified for sklearn server but append to them.

PR #774, opened 2019-08-09
Added mlflow model server

axsaucedo:
This PR contains an MLFlow model server. Key things that needs to be confirmed before landing include:

- [x] MLFlow Model Server Created
- [x] REST works as expected
- [ ] GRPC works as expected
- [x] Documentation added
- [x] Add samples 
- [ ] Add functionality to support from operator

Current issues
* Attempted running with a public google bucket, but getting AUTH permission errors
* GRPC seems to build as expected but when sending requests it doesn't seem to process


axsaucedo:
Using Seldon storage instead of MLFLow storage as MLFLow storage doesn't support anonymous access to google buckets.

axsaucedo:
Documentation has been added

axsaucedo:
Samples have been added (also to the google bucket), but can only be tested properly until operator changes are added.

axsaucedo:
Operator PR has been added via https://github.com/SeldonIO/seldon-operator/pull/49 - waiting for that PR to be landed, and for the GRPC in this PR to be tested in order to land

doc/source/servers/mlflow.md
```diff
@@ -0,0 +1,33 @@
+# SKLearn Server
```
cliveseldon: typo

proto/seldon_deployment.proto
```diff
@@ -108,6 +108,7 @@ message PredictiveUnit {
     SKLEARN_SERVER = 5; // default sklearn server
     XGBOOST_SERVER = 6; // default xgboost server
     TENSORFLOW_SERVER = 7; // default tensorflow server    
+    MLFLOW_SERVER = 8; // default tensorflow server
```
cliveseldon: typo

axsaucedo:
Added a full example using the MLFlow server which aligns to the blog post currently being written: https://github.com/axsaucedo/seldon-core/tree/mlflow_model_server/examples/models/mlflow_server_ab_test_ambassador

servers/mlflowserver/mlflowserver/MLFlowServer.py
```diff
@@ -0,0 +1,49 @@
...
+        if not self.ready:
+            self.load()
+            # TODO: Make sure this doesn't get called from here, but 
+            #   from the actual python wrapper. Raise exception instead
+            #raise requests.HTTPError("Model not loaded yet")
+        #if not feature_names is None and len(feature_names)>0:
+        #    df = pd.DataFrame(data=X, columns=feature_names)
+        #else:
+        #    df = pd.DataFrame(data=X)
+        result = self._model.predict(X)
```
phsiao: Is this contracting to the statement that

```* The input to the model is set to be pandas by default, so the numpy array passed will be converted into a pandas dataframe```

in doc/source/servers/mlflow.md?

Comment (phsiao) [reply servers/mlflowserver/mlflowserver/MLFlowServer.py]: The results should also be converted back to numpy array.

File: servers/mlflowserver/mlflowserver/MLFlowServer.py [hunk 4]
```diff
@@ -0,0 +1,49 @@
...
+            ) -> Union[np.ndarray, List, Dict, str, bytes]:
+
+        log.info(f"Requesting prediction with: {X}")
+        if not self.ready:
+            self.load()
+            # TODO: Make sure this doesn't get called from here, but 
+            #   from the actual python wrapper. Raise exception instead
+            #raise requests.HTTPError("Model not loaded yet")
+        #if not feature_names is None and len(feature_names)>0:
+        #    df = pd.DataFrame(data=X, columns=feature_names)
```
lennon310: Is there a reason this pandas conversion was commented out?

axsaucedo:
@phsiao @lennon310 the image that was pushed to DockerHub doesn't have the Pandas Dataframe commented out so it would still work as expected in the documentation, which is why I must have missed it - thanks for the heads up! We'll submit a PR right away 👍

servers/mlflowserver/mlflowserver/MLFlowServer.py
meoril: @cliveseldon is there any other to send a dataframe into a served MLFlow model? because this sort of breaks their expected API. maybe i'm missing something though..

PR #787, opened 2019-08-12
Document how to use envSecretRefName to customize S3-like object store

phsiao:


phsiao:
This probably can't go in till all the relevant changes of https://github.com/SeldonIO/seldon-operator/pull/50 went it.

cc @ryandawsonuk

PR #789, opened 2019-08-12
Updated MLFlow server removed commented code

axsaucedo:


servers/mlflowserver/mlflowserver/MLFlowServer.py
```diff
@@ -39,10 +39,10 @@ def predict(
...
-        #    df = pd.DataFrame(data=X, columns=feature_names)
-        #else:
-        #    df = pd.DataFrame(data=X)
+        if not feature_names is None and len(feature_names)>0:
+            df = pd.DataFrame(data=X, columns=feature_names)
+        else:
+            df = pd.DataFrame(data=X)
         result = self._model.predict(X)
         log.info(f"Prediction result: {result}")
         return result
```
phsiao: Should this be converted back to numpy also?

phsiao: cc @lennon310

axsaucedo: The output comes back as a numpy array so no need to convert back

PR #808, opened 2019-08-16
Add Chainer MNIST example

dtaniwaki:
I'm going to use Seldon in my project and since we use Chainer as an ML framework, I started it from the train mnist example first. I think this example will be helpful for Chainer users. Could you consider adding this example?


axsaucedo:
Looks great @dtaniwaki , thank you for the contribution, we'll review. Can you please make sure you add an `.nblink` file that points to your ipynb file as well?

dtaniwaki:
@axsaucedo Added and tested it with `make html` in the doc dir in my local env.

axsaucedo:
Lgtm!

axsaucedo:
Awesome, thanks for the contribution!

PR #817, opened 2019-08-23
Fixing multidimensional arrays being flattened 

### LINKED ISSUE #600: NDArray with values being lists of strings not supported
There is currently an issue that was uncovered when building the Kubeflow Seldon example ()https://github.com/SeldonIO/seldon-core/pull/589, which may require either constraining the implementation or extending the type-support functionality. This is basically in the way that SeldonCore currently handles numpy arrays of tokens (i.e. lists of strings).

Currently the example only sends one sentence as a test which works correctly. However if more than one sentence is sent to the SeldonEngine for processing, it then is unable to process this request.

The reason why this is the case is because the return value is of the format np.array(list(str), list(str)), which translates into np.array(list(str).extend(list(str)).

To be more specific with an example, if we send as example np.array(["example one", "example 2")], the spacytokenizer image would return the value np.array.([list("example", "1"), list("example", "2")]), which SeldonEngine then converts into np.array(["example", "1", "example", "2"]).

It may be possible that using JSON objects may solve this issue (i.e. https://github.com/SeldonIO/seldon-core/pull/595)

axsaucedo:
Fixes #600


axsaucedo:
@cliveseldon this fixes #600 and is ready to merge 👍

PR #818, opened 2019-08-24
Update all servers to use 0.12-SNAPSHOT and update to 0.2

cliveseldon:
* New 0.2 images using 0.12-SNAPSHOT
 * mlflowserver - changed pip command to install scikit-learn. Did not work with `sklearn` name.

Operator for 0.4.1-SNAPSHOT has been updated


axsaucedo:
lgtm

PR #819, opened 2019-08-25
Requirements version alignment across setup.py and requirements.txt

### LINKED ISSUE #767: Add versions to the wrapper requirements.txt and make sure they align with setup.py

axsaucedo:
Fixes #767


jklaise:
I would be cautious about pinning dependencies in `setup.py`, the way Python packaging works is that `setup.py` should only have abstract dependencies (i.e. names of packages), so whoever installs the package doesn't have specific versions of dependencies imposed upon them. On the flip side, `requirements.txt` should have either specific versions or minimum versions to reflect an environment that is tested. I have seen packages that programmatically read `requirements.txt` into `setup.py` and strip all the versions, but this would not be enough for us as we do want to pin some packages in `setup.py` (e.g. `jaeger`) for which it is crucial to have version bounds otherwise it breaks.

This is all a consequence of Python's packaging system mess... An alternative would be to investigate the `Pipfile` and `Pipenv` solutions which are getting ever more popular for managing dependencies etc.

cliveseldon:
@jklaise This makes sense. I think we should prune these changes to minimum for setup.py?

axsaucedo:
@jklaise @cliveseldon as discussed I've updated all the dependencies

axsaucedo:
@jklaise requirements have been updated to only modify `tensorflow` for now, and we'll investigate pipenv separately to manage dependencies as discussed.

PR #827, opened 2019-08-28
configurable secure argument for minio client

### LINKED ISSUE #825: Cannot disable secure connection to minio storage backend
When downloading models (e.g., for MLFLOW_SERVER deployment), from a minio backend, a secure connection is enforced. While this is a sane default, it is a problem for local deployments where TLS is tricky to set up. MLFlow itself does not pose such a requirement. 

This should either be configurable via an environment variable or detected automatically from the configured scheme for Minio

xaniasd:
with this change it is possible to configure the secure argument when instantiating a minio client in the following 2 ways:

* use an `S3_ENDPOINT` with https scheme;
* set environment variable USE_SSL to "false"

The default value remains True. 

Resolves #825

### DIFF
--- a/python/seldon_core/storage.py
+++ b/python/seldon_core/storage.py
@@ -21,6 +21,7 @@ import logging
 import tempfile
 import os
 import re
+from urllib.parse import urlparse
 from azure.storage.blob import BlockBlobService
 from google.auth import exceptions
 from google.cloud import storage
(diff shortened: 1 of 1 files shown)


cliveseldon:
We would need to check how this is being used by the Seldon Operator in the init container for this merged fix to be usable.

xaniasd:
@cliveseldon I think setting S3_ENDPOINT with http:// should do the trick, I'll let you know once I've tried it

PR #993, opened 2019-10-24
Decode binData before sending it to Predict function

### LINKED ISSUE #909: Update python wrapper to pass bytes instead of string
We had to implement a non-trivial change to support "proxy servers" in 0.4.0 (https://github.com/SeldonIO/seldon-core/issues/739), which required us to handle some of these points that were previously handled automatically by the proto object serialisation and deserialisation.

The handling of binary data was a point that we discussed in the PR on how to handle, and currently it was modified this way by design (as per some of the tests defining this functionality explicitly https://github.com/SeldonIO/seldon-core/blob/ed63ed736cc6911b90cd2be7bd7b572cf66c33ef/python/seldon_core/utils.py#L520)

Currently the data received can be transformed using:
```
base64.b64decode(X.decode('utf-8')))
```

It was discussed that it would make sense to pass the value to the predict function with this processed with the base64 decode, as it is possible to access the raw in the `predict_raw` function. 

This only needs to be changed in the `seldon_utils.py` file and in the tests.

lennon310:
So the client won't need to implement decode in `predict_raw` by themselves.

Unit tests all passed.

Fixes #909

### DIFF
--- a/python/seldon_core/flask_utils.py
+++ b/python/seldon_core/flask_utils.py
@@ -66,6 +66,9 @@ def get_request() -> Dict:
                 raise SeldonMicroserviceException("Can't find JSON in data")
     if message is None:
         raise SeldonMicroserviceException("Invalid Data Format - empty JSON")
+    if 'binData' in message and message['binData'] is not None:
+        binData = message['binData']
+        message['binData'] = base64.b64decode(binData).decode("utf-8")
     return message
 
 
--- a/python/seldon_core/flask_utils.py
+++ b/python/seldon_core/flask_utils.py
@@ -68,7 +68,7 @@ def get_request() -> Dict:
         raise SeldonMicroserviceException("Invalid Data Format - empty JSON")
     if 'binData' in message and message['binData'] is not None:
         binData = message['binData']
-        message['binData'] = base64.b64decode(binData).decode("utf-8")
+        message['binData'] = base64.b64decode(binData)
     return message
 
 
--- a/python/seldon_core/wrapper.py
+++ b/python/seldon_core/wrapper.py
@@ -16,7 +16,6 @@ from seldon_core.flask_utils import (
     ANNOTATION_GRPC_MAX_MSG_SIZE,
 )
 from seldon_core.proto import prediction_pb2_grpc
-import base64
 import os
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


seldondev:

To fully approve this pull request, please assign additional approvers.
We suggest the following additional approver: **gsunner**

If they are not already assigned, you can assign the PR to them by writing `/assign @gsunner` in a comment when ready.

The pull request process is described [here](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process)

<details open>

</details>

seldondev:
Hi @lennon310. Thanks for your PR.

<details>

Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [kubernetes/test-infra](https://github.com/kubernetes/test-infra/issues/new?title=Prow%20issue:) repository.
</details>

python/seldon_core/wrapper.py
```diff
@@ -113,6 +114,10 @@ def __init__(self, user_model):
         self.user_model = user_model
 
     def Predict(self, request_grpc, context):
+        data_type = request_grpc.WhichOneof("data_oneof")
```
cliveseldon: If people use proto buffers I don't think they need to base64 encode their data as its a binary protoc? So I don't think we need this.

lennon310: @cliveseldon Reverted

axsaucedo:
@lennon310 if you remove `.decode` from your change it should be good to go. 

We had a discussion and confirmed that the bytes may not always be expected to be string, so we would want to make it available as bytes for the user to receive it and use `decode` themselves.

python/seldon_core/flask_utils.py
```diff
@@ -66,6 +66,9 @@ def get_request() -> Dict:
                 raise SeldonMicroserviceException("Can't find JSON in data")
     if message is None:
         raise SeldonMicroserviceException("Invalid Data Format - empty JSON")
+    if 'binData' in message and message['binData'] is not None:
+        binData = message['binData']
+        message['binData'] = base64.b64decode(binData).decode("utf-8")
```
cliveseldon: remove `.decode("utf-8")`

Might need to update tests as well

cliveseldon:
Thanks @lennon310  !

PR #2221, opened 2020-07-31
featt(http2): Update service endpoints

### LINKED ISSUE #2189: Use http2 as service type for new Executor
Since the new executor multiplexes HTTP and GRPC over the same port, the k8's service that is created should also only have one port definition using `http2`. `http2` allows regular http1.x traffic to continue un changed, but also allows istio to properly multiplex the different protocols through the sidecar. The Istio virtual services will also need to be updated to point at the same port for both `http` and `grpc`. 

ex:
``` yaml
apiVersion: v1
kind: Service
...
spec:
  clusterIP: xx.xx.xx.xx
  ports:
  - name: http2
    port: 8000
    protocol: TCP
    targetPort: 8000
  selector:
    seldon-app: test
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```

https://github.com/SeldonIO/seldon-core/blob/9d7bc4932ac043e49854fadec7908189265d46dc/operator/controllers/seldondeployment_controller.go#L630

groszewn:

Currently, gRPC traffic to Istio-enabled SeldonDeployments does not work with the new executor.  This is due to fact that pilot does not allow for a second listener to forward to the same targetPort, causing all traffic destined for the shared targetPort to use the first protocol registered to pilot (which is usually http since it is defined first in the service).  This updates the service to only create a single service port using the http2 protocol, which should allow both gRPC and http traffic through the istio-proxy sidecar and through to the executor.

This also adds the `sidecar.istio.io/rewriteAppHTTPProbers` annotation to all SeldonDeployment pods, as the HTTP health check sent from Kubelet does not have the certificate issued by Istio and will fail if mTLS is enabled.  More information is laid out [here](https://istio.io/latest/docs/ops/configuration/mesh/app-health-check/#liveness-and-readiness-probes-with-http-request-option).

Fixes #2189 

**Special notes for your reviewer**:  I have built and tested this change in my own cluster where I had been experiencing issues.  This is the image built from these changes that I tested with: `groszewn/seldon-core-operator:http2_fix`

```release-note
This PR updates services created by the controller (when the executor is enabled) to only expose a single service endpoint (8000 with http2 as the protocol).  This removes the separate port exposed for gRPC traffic, as istio-enabled deployments are not able to handle two service ports forwarding to the same targetPort.  As the executor multiplexes traffic, only a single exposed port is sufficient for handling both http and gRPC.
```

### DIFF
--- a/operator/controllers/seldondeployment_controller.go
+++ b/operator/controllers/seldondeployment_controller.go
@@ -634,13 +634,7 @@ func createPredictorService(pSvcName string, seldonId string, p *machinelearning
 		},
 	}
 	if isExecutorEnabled(mlDep) {
-		if engine_http_port != 0 && len(psvc.Spec.Ports) == 0 {
-			psvc.Spec.Ports = append(psvc.Spec.Ports, corev1.ServicePort{Protocol: corev1.ProtocolTCP, Port: int32(engine_http_port), TargetPort: intstr.FromInt(engine_http_port), Name: "http"})
-		}
-
-		if engine_grpc_port != 0 && len(psvc.Spec.Ports) < 2 {
-			psvc.Spec.Ports = append(psvc.Spec.Ports, corev1.ServicePort{Protocol: corev1.ProtocolTCP, Port: int32(engine_grpc_port), TargetPort: intstr.FromInt(engine_http_port), Name: "http2"})
-		}
+		psvc.Spec.Ports = append(psvc.Spec.Ports, corev1.ServicePort{Protocol: corev1.ProtocolTCP, Port: int32(engine_http_port), TargetPort: intstr.FromInt(engine_http_port), Name: "http2"})
 	} else {
 		if engine_http_port != 0 && len(psvc.Spec.Ports) == 0 {
 			psvc.Spec.Ports = append(psvc.Spec.Ports, corev1.ServicePort{Protocol: corev1.ProtocolTCP, Port: int32(engine_http_port), TargetPort: intstr.FromInt(engine_http_port), Name: "http"})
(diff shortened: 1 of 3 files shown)


seldondev:
Hi @groszewn. Thanks for your PR.

<details>

</details>

seldondev:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **axsaucedo**
You can assign the PR to them by writing `/assign @axsaucedo` in a comment when ready.

<details open>

</details>

cliveseldon:
@groszewn @glindsell 

 I think we would need to check what we expose in the created VirtualServices and Ambassador configuration.
As they will at present expose 2 ports.
I think this will therefore turn into a breaking change? But sounds like we would need it.
Be good to understand why 2 ports worked in the past?

groszewn:
Agreed, seems like a breaking change since some people may be expecting 5001 to still be available.  Definitely would need to look into the implications for both Istio and Ambassador.  Theoretically we could just rename the two service ports to both be named `http2-*`, but that seems a bit hacky and as far as I can tell pilot would still complain under the hood (but traffic wouldn't be downgraded to http1.x since both protocols would be http2).  I tested this as well on my end and it works.

As far as 2 ports working in the past, the istio-proxy sidecar has no trouble when there are two separate target ports, since it just registers a listener on each.  As two service ports now point to a single target port, there is a conflict when pilot is registering listeners.  As far as I can tell, whichever service port is registered first is the one that will actually be used for all traffic destined for the target port.  This race condition could explain why I'm able to periodically get traffic to work (if the service port with http2 as the protocol is used instead of http).  However, bouncing the SeldonDeployment pod causes the istio-init container to rerun, which re-registers everything with pilot and then causes it to fall back to http.  [This comment](https://github.com/istio/istio/issues/9273#issuecomment-444220564) in particular gave me the idea to just define both as http2, which worked.

groszewn:
@cliveseldon @glindsell Istio by default won't autoinject sidecars into pods unless the namespace is annotated to enable that functionality.  Is there a chance that the integration tests deploy Seldon with Istio enabled, but don't deploy the `SeldonDeployment` into a namespace with the `istio-injection: enabled` label?  If so, that would explain why previous tests pass with 2 ports forwarding to the same targetPort, as the sidecar proxies aren't receiving any of the traffic (since they don't exist).  This would result in traffic going directly to the executor, which we already know works.

cliveseldon:
@groszewn Yes I think this may be the case with the test that were run.

groszewn:
After further discussion in #2189 and offline, we have decided the best path forward currently is to revert multiplexing.  [This](https://github.com/SeldonIO/seldon-core/pull/2365) PR will handle that update, so this PR is no longer needed.

PR #2304, opened 2020-08-20
Update install.md

omerfsen:
To state Isito or Ambassador is required for a normal setup

Fixes #

```release-note
```

### DIFF
--- a/doc/source/workflow/install.md
+++ b/doc/source/workflow/install.md
@@ -6,6 +6,9 @@
 * Installer method
     * Helm version equal or higher than 3.0
     * Kustomize version equal or higher than 0.1.0
+* Ingress
+    * Istio ( sample installation using Istio 1.5 can be found at https://github.com/SeldonIO/seldon-core/tree/master/examples/auth )
+    * Ambassador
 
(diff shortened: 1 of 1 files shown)


seldondev:
Hi @omerfsen. Thanks for your PR.

<details>

</details>

axsaucedo:
Thank you for the contribution Omer - in this case you are linking specifically the auth page (as opposed to the generic istio / ambassador page). The ingress is not necessary to set up seldon core (we do have a smaller number of users running seldon core in closed environments where non-supported ingress are configured), so probably we can leave it out. I do believe ingress is mentioned in the documentation lower down the page. Are there any reasons why the installation below that outlines istio/ambassador with detailed pages are not enough?

ryandawsonuk:
Would do you reckon to moving the 'Ingress Support' section from https://github.com/SeldonIO/seldon-core/blob/master/doc/source/workflow/install.md further up the page so that it comes before 'Install a specific version'? Currently we've got a paragraph that ends with 'install it with the relevant ingress (ambassador.enabled or istio.enabled) so you are able to send requests (instructions below).' and then 'instructions below' are quite a long way down.

PR #2713, opened 2020-11-28
Enable customizable bins for TIMER metric

### LINKED ISSUE #2663: Is it possible to overwrite the bins for TIMER metrcis
I see the bins are hard-coded: https://github.com/SeldonIO/seldon-core/blob/master/python/seldon_core/metrics.py#L38, just wondering if it's possible to overwrite it somehow.

More generally, I'm thinking of exporting prediction values to Grafana for visualization, which metrics type should I use?  I thought I would want to have a histogram over time kind of visualization like the one shown on https://grafana.com/docs/grafana/latest/getting-started/intro-histograms/.

zyxue:

An idea for making `BINS` for TIMER metrcis customerizable via a tag

Fixes #2663

* I'm not totally sure if this is the right way to customerize `BINS`.
* I used the tag `__SELDON_TIMER_BINS__`, let me know if there is more preferrable name.

```release-note
Added support to configuring BINS in TIMER metrics
```

### DIFF
--- a/python/seldon_core/metrics.py
+++ b/python/seldon_core/metrics.py
@@ -135,13 +135,35 @@ class SeldonMetrics:
                 value = worker_data.get(key, {}).get("value", 0)
                 worker_data[key] = {"value": value + metrics["value"], "tags": tags}
             elif metrics_type == "TIMER":
+                try:
+                    bins = json.loads(
+                        tags.get("__SELDON_TIMER_BINS__", json.dumps(BINS))
+                    )
+                    assert isinstance(bins, list)
+                    # ensure the elements are all numerical
+                    bins = np.array(bins).astype(float).tolist()
+                except (
+                    json.decoder.JSONDecodeError,
+                    AssertionError,
+                    ValueError,
+                ) as err:
+                    logger.exception(err)
+                    bins = BINS
+
+                tags_clean = {
+                    k: v for k, v in tags.items() if k != "__SELDON_TIMER_BINS__"
+                }
+
                 vals, sumv = worker_data.get(key, {}).get(
-                    "value", (list(np.zeros(len(BINS) - 1)), 0)
+                    "value", (list(np.zeros(len(bins) - 1)), 0)
                 )
+
                 # Dividing by 1000 because unit is milliseconds
                 worker_data[key] = {
-                    "value": self._update_hist(metrics["value"] / 1000, vals, sumv),
-                    "tags": tags,
+                    "value": self._update_hist(
+                        metrics["value"] / 1000, vals, sumv, bins
+                    ),
+                    "tags": tags_clean,
                 }
             elif metrics_type == "GAUGE":
                 worker_data[key] = {"value": metrics["value"], "tags": tags}
@@ -201,8 +223,8 @@ class SeldonMetrics:
         return "_".join(["-".join(i) for i in tags.items()])
 
     @staticmethod
-    def _update_hist(x, vals, sumv):
-        hist = np.histogram([x], BINS)[0]
+    def _update_hist(x, vals, sumv, bins):
+        hist = np.histogram([x], bins)[0]
         vals = list(np.array(vals) + hist)
         return vals, sumv + x
 
--- a/python/tests/test_metrics.py
+++ b/python/tests/test_metrics.py
@@ -1,5 +1,6 @@
 import os
 import logging
+import flask.wrappers
 import pytest
 import numpy as np
 from google.protobuf import json_format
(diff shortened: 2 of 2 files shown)


seldondev:
Hi @zyxue. Thanks for your PR.

<details>

</details>

seldondev:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign **sachinvarghese**
You can assign the PR to them by writing `/assign @sachinvarghese` in a comment when ready.

<details open>

</details>

axsaucedo:
Thank you very much for your contribution. This approach is quite interesting actually @zyxue - we were discussing this briefly earlier in the week. 

Looking at the PR it seems that you are using tags - to provide a bit more context, tags are used to expose information to prometheus as a metric in prometheus has the concept of tags, so this probably wouldn't be the way to go. 

It seems to me that perhaps instead of adding this propriety to tags, we could create a new dictionary key expected in metrics called "meta", which could contain further information like this. Although this is something that would also need updating on the protos, so may be a bit of a bigger change. 

The alternative is to allow for bins to be modified via an env variable (or set of env variables that could be passed to the np.logspace / np.linspace function - this could be the easiest / immediately-cleanest if you are ok setting this across all metrics.

seldondev:
Sat Nov 28 07:39:24 UTC 2020
The logs for [pr-build] [1] will show after the pipeline context has finished.  
https://github.com/SeldonIO/seldon-core/blob/gh-pages/jenkins-x/logs/SeldonIO/seldon-core/PR-2713/1.log

impatient try  
jx get build logs SeldonIO/seldon-core/PR-2713 --build=1

seldondev:
Sat Nov 28 07:39:37 UTC 2020
The logs for [lint] [2] will show after the pipeline context has finished.  
https://github.com/SeldonIO/seldon-core/blob/gh-pages/jenkins-x/logs/SeldonIO/seldon-core/PR-2713/2.log

impatient try  
jx get build logs SeldonIO/seldon-core/PR-2713 --build=2

seldondev:
@zyxue: The following test **failed**, say `/retest` to rerun them all:

Test name | Commit | Details | Rerun command
--- | --- | --- | ---
lint | 687d6c116d424a8088771f7e1ae7f8438809736f | [link]() | `/test lint`



<details>

Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [jenkins-x/lighthouse](https://github.com/jenkins-x/lighthouse/issues/new?title=Command%20issue:) repository. I understand the commands that are listed [here](https://go.k8s.io/bot-commands).
</details>

axsaucedo:
Btw @zyxue to ensure the lint passes you have to rebase and ensure that the post-commit hook is enabled, here is more information https://github.com/SeldonIO/seldon-core/blob/master/CONTRIBUTING.md#coding-conventions

zyxue:
The lint error is due to black version, locally I'm using `black 20.8b1`. The build used `19.10b0`. 

I feel 20.8b1 formatted one (red) quite a bit more readable. Would you consider upgrading black?

```diff
diff --git a/python/tests/test_metrics.py b/python/tests/test_metrics.py
index b0b91ed8a..b3e8967b6 100644
--- a/python/tests/test_metrics.py
+++ b/python/tests/test_metrics.py
@@ -823,10 +823,7 @@ def test_proto_seldon_metrics_endpoint(cls, client_gets_metrics):
 def test_seldon_metrics_predict_update_timer_with_custom_bins():
     class Predictor:
         fake_response = {
-            "data": {
-                "names": ["input-foo"],
-                "data": ["output-bar"],
-            },
+            "data": {"names": ["input-foo"], "data": ["output-bar"],},
             "meta": {
                 "metrics": [
                     {
@@ -872,42 +869,23 @@ def test_seldon_metrics_predict_update_timer_with_custom_bins():
     assert seldon_metrics_history == [
         {
             key: {
-                "value": (
-                    [0.0, 1.0],  # count per bin
-                    20.2 / 1000,  # sum of values
-                ),
-                "tags": {"method": "predict"},
-            }
-        },
-        {
-            key: {
-                "value": ([0.0, 2.0], 40.4 / 1000),
+                "value": ([0.0, 1.0], 20.2 / 1000,),  # count per bin  # sum of values
                 "tags": {"method": "predict"},
             }
         },
+        {key: {"value": ([0.0, 2.0], 40.4 / 1000), "tags": {"method": "predict"},}},
     ]
 
 
 @pytest.mark.parametrize(
-    "invalid_bins_value",
-    [
-        ["a", 1, 2],
-        "abc",
-        ["x"],
-        7,
-        21.3,
-        {},
-    ],
+    "invalid_bins_value", [["a", 1, 2], "abc", ["x"], 7, 21.3, {},],
 )
 def test_seldon_metrics_predict_update_timer_with_custom_bins_fallback_to_default_BINS_when_custom_bins_are_not_valid(
     invalid_bins_value,
 ):
     class Predictor:
         fake_response = {
-            "data": {
-                "names": ["input-foo"],
-                "data": ["output-bar"],
-            },
+            "data": {"names": ["input-foo"], "data": ["output-bar"],},
             "meta": {
                 "metrics": [
                     {
@@ -939,13 +917,7 @@ def test_seldon_metrics_predict_update_timer_with_custom_bins_fallback_to_defaul
         f"__SELDON_TIMER_BINS__-{json.dumps(invalid_bins_value)}_method-predict",
     )
     assert seldon_metrics.data[os.getpid()] == {
-        key: {
-            "value": (
-                [1] + [0.0] * 50,
-                1e-6,
-            ),
-            "tags": {"method": "predict"},
-        }
+        key: {"value": ([1] + [0.0] * 50, 1e-6,), "tags": {"method": "predict"},}
     }
     assert len(seldon_metrics.data[os.getpid()][key]["value"][0]) == len(BINS) - 1
 
@@ -954,10 +926,7 @@ def test_proto_seldon_metrics_predict_update_timer_with_custom_bins():
     class Predictor:
         fake_response = prediction_pb2.SeldonMessage(
             data=prediction_pb2.DefaultData(
-                tensor=prediction_pb2.Tensor(
-                    shape=(1, 2),
-                    values=np.array([3, 5]),
-                )
+                tensor=prediction_pb2.Tensor(shape=(1, 2), values=np.array([3, 5]),)
             ),
             meta=prediction_pb2.Meta(
                 metrics=[
@@ -986,8 +955,7 @@ def test_proto_seldon_metrics_predict_update_timer_with_custom_bins():
     seldon_metrics_history = []
     for _ in range(2):
         response = app.Predict(
-            request_grpc=prediction_pb2.SeldonMessage(),
-            context=None,
+            request_grpc=prediction_pb2.SeldonMessage(), context=None,
         )
         assert isinstance(response, prediction_pb2.SeldonMessage)
         assert response == Predictor.fake_response
@@ -1011,13 +979,5 @@ def test_proto_seldon_metrics_predict_update_timer_with_custom_bins():
                 "tags": {"method": "predict"},
             }
         },
-        {
-            key: {
-                "value": (
-                    [0.0, 0.0, 2.0],
-                    60 / 1000,
-                ),
-                "tags": {"method": "predict"},
-            }
-        },
+        {key: {"value": ([0.0, 0.0, 2.0], 60 / 1000,), "tags": {"method": "predict"},}},
     ]
```

RafalSkolasinski:
@zyxue we may consider upgrading black but probably would prefer to do it through separate issue as this would affect more of the code base

python/seldon_core/metrics.py
```diff
@@ -135,13 +135,35 @@ def update(self, custom_metrics: List[Dict], method: str):
                 value = worker_data.get(key, {}).get("value", 0)
                 worker_data[key] = {"value": value + metrics["value"], "tags": tags}
             elif metrics_type == "TIMER":
+                try:
+                    bins = json.loads(
+                        tags.get("__SELDON_TIMER_BINS__", json.dumps(BINS))
```
RafalSkolasinski, line 140: It'd be cleaner to have this as a helper method defined at the top of the file, next to `split_image_tag` helper method.

python/seldon_core/metrics.py
```diff
@@ -135,13 +135,35 @@ def update(self, custom_metrics: List[Dict], method: str):
                 value = worker_data.get(key, {}).get("value", 0)
                 worker_data[key] = {"value": value + metrics["value"], "tags": tags}
             elif metrics_type == "TIMER":
+                try:
```
RafalSkolasinski, line 138: instead of try...except we should just check if custom bins are provided

RafalSkolasinski (changes requested):
Thanks for the contribution, this is a definitely a desired feature to have.

If we only want to overwrite `BINS` value globally, I think this should be done by setting an environmental variable. This could be for example `SELDON_METRICS_TIMER_BINS` which would contain coma-separate values. 

If we want to have on per metrics item then actually would suggest `TIMER` type to return `bins` explicitly, something like
```
{"type": "TIMER", "key": "mytimer", "value": 20.2, bins=[1, 10, 20, 30]},
```

RafalSkolasinski:
codewise could be sth like
```
            elif metrics_type == "TIMER":
                if "bins" in metrics:
                    bins = metrics.get("bins")
                else:
                    bins = BINS
```
(or calling some helper method that sets bins)

zyxue:
@RafalSkolasinski , why is this metrics type called `TIMER`, I wonder why is it not called `HISTOGRAM` as its unit doesn't have to milliseconds or seconds, right?

RafalSkolasinski:
I believe `TIMER` is historical name. It was under this name defined in our legacy orchestrator `Engine`. 
I suspect that Prometheus changed how this metric type is named over the time but I never checked.

RafalSkolasinski:
@zyxue let me know if you have any questions on the point raised above, happy to help and discuss more if needed

zyxue:
@RafalSkolasinski , is it appropriate to consider `TIMER` as just an alias for `HISTOGRAM` in Seldon (although 1000x scaling seconds <=> milliseconds) may add some surprise)?

RafalSkolasinski:
@zyxue Apologies for later reply. Timer was the name of this metrics used in the [Micrometer](https://micrometer.io/docs/registry/prometheus) library that we used in deprecated now java `Engine`. 

In model's `metrics` method user specify unit of TIMER in millisecond. This is then translated to seconds and as such exposed to the Prometheus.

zyxue:
@RafalSkolasinski , I mean the unit doesn't have to just about seconds/milliseconds. It can also mean e.g. kilograms/grams etc.,  right? TIMER makes it sound the metric has to be about time (e.g. latency), But after all if it's just histogram under the hood, it can be about anything numerical, especially if the BINS become customizable. Do I understand correctly?

RafalSkolasinski:
In principle, yes. But here it is used in context of the network metrics so it kind-of have a meaning of time.

We could introduce `HISTOGRAM` type explicitly for which we would not scale it by factor of `1000` by default - that could cover other units.

Out of curiosity - do you have an use case (with Seldon) for which you would like to store metrics in unit of kg?

zyxue:
>Out of curiosity - do you have an use case (with Seldon) for which you would like to store metrics in unit of kg?
I'm thinking of outputting the histogram of predicted values for monitoring purpose. Later on, when the ground truth is available (there is a delay of the availability of predicted value and ground truth), I'd also like to output the residual for monitoring. The model can be about predicting anything, not just time-related quantities. 

I'm really looking for a histogram metrics type. And to me under the hood of the TIMER metric, it's just a histogram, the name itself feels kind of over-specific. If you implement a histogram type, it's would be pretty much the same code as TIMR, wouldn't it?

Also, I don't understand why didn't seldon ask user to specify the unit in seconds directly? I'm not sure why two units are involved.

RafalSkolasinski:
Yes, it would be the same. We are happy to accept a PR that would allow to specify `HISTOGRAM` type without any unit conversion.

RafalSkolasinski:
@zyxue just accepted your other PR regarding `black` update so the link errors will probably go away.
There are still however few changes left to be made before we can merge it.

RafalSkolasinski:
@zyxue Are you still interested in contributing this feature?

zyxue:
@RafalSkolasinski , sorry for the delay. I haven't got a chance to revisit this yet. It's nice that https://github.com/SeldonIO/seldon-core/pull/2769 has been accepted. I'll try to update this pull request with `HISTOGRAM` type approach we discussed next week.

RafalSkolasinski:
Yes, we can add `HISTOGRAM` as alias to `TIMER` but main adjustment was the way user would be defining custom bins. 
I think having it explicit as 
```
{"type": "TIMER", "key": "mytimer", "value": 20.2, bins=[1, 10, 20, 30]},
```
seems most elegant.

zyxue:
I need to recall where I was before. but my overall impression is that with `HISTOGRAM`, `TIMER` could possibly be deprecated (because it's indeed just a HISTOGRAM), and the conversion between `ms` and `s` is confusing (surprising in a way). Or in the doc, it could be made explicit that `TIMER` is just `HISTOGRAM`, but with a specific set of bins pre-configured mainly for latency measurements (I assume this is the case), and it converts `ms` => `s` automatically.

RafalSkolasinski:
We should remember about backward compatibility. There are still out there codes that probably specify metrics as `TIMER` so we should not deprecate it. 

Adding option to  specify `HISTOGRAM` metrics sounds good to me.

zyxue:
@RafalSkolasinski , please take a look at https://github.com/SeldonIO/seldon-core/pull/2907

RafalSkolasinski:
superseded by https://github.com/SeldonIO/seldon-core/pull/2907

PR #3171, opened 2021-05-07
include seldonio/rclone-storageinitializer in release.py script

RafalSkolasinski:
Fix release script to include seldonio/rclone-storage-initializer introduced by https://github.com/SeldonIO/seldon-core/pull/3159


RafalSkolasinski:
I tested that it works

PR #3819, opened 2021-12-27
add CORS support

### LINKED ISSUE #3818: CORS Support for seldon-request-logger
Sending inference requests through the browser to a kserve inference server with the request logger proxying requests is a desired capability. In order to facilitate that, it would be convenient if the seldon-request-logger would allow for CORS configuration.

JakeNeyer:
Adds CORS support to the request logger for inference requests to kserve servers from the browser

Fixes #3818

```release-note
Added CORS support to seldon-request-logger
```

### DIFF
--- a/components/seldon-request-logger/README.md
+++ b/components/seldon-request-logger/README.md
@@ -158,6 +158,27 @@ Here is a truncated version of an example output for an image use-case with outl
 }
 ```
 
+# CORS Support
+```
+# Enable CORS (use "1" or "true")
+export CORS_ENABLED="1"
+
+# Comma-delimited string of origins or wildcard "*"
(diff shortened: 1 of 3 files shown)


axsaucedo:
Thank you very much @JakeNeyer - would you be able to rebase from master? Several of the tests have been fixed - we'll be able to review once that's updated

axsaucedo:
@JakeNeyer we have syncd internally and it seems like given we're looking to depricate this component in favour of a golang based component, as well as currently not having common situations where the request logger would be hit through the browser directly we won't be looking to add CORS as a supported feature, although we did add functionality to ensure that when the models get a request sent it is possible to add a header to ensure that the request is not forwarded to the request logger

PR #4487, opened 2022-12-07
Return meaningful error on empty bucket sync for model uri

cliveseldon:
rclone will happily sync an empty bucket without error which results in a later error on model load that no rclone folder can be found which is not very informative to the user. This update checks if no rclone folder was created in the sync and returns an informative error.

### DIFF
--- a/samples/models/error-uri-notexist.yaml
+++ b/samples/models/error-uri-notexist.yaml
@@ -1,0 +1,8 @@
+apiVersion: mlops.seldon.io/v1alpha1
+kind: Model
+metadata:
+  name: iris
+spec:
+  storageUri: "gs://seldon-models/scv2/xyz"
+  requirements:
+  - sklearn
--- a/scheduler/pkg/agent/rclone/rclone.go
+++ b/scheduler/pkg/agent/rclone/rclone.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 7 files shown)


scheduler/pkg/agent/repository/triton/config/model_config.pb.go
```diff
@@ -48,11 +48,11 @@ const (
 	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
 )
 
-//@@
-//@@.. cpp:enum:: DataType
-//@@
-//@@   Data types supported for input and output tensors.
-//@@
+// @@
```
sakoush, line 51: why is this file changing?

samples/models/error-uri-notexist.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: mlops.seldon.io/v1alpha1
```
sakoush, line 1: this is used?

sakoush (commented):
I left some clarification comments.

scheduler/pkg/agent/rclone/rclone.go
```diff
@@ -308,6 +320,14 @@ func (r *RCloneClient) Copy(modelName string, srcUri string, config []byte) (str
 	if err != nil {
 		return "", fmt.Errorf("Failed to sync/copy %s to %s %w", srcUpdated, dst, err)
 	}
+	// Even if we had success from rclone the src may be empty so need to check
```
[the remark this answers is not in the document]
cliveseldon: I think we assume at least a model-settings.json though?

scheduler/pkg/agent/rclone/rclone_test.go
```diff
@@ -19,25 +19,60 @@ package rclone
...
-	createTestRCloneMockResponders(host, port, status, body)
+	createTestRCloneMockRespondersBasic(host, port, status, body)
+	return r
+}
+
+func createFakeRcloneClientForCopy(t *testing.T, status int, body string, createLocalFolder bool) *RCloneClient {
+	logger := log.New()
+	log.SetLevel(log.DebugLevel)
+	host := "rclone-server"
+	port := 5572
```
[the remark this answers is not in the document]
cliveseldon: None of the ports are actually used as they are either mocked or not used in the tests.

samples/models/error-uri-notexist.yaml
cliveseldon: will rmeove

sakoush (approved):
LGTM

PR #4500, opened 2022-12-13
Update markdown transforming sed scripts

agrski:
## Motivation 
Jupyter notebooks are our primary method of providing interactive examples for reference and exploration.
We generate Markdown equivalents of these using Jupyter's nbconvert functionality.

The Markdown versions are easier to read for anyone skimming through or wanting reference rather than interactivity.
Unfortunately, the output of nbconvert is very much notebook focused.

This PR aims to improve the formatting in the generated Markdown documents so they don't assume being copy/pasted into Jupyter or IPython sessions.

## Changes
* Blocks containing plings/bangs (!) at the start of a line are treated as bash blocks rather than Python.
  * Blocks containing Python that assigns a shelled out element are not affected.
* Output blocks are treated as code/monospace blocks.
  * We try to be clever about detecting JSON and YAML, with everything else being simply treated as a code block.
* Multiple blank lines are coalesced to improve space efficiency on-screen.

This PR adds a set of sed scripts to improve the formatting of Markdown versions of samples notebooks.
Changes include:
* Recognising shell-outs and formatting these as shell blocks (removing the need for plings/bangs).
* Removing indentation from outputs.
* Treating outputs as YAML, JSON, or simply monospace blocks as appropriate (based on simple inference rules).
* Removing unnecessarily many blank lines between sections for legibility when viewing raw rather than rendered Markdown.


seldondev:

To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please assign 
You can assign the PR to them by writing `/assign ` in a comment when ready.

The full list of commands accepted by this bot can be found [here](https://jenkins-x.io/v3/develop/reference/chatops/?repo=SeldonIO%2Fseldon-core).

<details open>

- **[OWNERS](https://github.com/SeldonIO/seldon-core/blob/v2/OWNERS)**

</details>

agrski:
@adriangonz [this](https://github.com/SeldonIO/seldon-core/pull/4500/commits/65c5db5f60099171b73157f1335c259dd74b3e58) commit fixes the issues with formatting of Python functions you found (and actually also bash line continuations).

I've also added some comments to explain the transformation algorithms so the sed programs are less magical!

samples/trigger_joins.md
```diff
@@ -17,310 +16,326 @@ def get_request_string(use_trigger_1=True, use_trigger_2=True):
...
+      triggersJoinType: any
+  output:
+    steps:
+      - join_node
 
+```
 
-```python
+```bash
 request_string = get_request_string(use_trigger_1=True, use_trigger_2=True)
```
agrski, line 136: :information_source: This is a case of the shell script inference getting it wrong (this isn't a _pure_ bash block), but equally it's not a pure Python block, so we couldn't simply copy/paste anyway without IPython or Jupyter.  If using those then it's probably best to just use the notebooks directly as they're executable.

All in all, I think this is a reasonable mistake to let slip.

samples/inference.md
```diff
@@ -6,337 +6,353 @@ We will show:
    * REST and gRPC using seldon CLI, curl and grpcurl
  * Pipeline inference
    * REST and gRPC using seldon CLI, curl and grpcurl
- 
-
 
 
 ```python
 %env INFER_ENDPOINT=0.0.0.0:9000
```
agrski, line 12: :question: Do we think this should change?  This magic is essentially doing `export INFER_ENDPOINT=...`, so we could detect this and format it as a bash block.

If so, I'd prefer to add a follow-up PR, as this one is already fairly large and complex.

agrski:
This PR does not affect any of the Markdown notebooks under `samples/examples`, due to the globbing in the Makefile.

@cliveseldon are those notebooks still relevant?  If so, are you okay with me updating them in a separate PR.  Again, the idea is to prevent this PR becoming too large and complex and dragging on.

samples/k8s-batch-examples.md
```diff
@@ -15,675 +13,698 @@ os.environ['MESH_IP'] = MESH_IP
...
+2022-11-16 18:26:08,523 [mlserver] INFO - connection timeout: 60
+2022-11-16 18:26:08,523 [mlserver] INFO - micro-batch size: 1
+2022-11-16 18:26:08,620 [mlserver] INFO - Finalizer: processed instances: 100
+2022-11-16 18:26:08,620 [mlserver] INFO - Total processed instances: 100
+2022-11-16 18:26:08,620 [mlserver] INFO - Time taken: 0.10 seconds
 
+```
 
 ```bash
 %%bash
```
agrski, line 519: :information_source: This is another case of IPython magics causing issues.  A user can clearly see not to copy this, but it's still a bit unhelpful.  Much like the `%env` magic, we probably want to have a follow-up PR to handle cases like this.

samples/k8s-batch-examples.md
```diff
@@ -15,675 +13,698 @@ os.environ['MESH_IP'] = MESH_IP
...
 mlserver infer -u ${MESH_IP} -m tfsimple.pipeline -i batch-inputs/tfsimple-input.txt -o /tmp/tfsimple-pipeline-output.txt --workers 5
 ```
+```
```
agrski, line 522: :wrench: This is simply incorrect formatting.  Presumably I've added a `p` statement somewhere I shouldn't have

samples/k8s-batch-examples.md
```diff
@@ -15,675 +13,698 @@ os.environ['MESH_IP'] = MESH_IP
...
+kubectl delete -f pipelines/tfsimple.yaml -n ${NAMESPACE}
+```
 
+```
+model.mlops.seldon.io "tfsimple1" deleted
+pipeline.mlops.seldon.io "tfsimple" deleted
 
+```
 
 ```python
```
agrski, line 709: :information_source: There are at least a few files with empty cells at the end.  Another one for a follow-up PR would be to identify and remove these redundant cells as they're purely noise for viewers of the Markdown files.

samples/k8s-batch-examples.md
agrski: I think this is to do with the fact `nbconvert` outputs this as a bash block in the first place

agrski: Fixed in https://github.com/SeldonIO/seldon-core/pull/4500/commits/9c145773f5f7c767db35f89a43924b733e33cb98

cliveseldon:
> This PR does not affect any of the Markdown notebooks under `samples/examples`, due to the globbing in the Makefile.
> 
> @cliveseldon are those notebooks still relevant? If so, are you okay with me updating them in a separate PR. Again, the idea is to prevent this PR becoming too large and complex and dragging on.

Yes most notebooks under samples/examples are added to docs

samples/examples/huggingface/huggingface.md
```diff
@@ -108,381 +107,395 @@ To allow Alibi-Explain to more easily explain the sentiment we will need:
...
+{"label": "POSITIVE", "score": 0.9998691082000732}
 
+```
 
 We will wait for the explanation which is run asynchronously to the functional output from the Pipeline above.
 
-
 ```python
 while True:
     base64Res = !seldon pipeline inspect speech-to-sentiment.sentiment-explainer.outputs --format json \
```
agrski, line 501: :information_source: This is shelling out, but the block is mostly Python so I think it's better to leave this as it is rather than trying to be too clever and risking false positives based on `!` characters in the middle of things

samples/examples/huggingface/train.md
```diff
@@ -0,0 +1,212 @@
...
+    for j in r:
+        if j["label"] == "POSITIVE":
+            res.append(1)
+        else:
+            res.append(0)
+    return np.array(res)
+```
+
+```python
+predict_fn(["cambridge is great","Oxford is awlful"])
```
agrski, line 60: :smile: Apparently so _awful_ the mere mention of it causes typos!

samples/examples/image_classifier/infer.md
```diff
@@ -121,479 +117,620 @@ def show(X):
...
+  ]
+}
 
+```
 
 ```python
 infer("cifar10-production.pipeline",20, "normal")
 ```
 
+<!-- FIXME All these extra ``` are erroneous and added by my scripts -->
```
agrski, line 269: :wrench: As this comment says, these extra block delimiters are erroneous.  I'm guessing a terminating delimiter condition isn't correctly clearing the hold state.

samples/examples/image_classifier/train.md
```diff
@@ -0,0 +1,280 @@
...
+}
+
+```
+
+Save to local or remote storage the directory. Here we saved to Google Storage:
+
+```bash
+gsutil ls -R gs://seldon-models/mlserver/alibi-detect/cifar10-outlier
+
+gs://seldon-models/mlserver/alibi-detect/cifar10-outlier/:
```
agrski, line 160: :question: This looks like it should be output but for some reason it's part of this shell block.  There's no comparison here as this file wasn't added previously, but I suspect it was probably already the case.

In any case, I'd hope it's reasonably obvious a user shouldn't copy/paste these lines, but rather that they're output.

samples/examples/income_classifier/infer-k8s.md
```diff
@@ -0,0 +1,298 @@
...
+  * Run train.ipynb
+  * `gsutil cp -R gs://seldon-models/scv2/examples/income/infer-data .`
+
+```python
+import os
+os.environ["NAMESPACE"] = "seldon-mesh"
+```
+
+```python
+MESH_IP=!kubectl get svc seldon-mesh -n ${NAMESPACE} -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
```
agrski, line 14: :information_source: This is another case of not being able to catch everything (at least not reliably and robustly).  As before, I'd argue this is acceptable as it's a reasonably uncommon case and fairly obvious as to what's going on here and, thus, how to handle it.

samples/examples/income_classifier/infer-k8s.md
```diff
@@ -0,0 +1,298 @@
...
+    - income
+
+```
+
+```bash
+kubectl apply -f ../../pipelines/income-v2.yaml -n ${NAMESPACE}
+```
+
+```yaml
+Warning: resource pipelines/income-prod is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
```
agrski, line 176: :information_source: Sadly _not_ actually YAML, but the heuristic is to find something that looks like a word, followed by a colon, followed by anything that doesn't include colons - this is an unfortunate (but rare) false positive

samples/examples/income_classifier/train.md
```diff
@@ -169,98 +169,89 @@ train_pipeline = Pipeline(steps=[('preprocessor',preprocessor),('classifier',clf
...
-
-
+```
```
agrski, line 173: :information_source: Unfortunately this is another false positive, this time due to suitable indentation to be detected as an output block.

I don't think this is terrible because:
1. The above block of HTML already renders poorly and is not human-readable
2. [This](https://github.com/SeldonIO/seldon-core/blob/10a8dff21bf04fabe326a258197cb02470326576/samples/examples/income_classifier/train.md) renders in practically the same way as [what's already there](https://github.com/SeldonIO/seldon-core/blob/fb442883d7e9014405f5675f79f7164cf2619359/samples/examples/income_classifier/train.md), presumably due to the divs?

I think the best solution here is actually just to remove the offending content/modify the generating cell appropriately, as it's very much an edge case.

samples/local-experiments.md
```diff
@@ -1,1079 +1,1150 @@
...
 
-```python
-!seldon pipeline infer pipeline-add10 --show-headers --inference-mode grpc \
- '{"model_name":"add10","inputs":[{"name":"INPUT","contents":{"fp32_contents":[1,2,3,4]},"datatype":"FP32","shape":[4]}]}' 
 ```
+> /inference.GRPCInferenceService/ModelInfer HTTP/2
+> Host: 0.0.0.0:9000
+> seldon-model:[pipeline-add10.pipeline]
 
-    > /inference.GRPCInferenceService/ModelInfer HTTP/2
```
agrski, line 455: :information_source: We're definitively improving over the existing solution here, where cURL formatting is erroneously interpreted as a quotation (for outbound metadata)

agrski:
Moving back to draft as I've identified a couple of issues introduced by this PR

samples/examples/image_classifier/infer.md
agrski: These are added by the `format-output-blocks.sed` script specifically.
It turns out they _are_ outputs from the `infer` calls and, although it isn't easy to see in GitHub diffs, the links are wrapped with lines of four leading spaces, which is detected by the script as an output block.
The links themselves are not indented, thus terminate the output block, but the line following each again looks like an output block and ends up creating empty outputs.

So far as I can see, the beginning and terminating delimiters _are_ correctly matched, so this is more noise rather than being genuinely incorrect output.

I can try adding an exception for bangs/plings at the start of a line when in an output block to reduce the noise.

samples/markdown-transforms/format-output-blocks.sed
```diff
@@ -0,0 +1,95 @@
...
+  s/^\s{4}//
+  H
+  b END
+}
+
+# We're outside an output block
+x
+
+# Special case for unindented links
+/^.+$/ {
```
agrski, line 52: :wrench: Ideally we would render this link in the normal font rather than as monospace

cliveseldon (approved):

cliveseldon:
Looks good. Maybe we should add this to run automatically in a github action?

agrski:
> Looks good. Maybe we should add this to run automatically in a github action?

This is a good suggestion - I can do a follow-up PR (to keep the changes separate and allow rollback of the GitHub Action is undesired in future).

PR #4520, opened 2022-12-21
Seldon Core V2 Go 3rd party licenses

cliveseldon:
Adds Go 3rd party license tooling and current set of licenses.


sakoush (approved):
LGTM

Given that the target is repeated across several makefiles could we instead have one place where the logic is defined and then have a list of directories to operate on? I understand the urgency so it can be a follow up PR perhaps if it is not straightforward

sakoush:
Also any potentially not friendly licences discovered?

cliveseldon:
No. There are four LGPL projects but with added exceptions that allow for static and dynamic linking, e.g. [here](https://github.com/juju/qthttptest/blob/master/LICENCE).

PR #4526, opened 2022-12-22
Add data-flow 3rd party license generation

cliveseldon:
- Builds license dependencies on demand
- update summary report for all licenses. Taking first license for projects that have multiple licenses. At present this applies only to the rocksdbjni dependency where the first license is Apache 2. If we have more examples like this may need to do more complex logic but was unable to get Kotlin translation of [setup for plugin to work](https://github.com/hierynomus/license-gradle-plugin#license-reporting).


scheduler/data-flow/build.gradle.kts
```diff
@@ -88,3 +89,9 @@ tasks.withType<Jar> {
 application {
     mainClass.set(dataflowMainClass)
 }
+
+
+downloadLicenses {
+    includeProjectDependencies = true
```
agrski, line 95: :thought_balloon: We may want to use the `aliases` [property](https://github.com/hierynomus/license-gradle-plugin#license-reporting), as we appear to have some naming differences in `3rd-party-summary.txt` introduced in this PR, e.g.
```
  8 Apache 2.0
      3 Apache-2.0
    395 Apache License 2.0
      7 Apache License, Version 2.0
```

Happy to do a follow-up PR to tidy this up, as thus far this PR looks good otherwise

cliveseldon: yes - lets do a follow up. As stated in the description it was difficult to get all the settings working in Kotlin.

agrski:
>     * update summary report for all licenses. Taking first license for projects that have multiple licenses. At present this applies only to the rocksdbjni dependency where the first license is Apache 2.

From copying the licence JSON from this PR, there are actually two such examples, not just one:
```bash
$ cat <...> | jq '.dependencies[] | select(.licenses | length > 1) | .name' 
"org.rocksdb:rocksdbjni:6.29.4.1"
"org.checkerframework:checker-compat-qual:2.5.5"
```

agrski (approved):
Changes look good (and simple)

PR #4535, opened 2022-12-27
v2: Remove need for ONESHELL in go lint

cliveseldon:
Removes the need to use the [ONESHELL](https://www.gnu.org/software/make/manual/html_node/One-Shell.html) directive in Makefiles for Go lint download. ONESHELL is a global directive for the entire Makefile and causes issues in getting the go license updates to run as it stops normal error failure during Makefile runs, i.e. an error code from a command should stop the Make run.

From docs:

> Even with this special feature, however, makefiles with .ONESHELL will behave differently in ways that could be noticeable. For example, normally if any line in the recipe fails, that causes the rule to fail and no more recipe lines are processed. Under .ONESHELL a failure of any but the final recipe line will not be noticed by make. You can modify .SHELLFLAGS to add the -e option to the shell which will cause any failure anywhere in the command line to cause the shell to fail, but this could itself cause your recipe to behave differently. Ultimately you may need to harden your recipe lines to allow them to work with .ONESHELL.

### DIFF
--- a/components/tls/Makefile
+++ b/components/tls/Makefile
@@ -7,20 +7,12 @@ test:
 .GOLANGCILINT_VERSION := v1.48.0
 .GOLANGCILINT_PATH := $(shell go env GOPATH)/bin/golangci-lint/$(.GOLANGCILINT_VERSION)
 
-.PHONY: .install-golangci-lint
-.ONESHELL: .install-golangci-lint
-.install-golangci-lint:
-	@installed=false
-	if [ -e "${.GOLANGCILINT_PATH}/golangci-lint" ]; then
-		installed=true
-	fi
-	if ! $${installed}; then
-		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
+${.GOLANGCILINT_PATH}/golangci-lint: 
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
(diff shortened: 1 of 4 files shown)


components/tls/Makefile
```diff
@@ -7,20 +7,12 @@ test:
...
-.ONESHELL: .install-golangci-lint
-.install-golangci-lint:
-	@installed=false
-	if [ -e "${.GOLANGCILINT_PATH}/golangci-lint" ]; then
-		installed=true
-	fi
-	if ! $${installed}; then
-		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
+${.GOLANGCILINT_PATH}/golangci-lint: 
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \
```
sakoush, line 11: Are we always installing go lint every time now? If so how long does it usually take?

cliveseldon: We are not. It should only install if its not there.

sakoush: It is not clear to me how this is done given that we are removing the logic to check if the location of golint exist?

cliveseldon: `${.GOLANGCILINT_PATH}/golangci-lint:` should only run if needed.

sakoush (approved):
LGTM

PR #4543, opened 2023-01-03
Allow ServerConfig and Server labels and annotations to be passed to StatefulSet

### LINKED ISSUE #4542: could you add possibility to set annotations in Server's Pod?
This applies to Seldon Core v2.
It should be set here:
https://github.com/SeldonIO/seldon-core/blob/5a2ee046ee5e90f1886eaaa706cdd76c12041441/operator/controllers/reconcilers/statefulset/statefulset_reconciler.go#L72

I have a case where I need to provide annotations that configure Multus to define the second network interface in Server's Pod that is required by my model's workload.

cliveseldon:
- ServerConfig and Server annotations and labels added to StatefulSet and StatefulSet Pod Template
- Server values override any from ServerConfig

Fixes #4542

### DIFF
--- a/operator/controllers/reconcilers/statefulset/statefulset_reconciler.go
+++ b/operator/controllers/reconcilers/statefulset/statefulset_reconciler.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/seldonio/seldon-core/operator/v2/pkg/utils"
+
 	mlopsv1alpha1 "github.com/seldonio/seldon-core/operator/v2/apis/mlops/v1alpha1"
 	"github.com/seldonio/seldon-core/operator/v2/controllers/reconcilers/common"
 	"github.com/seldonio/seldon-core/operator/v2/pkg/constants"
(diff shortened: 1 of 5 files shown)


operator/controllers/reconcilers/statefulset/statefulset_reconciler.go
```diff
@@ -43,23 +45,35 @@ func NewStatefulSetReconciler(
 	podSpec *v1.PodSpec,
 	volumeClaimTemplates []mlopsv1alpha1.PersistentVolumeClaim,
 	scaling *mlopsv1alpha1.ScalingSpec,
+	serverConfigMeta metav1.ObjectMeta,
```
RafalSkolasinski, line 48: Just general thought: does it matter here if we pass by pointer vs value?

cliveseldon: will leave for now - not sure it will make much difference but good you highlighted

RafalSkolasinski (approved):
LGTM

PR #4557, opened 2023-01-06
Retry ModelFailed models

### LINKED ISSUE #4553: Retry models that fail during control plane load
Currently when a model fails to load on a particular server, we want to be able to:

- Retry this model on another (new) server that comes online
- Delete this model (currently there is a bug that prevents `LoadFailed` models to be deleted, that we need to investigate in this work)

sakoush:

Retry `ModelFailed` when a new agent comes online. We already retry on `ScheduleFailed`. 

We clean up `failedModels` data structure and search the underlying store instead.

Fixes #4553


cliveseldon (approved):

sakoush:
Just to note that not all ModelFailed should be in theory retryable eg in the case of a bad uri. But we dont have a way to differentiate currently and its not clear what kind of runtime issues to load we can get.

PR #4560, opened 2023-01-08
v2: update triton, grpc payload size, cli raw output processing

cliveseldon:
* Updates to a more recent triton server
 * Ensure CLI handles raw payload conversion when there is non-null payload but with all entries null
 * Increase max-grpc message sizes

Note - MLServer would need an increase for this to fully work. So far have not seen an issue with Triton server and increased payload sizes for gRPC above the 4MB default.


scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -119,6 +119,7 @@ func (rp *reverseGRPCProxy) Start() error {
 		opts = append(opts, grpc.Creds(rp.tlsOptions.Cert.CreateServerTransportCredentials()))
 	}
 	opts = append(opts, grpc.MaxConcurrentStreams(grpcProxyMaxConcurrentStreams))
+	opts = append(opts, grpc.MaxRecvMsgSize(util.GrpcMaxMsgSizeBytes))
```
sakoush, line 122: why we are not setting `MaxCallSendMsgSize` as the other ones here as well?

scheduler/pkg/agent/v2.go
```diff
@@ -116,6 +116,7 @@ func getV2GrpcConnection(host string, plainTxtPort int) (*grpc.ClientConn, error
 
 	opts := []grpc.DialOption{
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(util.GrpcMaxMsgSizeBytes), grpc.MaxCallSendMsgSize(util.GrpcMaxMsgSizeBytes)),
```
sakoush, line 119: do we need to set this for control plane operations?

scheduler/pkg/util/constants.go
```diff
@@ -21,5 +21,6 @@ import "time"
 const (
 	GrpcRetryBackoffMillisecs         = 100
 	GrpcRetryMaxCount                 = 5 // around 3.2s in total wait duration
+	GrpcMaxMsgSizeBytes               = 100 * 1024 * 1024
```
sakoush, line 24: is this aligning with max kafka payload size? as we are using it in the pipeline gateway.

k8s/Makefile
```diff
@@ -13,7 +13,7 @@ RCLONE_IMG ?= ${DOCKERHUB_USERNAME}/seldon-rclone:${CUSTOM_IMAGE_TAG}
 SCHEDULER_IMG ?= ${DOCKERHUB_USERNAME}/seldon-scheduler:${CUSTOM_IMAGE_TAG}
 
 MLSERVER_IMG ?= seldonio/mlserver:1.2.1
-TRITON_IMG ?= nvcr.io/nvidia/tritonserver:22.05-py3
+TRITON_IMG ?= nvcr.io/nvidia/tritonserver:22.11-py3
```
sakoush, line 16: any particular reason why are upgrading?

sakoush (commented):
I left some minor comments

sakoush:
> Note - MLServer would need an increase for this to fully work. So far have not seen an issue with Triton server and increased payload sizes for gRPC above the 4MB default.

Have you created an issue on MLServer just not to forget?

cliveseldon:
> > Note - MLServer would need an increase for this to fully work. So far have not seen an issue with Triton server and increased payload sizes for gRPC above the 4MB default.
> 
> Have you created an issue on MLServer just not to forget?

@adriangonz is aware

k8s/Makefile
cliveseldon: the pytorch backend is main reason as its tied to a particular version of pytorch

operator/pkg/cli/infer.go
```diff
@@ -562,12 +562,30 @@ func getDataSize(shape []int64) int64 {
...
+		return true
+	} else {
+		if contents.Fp32Contents == nil &&
+			contents.Fp64Contents == nil &&
+			contents.BoolContents == nil &&
+			contents.UintContents == nil &&
+			contents.Uint64Contents == nil &&
+			contents.BytesContents == nil &&
+			contents.IntContents == nil &&
+			contents.Int64Contents == nil {
```
agrski, line 576: :thought_balloon: Checking for just null values might not be very reliable.  We probably want to check that `len(contents.X) == 0`, as this will cover both the null case and the not-null-but-empty case.

scheduler/pkg/agent/v2.go
agrski: It's not a bad thing to be explicit about what our defaults are, I think, rather than relying on whatever a library happens to set (which may or may not make sense for us, or may not even be set).  With that said, I'd generally expect the control plane to use shorter messages, so wouldn't expect the same maxima.

I'm assuming this value might be used when creating buffers.  I'd imagine smaller buffers are used when possible and expanded up to the maximum, but just in case we might want to use a lower value to avoid guzzling too much memory.

scheduler/pkg/kafka/pipeline/grpcserver.go
```diff
@@ -88,6 +88,7 @@ func (g *GatewayGrpcServer) Start() error {
 		opts = append(opts, grpc.Creds(g.tlsOptions.Cert.CreateServerTransportCredentials()))
 	}
 	opts = append(opts, grpc.MaxConcurrentStreams(maxConcurrentStreams))
+	opts = append(opts, grpc.MaxRecvMsgSize(util.GrpcMaxMsgSizeBytes))
```
agrski, line 91: Same point here as in other server components about setting the max send size to match this.  A response from Kafka might be larger than the input (e.g. in the case of something generative, an up-scaler, etc.).

agrski (commented):
A few thoughts for improvements, but largely look sensible

scheduler/pkg/agent/rproxy_grpc.go
cliveseldon: will update

scheduler/pkg/agent/v2.go
cliveseldon: will remove here

scheduler/pkg/util/constants.go
cliveseldon: That's currently 1Gb so can increase by 10 here

operator/pkg/cli/infer.go
cliveseldon: I think we might want to be careful for non-null empty as that might be a valid result, e.g. existence

sakoush (approved):
LGTM

PR #4568, opened 2023-01-10
v2: Ensure grpc outgoing context has metadata

### LINKED ISSUE #4565: v2: grpc headers not passed to model inference
Model metadata from a grpc request is not passed onwards to model.

Creation of a new request id if not existing should be done earlier if needed so it can be passed onwards in headers. This may affect http as well.

cliveseldon:
Fixes #4565 

- Ensure a requestId is present (created or retrieved) before sending grpc request to server
- Ensures outgoing context for proxy grpc call has metadata from incoming context metadata

### DIFF
--- a/scheduler/pkg/agent/rproxy_grpc.go
+++ b/scheduler/pkg/agent/rproxy_grpc.go
@@ -203,6 +203,18 @@ func (rp *reverseGRPCProxy) createOutgoingCtxWithRequestId(ctx context.Context) 
 	return metadata.AppendToOutgoingContext(ctxNew, util.RequestIdHeader, requestId), requestId
 }
 
+func (rp *reverseGRPCProxy) setTrailer(ctx context.Context, trailer metadata.MD, requestId string) {
+	logger := rp.logger.WithField("func", "SetTrailer")
+	if trailer == nil {
+		trailer = metadata.MD{}
+	}
+	trailer.Set(util.RequestIdHeader, requestId)
+	errTrailer := grpc.SetTrailer(ctx, trailer) // pass on any trailers set by inference server such as MLServer
+	if errTrailer != nil {
+		logger.WithError(errTrailer).Error("Failed to set trailers")
+	}
+}
+
 func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequest) (*v2.ModelInferResponse, error) {
 	logger := rp.logger.WithField("func", "ModelInfer")
 	internalModelName, externalModelName, err := rp.extractModelNamesFromContext(ctx)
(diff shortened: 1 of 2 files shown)


scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -235,22 +235,27 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
...
-		if errTrailer != nil {
-			logger.WithError(errTrailer).Error("Failed to set trailers")
-		}
+	if trailer == nil {
+		trailer = metadata.MD{}
+	}
+	trailer.Set(util.RequestIdHeader, requestId)
+	errTrailer := grpc.SetTrailer(ctx, trailer) // pass on any trailers set by inference server such as MLServer
+	if errTrailer != nil {
+		logger.WithError(errTrailer).Error("Failed to set trailers")
```
sakoush, line 258: could we extract this logic as separate helper?

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -235,22 +235,27 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
 		go rp.metrics.AddModelInferMetrics(externalModelName, internalModelName, metrics.MethodTypeGrpc, elapsedTime, codes.NotFound.String())
 		return nil, status.Error(codes.NotFound, fmt.Sprintf("Model %s not found (err: %s)", r.ModelName, err))
 	}
+
+	// Create an outgoing context for the proxy call to service from incoming context
+	outgoingCtx, requestId := rp.createOutgoingCtxWithRequestId(ctx)
```
sakoush, line 240: do we need to do something similar to `ModelMetadata` and `ModelReady` grpc calls?

scheduler/pkg/agent/rproxy_grpc_test.go
```diff
@@ -316,12 +299,13 @@ func TestAddRequestIdToTrailer(t *testing.T) {
 			rp := reverseGRPCProxy{
 				logger: log.New(),
 			}
-			rp.addRequestIdToTrailer(test.ctx, test.trailer)
-			data := test.trailer.Get(util.RequestIdHeader)
```
sakoush, line 320: As we are removing the trailer tests? could we add them somewhere else?

sakoush (commented):
Left some minor comments

scheduler/pkg/agent/rproxy_grpc.go
cliveseldon: Probably need to decide if requestId is needed for these calls. Maybe a follow on or when we decide its needed.

scheduler/pkg/agent/rproxy_grpc_test.go
cliveseldon: The function had been removed

sakoush (approved):

Issue #4573, opened 2023-01-11
Core V2 Ansible Playbooks should use Helm installation


RafalSkolasinski:
Core V2 Ansible playbooks should install using Helm and not manifests to easier understand installation state.

RafalSkolasinski:
Fixed by https://github.com/SeldonIO/seldon-core/pull/4574

PR #4574, opened 2023-01-11
use helm to install Core v2 via Ansible

RafalSkolasinski:

Uses Helm installation instead of manifests when installing Seldon Core V2 using Ansible.

Fixes https://github.com/SeldonIO/seldon-core/issues/4573


cliveseldon:
Just to clarify - has this changed the default namespace for kafka install as I thought Kafka was namespace local by default so user.yaml and cluster.yaml files in examples are installed in seldon-mesh,

RafalSkolasinski:
That's a good point. Change of Kafka namespace is not necessary but I got the feeling it is easier to get orientation on what is what, see screenshot
image

For the example references, are these the two you have in mind?
- https://docs.seldon.io/projects/seldon-core/en/v2/contents/getting-started/kubernetes-installation/tls/strimzi.html
- https://docs.seldon.io/projects/seldon-core/en/v2/contents/kubernetes/kafka/index.html

Second one currently have a wrong namespace anyway. I could update them as well then if you are okay to install Kafka into `seldon-kafka` namespace? 

Also note another change was to install CRDs into `default`.

RafalSkolasinski:
For now I am leaving Kafka in the same namespace as Core V2, `seldon-mesh`. I enabled mTLS for Kafka in default Ansible setup and tested with image build from https://github.com/SeldonIO/seldon-core/pull/4584

cliveseldon (approved):

Issue #4572, opened 2023-01-11
Add security scans for Core V2 images


RafalSkolasinski:
In similar fashion to [V1 workflow](https://github.com/SeldonIO/seldon-core/blob/master/.github/workflows/security_tests.yml) we should add task to Snyk scans Core V2 images.

RafalSkolasinski:
Closing as complete via https://github.com/SeldonIO/seldon-core/pull/4575

PR #4575, opened 2023-01-11
introduce Snyk security scans for V2 images

RafalSkolasinski:

This PR introduced GitHub Workflow to run Snyk security scans on Core V2 images.

Fixes https://github.com/SeldonIO/seldon-core/issues/4572

Test run results in [here](https://github.com/RafalSkolasinski/seldon-core/actions/workflows/security_tests_v2.yml)

### DIFF
--- a/.github/workflows/security_tests_v2.yml
+++ b/.github/workflows/security_tests_v2.yml
@@ -1,0 +1,189 @@
+name: V2 Security Tests
+
+on:
+  push:
+    branches: [ v2 ]
+  workflow_dispatch:
+    inputs:
+      docker-tag:
+        description: 'Docker tag for scan'
+        default: 'latest'
+        required: false
+
+jobs:
+  security-operator:
+    runs-on: ubuntu-latest
+    container: snyk/snyk:golang
+    steps:
+    - uses: actions/checkout@v2
+    - name: security-golang
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      run: |
+        make -C operator fmt
+        snyk test --file=operator/go.mod --fail-on=upgradable --severity-threshold=high
+
+  security-scheduler:
+    runs-on: ubuntu-latest
+    container: snyk/snyk:golang
+    steps:
+    - uses: actions/checkout@v2
+    - name: security-golang
+      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/security_tests_v2.yml
```diff
@@ -0,0 +1,189 @@
...
+  security-scheduler:
+    runs-on: ubuntu-latest
+    container: snyk/snyk:golang
+    steps:
+    - uses: actions/checkout@v2
+    - name: security-golang
+      env:
+        SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+      run: |
+        make -C scheduler lint
```
adriangonz, line 35: Do you know why need to run the linter here, instead of using Snyk's GH action?

.github/workflows/security_tests_v2.yml
```diff
@@ -0,0 +1,189 @@
...
+
+  security-image-operator:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set default docker tag for builds from v2 branch
+      id: docker-tag
+      run: |
+        USER_INPUT="${{ github.event.inputs.docker-tag }}"
+        echo ::set-output name=value::${USER_INPUT:-"latest"}
```
adriangonz, line 46: Oh wow - is this wizardry to default to `latest` if no tag is present?

adriangonz (approved):
Changes look good! Added a couple minor questions below, but shouldn't block a merge.

.github/workflows/security_tests_v2.yml
RafalSkolasinski: Yeah, that's to use default for input usually coming as parameter to "on_dispatch" in workflows running "on_push". There may be more elegant way (though I doubt, unless some new features were added)  but that is what I found when working on SCv2 release workflow.

.github/workflows/security_tests_v2.yml
RafalSkolasinski: In all other tasks there was `make -C <directory> fmt` but `scheduler` did not have `fmt` but `lint` command. Not sure exactly why it is there. I first assumed it was for Go to pull modules but that does not make sense actually... I may try without it. 

But to why these are here: it's because these were present in V1 tasks.

PR #4583, opened 2023-01-13
Fix model delete hang when FailedSchedule

sakoush:
This PR fixes a bug that when the model fails scheduling (i.e. no server is assigned) then it gets stuck when deleting it afterwards. The reason being is that previously when we delete a model that has no server associated, we did not send an update status event. Therefore the controller never received a `ModelTerminated` status change and therefore got stuck in the finaliser during the delete process.

The fix is to allow an event to published with state change of deleted models even if they have no associated server.


sakoush:
This has been test with a model that loaded successfully and a model that failed scheduling.

cliveseldon (approved):

PR #4586, opened 2023-01-14
Add bytes rawcontents correct handling and ensure headers are passed

cliveseldon:
- Updates handling a v2 rawContents to ensure we can decode BYTES payloads correctly that are returned with rawContents which is an NVIDIA Triton extension where BYTES rawcontents are length encoded <len><bytes><len><bytes> where <len> is a 4 bytes unsigned little endian integer representing the number of bytes for the first element. This allows mixed length bytes to be returned in a single byte block and has also been implemented in MLServer also.
- ensure user headers are passed in grpc request - those starting with `x-` but ignoring the `x-seldon-route` which is used by us for direct envoy routing. This ensures grpc models can use user headers if needed.

Future PR should create a new shared library to stop duplicated code in CLI (operator) and Pipeliegateway (scheduler) projects.

### DIFF
--- a/scheduler/pkg/kafka/gateway/worker.go
+++ b/scheduler/pkg/kafka/gateway/worker.go
@@ -26,6 +26,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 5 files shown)


scheduler/pkg/kafka/pipeline/v2.go
```diff
@@ -311,8 +311,7 @@ func updateResponseFromRawContents(res *v2_dataplane.ModelInferResponse) error {
 			output.Contents.Fp64Contents = make([]float64, getDataSize(output.Shape))
 			err = binary.Read(bytes.NewBuffer(rawOutput), binary.LittleEndian, &output.Contents.Fp64Contents)
 		case tyBytes:
-			output.Contents.BytesContents = make([][]byte, 1)
-			output.Contents.BytesContents[0] = rawOutput
+			output.Contents.BytesContents = convertRawBytesToByteContents(rawOutput)
```
sakoush, line 314: Is this called when returning results to user on request (REST) path?

cliveseldon: yes

scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -301,17 +302,28 @@ func (iw *InferWorker) restRequest(ctx context.Context, job *InferWork, maybeCon
 	return iw.produce(ctx, job, iw.topicNamer.GetModelTopicOutputs(job.modelName), b, false, extractHeadersHttp(response.Header))
 }
 
+// Add all external headers to request metadata
+func (iw *InferWorker) addMetadataToOutgoingContext(ctx context.Context, job *InferWork) context.Context {
```
sakoush, line 306: Could we add a smaller test to check which headers are supposed to be returned? Serves also as a kind of documentation.

sakoush (approved):

scheduler/pkg/kafka/gateway/worker.go
cliveseldon: added

PR #4595, opened 2023-01-17
Add notes on Go module tags in release docs

agrski:
This PR does two things:
* First, and foremost, it explains the need for multiple tags for every release.
* Secondly, and incidentally, it provides a number of minor grammatical and formatting changes in the release docs.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -73,7 +73,8 @@ It must be triggered manually using the [Actions](https://github.com/SeldonIO/se
 
 When triggering the workflow, you must:
 - Select the release branch (here `release-0.1`)
-- Specify the release `version` (here `0.1.0-rc1` - note lack of the `v` prefix!)
+- Specify the release `version`, e.g. `0.1.0-rc1`
+  * Note the lack of the `v` prefix here
 
 ![Triggering Build images](.images/release-6.png)
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -123,4 +123,4 @@ To add the appropriate tags:
   ```
 
 If you are feeling confident in the process, you can chain these together into a longer pipeline.
-Either way, it is best to confirm that the tags appear as expected both via the git CLI and also in the GitHub UI.
+In any case, it is best to confirm that the tags appear as expected both via the git CLI and also in the GitHub UI.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -115,7 +115,7 @@ To add the appropriate tags:
   ```
 * Confirm that all tags point to the same place, e.g. with
   ```
-  git tag contains v2.0.0
+  git tag --contains v2.0.0
   ```
 * Push the tags to the upstream repository, e.g.
   ```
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -23,11 +24,11 @@ The development process follows a standard GitHub workflow.
...
 ![Development Graph](.images/release-1.png)
 
-The main development is happening in `master` branch.
+The main development is happening in the `v2` branch.
 This is where new features land through Pull Requests.
-When all features for new release are merged, for example `v0.1.0`, we cut a branch for `release-0.1`.
+When all features for a new release have been merged, for example `v0.1.0`, we cut a branch for that release, e.g. `release-0.1`.
 
-The `release-0.1` branch will be base for `v0.1.0` release as well as the release candidates, e.g. `v0.1.0-rcX` and successive patch releases `v0.1.X`.
+The `release-0.1` branch will be the base for the `v0.1.0` release as well as the release candidates, i.e. `v0.1.0-rcX`, and successive patch releases, i.e. `v0.1.X`.
```
RafalSkolasinski, line 31: meaning of e.g. is "for example" - i.e. stands for "in other words" which I think does not apply here

RafalSkolasinski: Though....

agrski: I think this is accurate because it's saying this is the form those things would take.
With that said, I'm happy to change both to `e.g.` if you find that less jarring :slightly_smiling_face:

RELEASE.md
```diff
@@ -41,38 +42,59 @@ The [Draft New Release](./.github/workflows/draft-release.yml) workflow is the f
...
-- specify the release `version` (here `0.1.0-rc1` - note lack of `v` prefix!).
+- Select the release branch (here `release-0.1`)
+- Specify the release `version` (here `0.1.0-rc1` - note lack of `v` prefix!).
 
 ![Triggering Build images](.images/release-6.png)
 
 This workflow will then run unit tests and build a series of Docker images that will be automatically pushed to [DockerHub](https://hub.docker.com/).
+
+
+### Add Go module tags
```
RafalSkolasinski, line 83: Could we add required command to run (unless we'd want to automate this via GitHub action).

RafalSkolasinski (approved):
LGTM

RELEASE.md
agrski: Was already on this -- that's why I moved back to draft!  Great minds think alike :wink:

RELEASE.md
RafalSkolasinski: No, you are right. `i.e.` works here better :)

RELEASE.md
RafalSkolasinski: :) - thought it will be something simpler like
```bash
TAG=v2.0.0
git tag apis/go/${TAG}
git tag components/tls/${TAG}
...
```

RafalSkolasinski (approved):
LGTM

RELEASE.md
agrski: I like to describe processes -- if it's completely defined, it should just be a script and the instruction would be "run X"

PR #4611, opened 2023-01-20
Stop Agent if subservices are not ready

### LINKED ISSUE #4603: Deal with cases where agent starts then the underlying services die
We currently check when agent is starting that it can connect to the underlying services such as rclone and mlserver/triton

However these services can restart, die and  we want the agent to be aware of that after startup.

This also include services such as reverse proxies.

The impact is that the system might be failing to load new models and / or inference requests failing.

sakoush:

After the Agent starts we need to keep checking the status of the underlying services (e.g. reverse proxies etc). If there is an error in any of these services we stop the Agent.

In  this PR we also switch all readiness checks for the inference servers to `/v2/health/live` given that during a model load `v2/health/ready` is false and we are only care about if the inference server is up.

Fixes #4603


tests/k6/delete.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
cliveseldon, line 1: is this needed in this PR?

sakoush: no will revert

scheduler/pkg/agent/client.go
```diff
@@ -48,6 +47,11 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+const (
+	// period for subservice ready "cron"
+	periodReadySubService = 10 * time.Second
```
cliveseldon, line 52: If each 2> subservice has 5 secs to be ready thats more than the 10 sec between ticks. Maybe doesn't matter?

sakoush: We could do the check in parallel perhaps? I can increase also the period of cron to 30seconds?

scheduler/pkg/agent/client.go
```diff
@@ -230,66 +264,86 @@ func (c *Client) createConnection() error {
...
+		c.settings.maxElapsedTimeReadySubServiceBeforeStart,
+		c.settings.maxElapsedTimeReadySubServiceAfterStart,
+	); err != nil {
 		return err
 	}
 
 	// grpc reverse proxy
-	if err := startSubService(c.rpGRPC, logger); err != nil {
-		logger.WithError(err).Error("Failed to start grpc proxy")
+	if err := isReadyChecker(isStartup, c.rpGRPC, logger, "Grpc proxy not ready",
```
cliveseldon, line 319: Not sure I follow why internall services need to be rechecked for readiness?

sakoush: As these internal services are really running as separate go routines, I am assuming that there could be errors that cause them to stop for any reason? TBH I have not spotted any error of this kind yet though.

cliveseldon: Could the code be simplified by setting a single var for time rather than passing 2 parameters to each call?

sakoush: done in [af21229](https://github.com/SeldonIO/seldon-core/pull/4611/commits/af2122998018e958d2d970fd8f0b6913344ae2c7)

cliveseldon (approved):

PR #4618, opened 2023-01-24
Dont hold a global lock when publishing to `EventHub`

sakoush:

We should not be holding any locks when we publish to the `eventHub`. This ensures that there is no potential deadlock as we can block while publishing an message in high load scenarions.

This PR removes a couple of global locks towards this object.

There are still places where this can happen though but let as future PR:

- (Agent) Server `Sync` grabbing a model lock while doing `UpdateModelState` : https://github.com/SeldonIO/seldon-core/blob/3165f0fa252ba6f44e224a477ad63dd7a781fa54/scheduler/pkg/agent/server.go#L232-L279
- VersionCleaner `cleanupOldVersions` grabbing a model lock while doing `UnloadVersionModels` : https://github.com/SeldonIO/seldon-core/blob/3165f0fa252ba6f44e224a477ad63dd7a781fa54/scheduler/pkg/scheduler/cleaner/versions.go#L59-L76
- (Agent) Server `AgentEvent` grabbing a model lock while doing `UpdateModelState`: https://github.com/SeldonIO/seldon-core/blob/3165f0fa252ba6f44e224a477ad63dd7a781fa54/scheduler/pkg/agent/server.go#L323-L344

Fixes #

### DIFF
--- a/scheduler/pkg/agent/server.go
+++ b/scheduler/pkg/agent/server.go
@@ -419,9 +419,6 @@ func (s *Server) StopAgentStreams() {
 }
 
 func (s *Server) syncMessage(request *pb.AgentSubscribeRequest, stream pb.AgentService_SubscribeServer) error {
-	s.mutex.Lock()
-	defer s.mutex.Unlock()
-
 	s.logger.Debugf("Add Server Replica %+v with config %+v", request, request.ReplicaConfig)
 	err := s.store.AddServerReplica(request)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


scheduler/pkg/agent/server.go
```diff
@@ -419,9 +419,6 @@ func (s *Server) StopAgentStreams() {
 }
 
 func (s *Server) syncMessage(request *pb.AgentSubscribeRequest, stream pb.AgentService_SubscribeServer) error {
-	s.mutex.Lock()
```
cliveseldon, line 422: Not sure why we are removing this lock

sakoush: `syncMessage` is calling 
- `AddServerReplica` that has already a lock internally (but outside of  `PublishModelEvent`)
- `ScheduleFailedModels` that looks not required a lock

Mainly removing the lock so that `PublishModelEvent` is called with no locks held but also it doesnt look that `syncMessage` requires a lock as explained above.

cliveseldon (approved):

PR #4626, opened 2023-01-30
k6 multiple models run

sakoush:
This PR adds the ability to run k6 tests with multiple types of models .

It also fixes  `GetServerName` panic with nil pointer exception in cases of `ServerReplica` snapshots.


cliveseldon (approved):
Looks good. Can be later PR but maybe we should add some clearer docs on all tests and what they do to make it easier for new people to find and use them?

sakoush:
> Looks good. Can be later PR but maybe we should add some clearer docs on all tests and what they do to make it easier for new people to find and use them?

yes. A follow up PR on docs makes sense.

Issue #4613, opened 2023-01-23
Support UBI-based images for Core V2


adriangonz:

agrski:
Are we happy that this issue has been resolved now?

RafalSkolasinski:
We are indeed

PR #4630, opened 2023-01-31
Move Envoy image to UBI

adriangonz:

Part of #4613


scheduler/Dockerfile.envoy
```diff
@@ -1,9 +1,17 @@
-FROM envoyproxy/envoy:v1.19.1
+FROM envoyproxy/envoy:v1.19.1 as envoy
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal
```
cliveseldon, line 3: Why minimal and not micro?

RafalSkolasinski (approved):
Nice one! Glad it was so simple in the end. Tested on OpenShift and all works as expected.
LGTM

PR #4647, opened 2023-02-06
Retry on server load fail

### LINKED ISSUE #4623: Race condition if model and explainer get rescheduled at the same time
404/503 can happen if we have 2 models (inference model and explainer model) deployed on one instance, if this instance dies and these 2 models gets rescheduled there is going to be a race condition in getting them loaded. If the explainer model tries to load first before the inference model we can get 404 / 503

```
MLServer 1.2.0. To access the new field, you can either update the `settings.json` file, or update the `MLSERVER_PARALLEL_WORKERS` environment variable. The current value of the server-level's `parallel_workers` field is '1'.
2023-01-27 16:34:24,507 [mlserver.grpc] ERROR - Predictor failed to be called on x=['Hello world']. Check that `predictor` works with inputs of type List[str].
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/alibi/explainers/anchors/anchor_text.py", line 519, in _transform_predictor
    prediction = predictor(x)
  File "/opt/conda/lib/python3.8/site-packages/mlserver_alibi_explain/explainers/black_box_runtime.py", line 75, in _infer_impl
    self.infer_metadata = remote_metadata(
  File "/opt/conda/lib/python3.8/site-packages/mlserver_alibi_explain/common.py", line 76, in remote_metadata
    raise RemoteInferenceError(response_raw.status_code, response_raw.reason)
mlserver_alibi_explain.errors.RemoteInferenceError: Remote inference call failed with 503, Service Unavailable
```

_Originally posted by @sakoush in https://github.com/SeldonIO/seldon-core/issues/4562#issuecomment-1406809338_

sakoush:
This PR adds the ability to retry loading / unloading models onto a server (control plane) with max number of attempts. This is different from the current usage of retries with backoff as we also cap retries with a number of attempts given that different models will have different latencies for loading / unloading.

We only do this on the control plane to get around the issue where we are loading an inference model and an associated explainer concurrently. So if an explainer tries to load first, it might fail if the underlying inference model has not loaded yet. With retries this should be able to eventually load the explainer.

Fixes #4623 

### DIFF
--- a/scheduler/cmd/agent/main.go
+++ b/scheduler/cmd/agent/main.go
@@ -59,7 +59,7 @@ const (
 	// period for subservice ready "cron"
 	periodReadySubService = 60 * time.Second
 	// number of retries for loading a model onto a server
-	maxLoadRetryCount = 3
+	maxLoadRetryCount = 6
 	// number of retries for unloading a model onto a server
 	maxUnloadRetryCount = 1
 )
--- a/scheduler/cmd/agent/main.go
+++ b/scheduler/cmd/agent/main.go
@@ -278,7 +278,7 @@ func main() {
 			maxElapsedTimeReadySubServiceBeforeStart,
 			maxElapsedTimeReadySubServiceAfterStart,
 			maxLoadRetryCount,
-			maxLoadRetryCount,
+			maxUnloadRetryCount,
 		),
 		logger,
 		modelRepository,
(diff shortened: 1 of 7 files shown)


scheduler/cmd/agent/main.go
```diff
@@ -58,6 +58,10 @@ const (
 	maxElapsedTimeReadySubServiceBeforeStart = 15 * time.Minute // 15 mins is the default MaxElapsedTime
 	// period for subservice ready "cron"
 	periodReadySubService = 60 * time.Second
+	// number of retries for loading a model onto a server
+	maxLoadRetryCount = 6
```
RafalSkolasinski, line 62: nit: we could think about making it configurable in the future

RafalSkolasinski (approved):
LGTM

PR #4650, opened 2023-02-07
Small CLI Updates

cliveseldon:
- [x] Allow a `--truncate` option to not show data in `seldon inspect` as sometimes data can be very large and you want to just see outline of message
- [x] Allow `seldon pipeline unload` to take a `-f <filename>` like load and similar to k8s kubectl
- [x] Allow `seldon model unload` to take a `-f <filename>`
- [x] Allow `seldon experiment unload` to take a `-f <filename>`

### DIFF
--- a/operator/cmd/seldon/cli/model_unload.go
+++ b/operator/cmd/seldon/cli/model_unload.go
@@ -28,7 +28,6 @@ func createModelUnload() *cobra.Command {
 		Use:   "unload <modelName>",
 		Short: "unload a model",
 		Long:  `unload a model`,
-		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			flags := cmd.Flags()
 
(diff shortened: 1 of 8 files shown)


sakoush (commented):
Do we already allow  models / experiments to be unloaded similarly?

cliveseldon:
> Do we already allow models / experiments to be unloaded similarly?

Updated to include models and experiments

operator/cmd/seldon/cli/model_unload.go
```diff
@@ -49,14 +48,25 @@ func createModelUnload() *cobra.Command {
...
+				return err
+			}
+			var fileBytes []byte
+			if filename != "" {
+				fileBytes = loadFile(filename)
+			}
+			modelName := ""
+			if len(args) > 0 {
+				modelName = args[0]
+			}
```
sakoush, line 62: these blocks of code look that they can be extracted as a helper for the 3 use cases (pipeline / model / experiment).

sakoush (approved):
LGTM. I left a refactor comment but logic is ok.

PR #4658, opened 2023-02-10
V2: TLS Kafka

J11522:
This PR adds support for plain TLS Kafka communication.
Currently communication to a Kafka cluster is only possible with client authentication, e.g. using mTLS or SASL.
This PR makes changes to the existing code to facilitate plain tls support.

With this PR client certificate information becomes optional and support for plain TLS is added when communicating with the Kafka cluster.
This allows for unauthenticated cluster access.

A new TLS Security Protocol 'TLS' is added. However, we suggest do refactor the existing code to

- either distinguish mTLS and TLS based on provided parameters
- or rename the existing protocol 'SecurityProtocolSSL' to 'SecurityProtocolMTLS'

Fixes #

This is an initial request. If the general concept and direction suits you, we can make the changes proposed above as well as update the required documentation.

### DIFF
--- a/scheduler/pkg/kafka/config/tls.go
+++ b/scheduler/pkg/kafka/config/tls.go
@@ -34,8 +34,10 @@ const (
 func AddKafkaSSLOptions(config kafka.ConfigMap) error {
 	protocol := tls.GetSecurityProtocolFromEnv(tls.EnvSecurityPrefixKafka)
 	switch protocol {
-	case tls.SecurityProtocolSSL:
+	case tls.SecurityProtocolTLS:
 		return setupTLSAuthentication(config)
+	case tls.SecurityProtocolSSL:
+		return setupMTLSAuthentication(config)
 	case tls.SecurityProtocolSASLSSL: // Note: we don't yet support SASL_PLAINTXT
 		return setupSASLSSLAuthentication(config)
 	case tls.SecurityProtocolPlaintxt:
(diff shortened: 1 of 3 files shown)


seldondev:
Hi @J11522. Thanks for your PR.

<details>

</details>

scheduler/pkg/kafka/config/tls.go
```diff
@@ -34,8 +34,10 @@ const (
 func AddKafkaSSLOptions(config kafka.ConfigMap) error {
 	protocol := tls.GetSecurityProtocolFromEnv(tls.EnvSecurityPrefixKafka)
 	switch protocol {
-	case tls.SecurityProtocolSSL:
+	case tls.SecurityProtocolTLS:
 		return setupTLSAuthentication(config)
+	case tls.SecurityProtocolSSL:
```
adriangonz, line 39: It may be a bit confusing to have both `TLS` and `SSL` as different protocol options. Given that, `SSL` seems to be just doing mTLS - should we rename the `SSL` flag to `MTLS`?

adriangonz (commented):
Hey @J11522 ,

Nice one! :rocket: 

Just did a first pass at the changes and added a small comment there. Would be good to hear your thoughts!

scheduler/pkg/kafka/config/tls.go
J11522: Upon further investigation, I would propose to use a single SSL option.
Reason: if I read the code correct, the security parameter is directly mapped to the [Kafka Security Protocol ](https://kafka.apache.org/25/javadoc/org/apache/kafka/common/security/auth/SecurityProtocol.html#SSL), at least in the dataflow component.
Kafka does not distinguish between TLS encryption (1-way) and authentication (2-way/mutual/client), as the protocol in both cases is TLS.

cliveseldon: +1

RafalSkolasinski: Yes, I would leave it only as single `security.kafka.protocol: SSL` that corresponds to `security.protocol=SSL` option for the Kafka client. Then everything should be driven by:
- `security.kafka.ssl.client.secret`
- `security.kafka.ssl.client.brokerValidationSecret`

RafalSkolasinski:
I will be finalizing this within the https://github.com/SeldonIO/seldon-core/pull/4916 to keep the view on changes here unaffected so we can close this one

cliveseldon:
closing as new PR will supersede this

PR #4660, opened 2023-02-13
Add resource generic load, unload, status to CLI

cliveseldon:
- [x] Allow generic load, unload, status from stream or file

Examples:
Run kustomize to create a set of Seldon Core V2 artifacts and pipe to load:

`kustomize build config | seldon load`

Also, wait for a set of resources to be ready

`kustomize build config | seldon status -w`

A future PR can also allow for `-f` to take a folder rather than a file like kubectl.

### DIFF
--- a/operator/cmd/seldon/cli/root.go
+++ b/operator/cmd/seldon/cli/root.go
@@ -106,11 +106,16 @@ func GetCmd() *cobra.Command {
 	cmdConfigRemove := createConfigRemove()
 	cmdConfigList := createConfigList()
 
+	// Generic commands
+	cmdLoad := createLoad()
+	cmdUnload := createUnload()
+	cmdStatus := createStatus()
+
 	var rootCmd = &cobra.Command{Use: "seldon", SilenceErrors: false, SilenceUsage: true}
 
 	rootCmd.DisableAutoGenTag = true
 
-	rootCmd.AddCommand(cmdModel, cmdServer, cmdExperiment, cmdPipeline, cmdConfig)
+	rootCmd.AddCommand(cmdModel, cmdServer, cmdExperiment, cmdPipeline, cmdConfig, cmdLoad, cmdUnload, cmdStatus)
 	cmdModel.AddCommand(cmdModelLoad, cmdModelUnload, cmdModelStatus, cmdModelInfer, cmdModelMeta, cmdModelList)
 	cmdServer.AddCommand(cmdServerStatus, cmdServerList)
 	cmdExperiment.AddCommand(cmdExperimentStart, cmdExperimentStop, cmdExperimentStatus, cmdExperimentList)
--- a/operator/cmd/seldon/cli/unload.go
+++ b/operator/cmd/seldon/cli/unload.go
@@ -1,0 +1,62 @@
+package cli
+
+import (
+	"k8s.io/utils/env"
+
+	"github.com/seldonio/seldon-core/operator/v2/pkg/cli"
+	"github.com/spf13/cobra"
+)
+
+func createUnload() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "unload resources",
+		Short: "unload resources",
+		Long:  `unload resources`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			flags := cmd.Flags()
+
+			schedulerHostIsSet := flags.Changed(flagSchedulerHost)
+			schedulerHost, err := flags.GetString(flagSchedulerHost)
+			if err != nil {
+				return err
+			}
+			authority, err := flags.GetString(flagAuthority)
+			if err != nil {
+				return err
+			}
+			showRequest, err := flags.GetBool(flagShowRequest)
+			if err != nil {
+				return err
+			}
+			showResponse, err := flags.GetBool(flagShowResponse)
+			if err != nil {
+				return err
+			}
+			filename, err := flags.GetString(flagFile)
+			if err != nil {
+				return err
+			}
+			var dataFile []byte
+			if filename != "" {
+				dataFile = loadFile(filename)
+			}
+
+			schedulerClient, err := cli.NewSchedulerClient(schedulerHost, schedulerHostIsSet, authority)
+			if err != nil {
+				return err
+			}
+
+			err = schedulerClient.Unload(dataFile, showRequest, showResponse)
+			return err
+		},
+	}
+
+	flags := cmd.Flags()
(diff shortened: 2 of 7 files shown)


operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+
+func getNextResource(dec *yaml2.YAMLOrJSONDecoder) (string, string, []byte, bool, error) {
+	unstructuredObject := &unstructured.Unstructured{}
+	if err := dec.Decode(unstructuredObject); err != nil {
+		if errors.Is(err, io.EOF) {
+			return "", "", nil, false, nil
+		}
+		return "", "", nil, false, err
+	}
+	// Get the resource kind nad original bytes
```
sakoush, line 32:
```suggestion
	// Get the resource kind and original bytes
```

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(data) > 0 {
+		reader = bytes.NewReader(data)
+	} else {
+		reader = io.Reader(os.Stdin)
+	}
+	dec := yaml2.NewYAMLOrJSONDecoder(reader, 10)
+	return dec
+}
+
+func getNextResource(dec *yaml2.YAMLOrJSONDecoder) (string, string, []byte, bool, error) {
```
sakoush, line 24: too many returns and it is hard to know what we are getting back. Consider adding at least description.

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
+package cli
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"os"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	yaml2 "k8s.io/apimachinery/pkg/util/yaml"
```
sakoush, line 10: does it have to be `yaml2`?

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return "", "", nil, false, err
+	}
+	return gvk.Kind, unstructuredObject.GetName(), data, true, nil
+}
+
+func (sc *SchedulerClient) Load(data []byte, showRequest bool, showResponse bool) error {
+	dec := createDecoder(data)
+	for {
+		kind, _, data, keepGoing, err := getNextResource(dec)
```
sakoush, line 44: we are also not checking `err` before the switch

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+		if !keepGoing {
+			return err
+		}
+		switch kind {
```
sakoush, line 48: should we add a warning if it is not part of these 3 resources? should we stop?

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			return err
+		}
+	}
+}
+
+func (sc *SchedulerClient) Unload(data []byte, showRequest bool, showResponse bool) error {
+	dec := createDecoder(data)
+	for {
+		kind, _, data, keepGoing, err := getNextResource(dec)
```
sakoush, line 65: ditto

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+				waitCondition = "ModelAvailable"
+			}
+			err = sc.ModelStatus(name, showRequest, showResponse, waitCondition)
+		case "Pipeline":
+			if wait {
+				waitCondition = "PipelineReady"
+			}
+			err = sc.PipelineStatus(name, showRequest, showResponse, waitCondition)
+		case "Experiment":
+			err = sc.ExperimentStatus(name, showRequest, showResponse, wait)
```
sakoush, line 103: no wait on Experiments?

operator/pkg/cli/generic.go
```diff
@@ -0,0 +1,109 @@
...
+func (sc *SchedulerClient) Status(data []byte, showRequest bool, showResponse bool, wait bool) error {
+	dec := createDecoder(data)
+	for {
+		kind, name, _, keepGoing, err := getNextResource(dec)
+		if !keepGoing {
+			return err
+		}
+		waitCondition := ""
+		switch kind {
+		case "Model":
```
sakoush, line 92: consider extracting the values of the switch into enum as it is used in 3 different places already.

operator/cmd/seldon/cli/unload.go
```diff
@@ -0,0 +1,62 @@
...
+			if filename != "" {
+				dataFile = loadFile(filename)
+			}
+
+			schedulerClient, err := cli.NewSchedulerClient(schedulerHost, schedulerHostIsSet, authority)
+			if err != nil {
+				return err
+			}
+
+			err = schedulerClient.Unload(dataFile, showRequest, showResponse)
```
sakoush, line 49: question: why we are calling `Unload` here?

operator/cmd/seldon/cli/root.go
```diff
@@ -106,11 +106,16 @@ func GetCmd() *cobra.Command {
...
+	cmdLoad := createLoad()
+	cmdUnload := createUnload()
+	cmdStatus := createStatus()
+
 	var rootCmd = &cobra.Command{Use: "seldon", SilenceErrors: false, SilenceUsage: true}
 
 	rootCmd.DisableAutoGenTag = true
 
-	rootCmd.AddCommand(cmdModel, cmdServer, cmdExperiment, cmdPipeline, cmdConfig)
+	rootCmd.AddCommand(cmdModel, cmdServer, cmdExperiment, cmdPipeline, cmdConfig, cmdLoad, cmdUnload, cmdStatus)
```
sakoush, line 118: just to keep consistency with previous usage (i.e. `seldon model load` and `seldon pipeline load`), an alternative is perhaps to use 
`seldon all load`? I am personally not entirely sure which approach is better so happy to leave it as is.

operator/cmd/seldon/cli/unload.go
cliveseldon: not sure I follow - this is the unload method

operator/cmd/seldon/cli/root.go
cliveseldon: seems a bit verbose - maybe will leave for now

sakoush (approved):
LGTM

PR #4674, opened 2023-02-18
Allow for Kafka topic prefix customization

### LINKED ISSUE #4663: Allow users to configure Kafka topic prefix
In some contexts, users may be required to specify a Kafka topic prefix different than the current `seldon.*` default. 

To avoid this becoming a blocker, it may be useful to expose the Kafka topic prefix as a new field in the `seldon-core-v2-setup` Helm chart values.

cliveseldon:

Fixes #4663 

### DIFF
--- a/scheduler/Dockerfile.scheduler
+++ b/scheduler/Dockerfile.scheduler
@@ -1,13 +1,13 @@
-FROM golang:1.19.3-alpine as builder
+FROM golang:1.19.3-buster as builder
 
 WORKDIR /build
 COPY . .
 
 # Build the binary
-RUN apk add --no-cache make
+RUN apt-get install make
 RUN make -C scheduler -o test-go build-scheduler
 
-# Copy binary
+# Kafka dependencies necessitate leaving CGo enabled and using a base image with C dependencies
 FROM registry.access.redhat.com/ubi9/ubi-micro
 COPY --from=builder /build/scheduler/bin/scheduler /bin/scheduler
 ENTRYPOINT ["/bin/scheduler"]
--- a/scheduler/config/kafka-internal.json
+++ b/scheduler/config/kafka-internal.json
@@ -1,4 +1,5 @@
 {
+    "topicPrefix":"foo",
     "bootstrap.servers":"kafka:9093",
     "consumer":{
 	"session.timeout.ms":6000,
--- a/scheduler/pkg/kafka/topics.go
+++ b/scheduler/pkg/kafka/topics.go
@@ -26,35 +26,40 @@ import (
 )
 
 const (
-	seldonTopicPrefix = "seldon"
-	modelTopic        = "model"
-	pipelineTopic     = "pipeline"
-	errorsTopic       = "errors"
-	inputsSuffix      = "inputs"
-	outputsSuffix     = "outputs"
-	errorsSuffix      = "errors"
-	TopicErrorHeader  = "seldon-pipeline-errors"
+	defaultSeldonTopicPrefix = "seldon"
+	modelTopic               = "model"
+	pipelineTopic            = "pipeline"
+	errorsTopic              = "errors"
+	inputsSuffix             = "inputs"
+	outputsSuffix            = "outputs"
+	errorsSuffix             = "errors"
+	TopicErrorHeader         = "seldon-pipeline-errors"
 )
 
 type TopicNamer struct {
-	namespace string
+	namespace   string
+	topicPrefix string
 }
 
-func NewTopicNamer(namespace string) *TopicNamer {
+func NewTopicNamer(namespace string, topicPrefix string) *TopicNamer {
 	if namespace == "" {
 		namespace = "default"
 	}
+	if topicPrefix == "" {
+		topicPrefix = defaultSeldonTopicPrefix
+	}
 	return &TopicNamer{
-		namespace: namespace,
+		namespace:   namespace,
+		topicPrefix: topicPrefix,
 	}
 }
 
 func (tn *TopicNamer) GetModelErrorTopic() string {
-	return fmt.Sprintf("%s.%s.%s.%s", seldonTopicPrefix, tn.namespace, errorsTopic, errorsSuffix)
+	return fmt.Sprintf("%s.%s.%s.%s", tn.topicPrefix, tn.namespace, errorsTopic, errorsSuffix)
 }
 
 func (tn *TopicNamer) GetKafkaModelTopicRegex() string {
-	return fmt.Sprintf("^%s.%s.%s.*.%s", seldonTopicPrefix, tn.namespace, modelTopic, inputsSuffix)
+	return fmt.Sprintf("^%s.%s.%s.*.%s", tn.topicPrefix, tn.namespace, modelTopic, inputsSuffix)
 }
 
 func (tn *TopicNamer) GetModelNameFromModelInputTopic(topic string) (string, error) {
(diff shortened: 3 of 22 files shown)


scheduler/config/kafka-internal.json
```diff
@@ -1,4 +1,5 @@
 {
+    "topicPrefix":"foo",
```
sakoush, line 2: should this be `foo`?  and is there an easier way of setting this up for compose?

cliveseldon: Good catch - That's a bug from testing - will revert

sakoush (approved):
LGTM.

Should we add a note in the docs to explain how this can be set?

apis/go/mlops/chainer/chainer.pb.go
```diff
@@ -448,6 +448,7 @@ type PipelineTopic struct {
 
 	PipelineName string `protobuf:"bytes,1,opt,name=pipelineName,proto3" json:"pipelineName,omitempty"`
 	TopicName    string `protobuf:"bytes,2,opt,name=topicName,proto3" json:"topicName,omitempty"`
+	Tensor       bool   `protobuf:"varint,3,opt,name=tensor,proto3" json:"tensor,omitempty"`
```
sakoush, line 451: nit: rename to `isReferenceTensor`?

scheduler/pkg/kafka/topics.go
```diff
@@ -50,51 +50,54 @@ func NewTopicNamer(namespace string, topicPrefix string) *TopicNamer {
 	}
 	return &TopicNamer{
 		namespace:   namespace,
-		topicPrefix: topicPrefix,
+		topicPrefix: fmt.Sprintf("%s.", topicPrefix),
```
sakoush, line 53: why are we adding a dot here? and removing the dot from string formatter in `GetModelErrorTopic` etc.? easier to construct a regex?

scheduler/pkg/kafka/topics.go
```diff
@@ -50,51 +50,54 @@ func NewTopicNamer(namespace string, topicPrefix string) *TopicNamer {
...
 }
 
 func (tn *TopicNamer) GetModelErrorTopic() string {
-	return fmt.Sprintf("%s.%s.%s.%s", tn.topicPrefix, tn.namespace, errorsTopic, errorsSuffix)
+	return fmt.Sprintf("%s%s.%s.%s", tn.topicPrefix, tn.namespace, errorsTopic, errorsSuffix)
 }
 
 func (tn *TopicNamer) GetKafkaModelTopicRegex() string {
-	return fmt.Sprintf("^%s.%s.%s.*.%s", tn.topicPrefix, tn.namespace, modelTopic, inputsSuffix)
+	return fmt.Sprintf("^%s%s.%s.*.%s", tn.topicPrefix, tn.namespace, modelTopic, inputsSuffix)
```
sakoush, line 62: maybe add a note that `topicPrefix` as a dot suffix?

apis/go/mlops/chainer/chainer.pb.go
sakoush: or consider setting `Tensor` as string and being optional?

scheduler/pkg/kafka/topics_test.go
```diff
@@ -39,9 +39,23 @@ func TestGetModelTopic(t *testing.T) {
...
+			name:          "model with custom prefix",
+			pipelineName:  "test",
+			topicNamer:    NewTopicNamer("default", "foo.bar"),
+			stepReference: "model1.inputs",
+			expected:      "foo.bar.default.model.model1.inputs",
+		},
+		{
+			name:          "model with custom prefix with a dot",
+			pipelineName:  "test",
+			topicNamer:    NewTopicNamer("default", "foo.bar."),
```
sakoush, line 56: nit: it is perhaps from a user experience point of view that is the prefix is set to `foo.bar.` then it is likely that the user is not requiring an extra dot in the full topic name.

sakoush (approved):
LGTM. I left some minor comments / questions.

samples/examples/income_classifier/train.py
```diff
@@ -1,3 +1,17 @@
+# Copyright 2023 Seldon Technologies Ltd.
```
agrski, line 1: :thought_balloon: We should add a GitHub Action that checks if any (code) files do not include this header, so it has to be present whenever a new file is introduced

scheduler/Dockerfile.scheduler
```diff
@@ -1,13 +1,13 @@
...
 WORKDIR /build
 COPY . .
 
 # Build the binary
-RUN apk add --no-cache make
+RUN apt-get install make
 RUN make -C scheduler -o test-go build-scheduler
 
-# Copy binary
+# Kafka dependencies necessitate leaving CGo enabled and using a base image with C dependencies
```
agrski, line 10: :question: I'm probably missing something, but why does the _scheduler_ need to have any interaction with Kafka?

scheduler/pkg/kafka/topics.go
agrski: Likewise find this a bit confusing, as I'd expect to see the delimiters all together for clarity rather than having one over here

agrski (commented):
I think I _mostly_ understand how this PR approaches things, at least at a reasonably high level, but a couple of questions (and agree with @sakoush's comments too!)

scheduler/pkg/kafka/topics.go
cliveseldon: We want to allow any prefix and then add a dot which is our separator. At present, don't check if their prefix ends in a dot as doesn't really matter.

samples/examples/income_classifier/train.py
cliveseldon: Agree. Will update any missing for this PR

scheduler/pkg/kafka/topics_test.go
cliveseldon: It seemed simpler not to even check and allow this.

scheduler/Dockerfile.scheduler
cliveseldon: As the Kafka Config contains the prefix and that has a reference to the Kafka library.

scheduler/pkg/kafka/topics_test.go
cliveseldon: actions:
 - Clarify valid characters for Kafka topic
 - Update docs to suggest not having dot at end

scheduler/Dockerfile.scheduler
cliveseldon: future PR to allow conversion of JSON to KafkaConfig to separate package so Scheduler does not need Kafka library

scheduler/pkg/kafka/topics.go
cliveseldon: actions:
 - Helper function for joining topic names

scheduler/pkg/kafka/topics.go
```diff
@@ -164,6 +175,15 @@ func (tn *TopicNamer) CreateStepReferenceFromPipelineInput(pipelineInputs string
 // 1. step
 // 2. step.(inputs|outputs)
 // 3. step.(inputs|outputs).tensor
-func (tn *TopicNamer) IsReferenceTensor(reference string) bool {
-	return len(strings.Split(reference, ".")) > 2
+// Return the repference with tensor if there is one else just return reference
```
sakoush, line 178:
```suggestion
// Return the reference with tensor if there is one else just return reference
```

sakoush (approved):
LGTM, just a minor typo in comment.

PR #4675, opened 2023-02-19
Ensure settings for max message size

### LINKED ISSUE #4668: The max message size for kafka is set only on pipeline topics and not on the respective model topics
## Describe the bug
In Kubernetes Deployment, we are able to specify the "message.max.bytes" property for Kafka topics created by seldon-core-v2. It seems, that this property is only set on the topics for the pipelines, but not for the models, yielding inconsistencies. If one of the models produces a message >1MB (our current broker default) the request gets stuck without any error indicating, what's happening.

## To reproduce

1. Create models
2. Create a pipeline
3. Examine the created Kafka topics. The Pipeline topics contains the "message.max.bytes" property, the model ones doesn't

## Expected behaviour
I'd expect the property to be consistently set for all topics created by Kafka.

## Environment

* Cloud Provider: AWS (EKS + MSK)
* Kubernetes Cluster Version: 1.24
* Deployed Seldon System Images: all images for seldon core v2.2.0

cliveseldon:
- Ensures admin client that creates topics uses consumer config and thus inherits `message.max.bytes`. This ensures that when topics are created they reflect the message bytes needed for them.
- Adds some missing settings to dataflow engine to allow messages of size `message.max.bytes` to be succsefully transmitted. This includes `buffer.memory` setting which fails if messages are sent of large size for `message.max.bytes`.
- Update CLI to handle large messages in `seldon pipeline inspect`
- Update MLServer max grpc message size in docker and k8s.
- Update agent grpc dial options.

Tested via CIFAR10 example with a batch size of 10000.

Fixes #4668


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
```diff
@@ -139,10 +141,13 @@ fun getKafkaProperties(params: KafkaStreamsParams): KafkaProperties {
         this[ConsumerConfig.AUTO_OFFSET_RESET_CONFIG] = "latest"
         this[ConsumerConfig.MAX_PARTITION_FETCH_BYTES_CONFIG] = KAFKA_MAX_MESSAGE_BYTES
         this[ConsumerConfig.FETCH_MAX_BYTES_CONFIG] = KAFKA_MAX_MESSAGE_BYTES
+        this[ConsumerConfig.SEND_BUFFER_CONFIG] = KAFKA_MAX_MESSAGE_BYTES
+        this[ConsumerConfig.RECEIVE_BUFFER_CONFIG] = KAFKA_MAX_MESSAGE_BYTES
         this[ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG] = 60000
 
         this[ProducerConfig.LINGER_MS_CONFIG] = 0
         this[ProducerConfig.MAX_REQUEST_SIZE_CONFIG] = KAFKA_MAX_MESSAGE_BYTES
+        this[ProducerConfig.BUFFER_MEMORY_CONFIG] = KAFKA_MAX_MESSAGE_BYTES
```
sakoush, line 150: not part of this PR, but why do we need to set steams, producer and consumer configs with the same values?

sakoush (approved):
LGTM

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
cliveseldon: Probably does not further validation only a subset is needed but I assume they are used by different parts of kstreams for setting tings up.

PR #4683, opened 2023-02-24
Ansible refactor

RafalSkolasinski:
Analogous to https://github.com/SeldonIO/seldon-core/pull/4681


cliveseldon:
Should we update the docs if this changes any preferred install mechanisms?

RafalSkolasinski:
Ansible is still more or less the same - we removed dependency on https://github.com/SeldonIO/ansible-k8s-collection/ but these continue to be the same playbooks. 

I can do as follow up to both (this and V2 one) make sure that regular docs reference this properly and contains a proper warning that this is dev / trial automation but for production recommended is Helm chart instalaltion

cliveseldon (approved):

PR #4694, opened 2023-02-28
Add note that Ansible is not for production

RafalSkolasinski:
Add note that Ansible is not for production


agrski:
Duplicated by #4695 ?

edshee (changes requested):
Could we add a link to the helm installation page?

edshee:
> Duplicated by #4695 ?

Yes both seem to be merging to v2, did you mean to do one for v2 and one for v1?

RafalSkolasinski:
Yeah, my fault

PR #4695, opened 2023-02-28
Add note that Ansible is not for production

RafalSkolasinski:
Add note that Ansible is not for production


edshee (approved):

RafalSkolasinski:
> Only suggestion is to use one sentence per line and one line per sentence, but looks good.

Not sure it renders properly then in the `> ...` block.

RafalSkolasinski:
Done. I don't think I was able to do `../helm.md` as Ansible's `readme.md` is at different location then rendered in docs.

edshee (approved):

PR #4706, opened 2023-03-06
Add model server controlplane interface

sakoush:

This PR adds an interface for  control plane with a specific implementation of the Open Inference Protocol (v2). This allows then other plugable implementations for the control plane for backends that do not support the v2 protocol (e.g. TorchServe).

This PR also adds a couple of refactors:
- Introduce internal packages for testing utils
- Remove Rest support for current v2 client, we only use Grpc in production
- Add Rest v2 client as a client for testing purposes only

Fixes #


cliveseldon:
Looks good - please run notebook tests locally and on k8s

cliveseldon (approved):

sakoush:
> Looks good - please run notebook tests locally and on k8s

done.

PR #4713, opened 2023-03-07
Increase idle connection of REST proxy to timeout to 30 minutes

sakoush:

Increase default idle timeout of REST proxy to 30 minutes. This is the connection between agent and model server. Otherwise for a single inference requests that takes more than the default threshold (previously 1 minutes) the connection drops.

Fixes #


agrski (approved):

adriangonz:
How hard would it be to make this configurable?

sakoush:
> How hard would it be to make this configurable?

not hard but it is probably a set of constants that we need to make configurable not just this one. So will leave to a future PR perhaps,

PR #4717, opened 2023-03-09
Allow jaeger image to be customized

cliveseldon:
Allow jaeger image to be set when we start seldon core v2 docker compose


sakoush (approved):
LGTM

Issue #4689, opened 2023-02-27
missing "serviceStatus" field in "status" of sdep resource


caozhuozi:
## Describe the bug

We want to use  the `serviceStatus` field of `status` of a `sdep` resource to get the svc info of all predictors. But sometimes it is missing in the sdep resource. Could you please help clarify when this field will be shown and when will not?

## Expected behaviour

The `serviceStatus` field SHOULD  be shown in `status` if the sdep is created successfully.

## Environment

I think it is environment independent.
seldon-core version: 1.14.0

## Model Details 
I think it is model-independent.

cliveseldon:
As discussed on community slack - this seems not to be an issue. Please reopen if its still an issue.

caozhuozi:
Let me share my findings here.
(The line numbers given following are based on the master branch of [seldondeployment_controller.go](https://github.com/SeldonIO/seldon-core/blob/ffa8d64a03f7b38686c3eb5966d6a7f7f53928b2/operator/controllers/seldondeployment_controller.go)).
### related variables:

the **wanted** `svc`(the svc that seldon-core wants to create/update. I don't use the word “**desired**” here as there is also a related variable called `desiredSvc`): first appears at line 1241

the `found` one(the svc may already exist, but it may have some differences from the **wanted** one): first appears at line 1245

`desiredSvc` (which is actually a deep copy of the `found` one): first appear sat line 1279

### The case I encountered that would result in the difference between the `found` one and the **wanted** `svc`:

When creating a new svc, there will be also a  `clusterIPs` field (and some other fields) set by default by my Kubernetes(v1.22.17-gke.3100). This field does not exist in the **wanted** `svc` but appears in the `found` one.

 
### The above difference may lead to infinite re-updating in every reconcile: 

Because of the differences between the `found` one and the **wanted** `svc`, the `equality.Semantic.DeepEqual` check at line 1262 will return false as expected.

[...]

PR #4725, opened 2023-03-11
fix(seldon operator): infinite service updating if extra default fileds applied by k8s

caozhuozi:
This PR tries to fix issue https://github.com/SeldonIO/seldon-core/issues/4689. The issue and the cause are detailed described there.

Fixes (https://github.com/SeldonIO/seldon-core/issues/4689)

1. As I metioned in issue  (https://github.com/SeldonIO/seldon-core/issues/4689), the root cause of the infinite updating services in every reconcile is due to the extra fileds applied by k8s. To solve this, I enhanced the check statement by just comparing those required fields and ignoring the default fields  applied by k8s when comapring the desired svc and the found one. Specifically,

   * Since several fields need to be compared/updated, I organized the check/update logic into  separate functions.
   * As the code has already ignored the extra fields set by k8s,  I also removed the [line 1261](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1261) `svc.Spec.ClusterIP = found.Spec.ClusterIP`.
2. I think there is also a potential bug at [line 1282](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1282). The deep copy should be performed **after** the update to `found`. Otherwise, the comparing logic at [line 1293](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1293) is meaningless. Why we re campare the old found svc(stored in `desiredSvc`) with the updated one(`found` is subsequently updated **twice** at [line 1284-1285](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1284-L1285) and [line 1287](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1287) sperately) given that we already know they are not the same? I think this behavior also conflicts with what the code comment at [line 1292](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1293) intends to do: 
    > // Check if what came back from server modulo the defaults applied by k8s is the same or not
    
    In my opinion, "what came back from server" refers to the second update to `found` at [line 1287](https://github.com/SeldonIO/seldon-core/blob/8004de80ec447d5337d78ef630328584b3177829/operator/controllers/seldondeployment_controller.go#L1287), that is, 
    ```go
    err = r.Update(context.TODO(), found)
    ```
    If I understand correctly, the comment originally wants to compare the desired Service  just before sending update request to the API server and the result back from the server.
    By moving the deep copy statement just after the first update to `found` and before the second update, I think the comapring statement now matches the what the above comment wants to do as well as the variable name itself: **desired**Svc( I was very confused about why the deep copy of the old found one is called `desiredSvc`).
    Please corrrect me if I'm wrong. Thanks.

### DIFF
--- a/operator/controllers/seldondeployment_controller.go
+++ b/operator/controllers/seldondeployment_controller.go
@@ -866,6 +866,39 @@ func createContainerService(deploy *appsv1.Deployment,
 	return svc
 }
 
+// compareServices ignores those fields set by k8s by default.
+// Specifically, the fields to be compared are:
+//   - Annotations
+//   - Labels
+//   - Spec.Ports
+//   - Spec.Type
+//   - Spec.Selector
+//   - Spec.SessionAffinity
+func compareServices(desiredSvc, found *corev1.Service) bool {
+	return equality.Semantic.DeepEqual(desiredSvc.Annotations, found.Annotations) &&
+		equality.Semantic.DeepEqual(desiredSvc.Labels, found.Labels) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.Ports, found.Spec.Ports) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.Selector, found.Spec.Selector) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.Type, found.Spec.Type) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.SessionAffinity, found.Spec.SessionAffinity)
+}
+
+// updateService ignores those immutable fields and fields set by k8s by default.
+// Specifically, the updated fields are:
+//   - Annotations
+//   - Labels
+//   - Spec.Ports
+//   - Spec.Type
+//   - Spec.SessionAffinity
+func updateService(foundSvc, desiredSvc *corev1.Service) *corev1.Service {
+	foundSvc.Annotations = desiredSvc.Annotations
+	foundSvc.Labels = desiredSvc.Labels
+	foundSvc.Spec.Ports = desiredSvc.Spec.Ports
+	foundSvc.Spec.Type = desiredSvc.Spec.Type
+	foundSvc.Spec.SessionAffinity = desiredSvc.Spec.SessionAffinity
+	return foundSvc
+}
+
 func addModelEnvs(pu *machinelearningv1.PredictiveUnit, con *corev1.Container) {
 	if len(pu.Parameters) > 0 {
 		// Set V1 env vars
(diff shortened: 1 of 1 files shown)


seldondev:
Hi @caozhuozi. Thanks for your PR.

<details>

</details>

operator/controllers/seldondeployment_controller.go
```diff
@@ -866,6 +866,39 @@ func createContainerService(deploy *appsv1.Deployment,
...
+//   - Spec.Type
+//   - Spec.Selector
+//   - Spec.SessionAffinity
+func compareServices(desiredSvc, found *corev1.Service) bool {
+	return equality.Semantic.DeepEqual(desiredSvc.Annotations, found.Annotations) &&
+		equality.Semantic.DeepEqual(desiredSvc.Labels, found.Labels) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.Ports, found.Spec.Ports) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.Selector, found.Spec.Selector) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.Type, found.Spec.Type) &&
+		equality.Semantic.DeepEqual(desiredSvc.Spec.SessionAffinity, found.Spec.SessionAffinity)
```
cliveseldon, line 883: There are several other fields that can be set in the spec by the user: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#servicespec-v1-core
Have you validated none of these is applicable? Some have defaults and are usually not set so I suppose one would need to be careful if as we are doing now we are not copying the spec back.

caozhuozi: @cliveseldon, Thanks for you quick review. Do you mean in addition to the defaults set by Kubernetes, we also need to consider that the users may (intentionally or unintentionally) modify the services directly?

cliveseldon: Actually, as a user has very little control from  SeldonDeployment on the created service we can ignore fields like `externalIPs` etc as a user can't control these from Seldon.

caozhuozi: @cliveseldon. If we can ignore the impact of users, can I say that the change of svc fields are controlled by those defaults by Kubernetes and seldon operator? All the fields I selected to comapre/update are extracted from _createContainerServices_ and _ceratePredictorServices_ which are required by seldon operator side. I think keeping the defaults set by Kubernetes shouldn't bring side effects. Could you please give specific situations that we have to take other fields into account?

cliveseldon: No I think that is fine if we have covered all the items in the Spec that we control

cliveseldon: I think its ok as is

cliveseldon (approved):

PR #4737, opened 2023-03-15
 use Debian 11 for Executor & Operator images

RafalSkolasinski:

This is to address following CVEs:
- cve-2021-33574
- cve-2021-35942
- cve-2022-23218
- cve-2022-23219

present both in Executor and Operator images.

Fixes #


RafalSkolasinski:
integration passed here 
image

Issue #4770, opened 2023-03-30
artifactVersion in model spec does not redirect to subfolder


kevinnowland:
The `artifactVersion` option in the model spec does not appear to be working as expected.

I deployed a model with the following folder structure to a triton model server:
```
char_count/config.pbtxt
char_count/1/model.py
char_count/2/model.py
```

using a `model.yaml` that looked like

```yaml
apiVersion: mlops.seldon.io/v1alpha1
kind: Model
metadata:
  name: char-count
  namespace: seldon-mesh
spec:
  storageUri: "/var/models/char_count"
  artifactVersion: 1
  requirements:
    - triton
```

with the expectation that `char_count/1/model.py` would be loaded onto the triton server. However, the file `/mnt/agent/models/char-count_1/1/model.py` on the triton server was that of `char_count/2/model.py`.

https://github.com/SeldonIO/seldon-core/blob/ad4fce903fb1d763ba37b7c9d56b7ae67cb7ca82/operator/apis/mlops/v1alpha1/model_types.go#L103

cliveseldon:
Closed as #4774 fixes this.

kevinnowland:
Thank you!

PR #4774, opened 2023-03-31
Fix model artifact versions

cliveseldon:
Fix artifact version is not passed to scheduler from Model yaml.

Fixes: #4770

### DIFF
--- a/docs/source/contents/examples/index.md
+++ b/docs/source/contents/examples/index.md
@@ -11,6 +11,7 @@ This section will provide some examples to allow operations with Seldon to be te
 
  * [Huggingface models](huggingface.md)
  * [Model zoo](model-zoo.md)
+ * [Artifact versions](multi-version.md)
 
 ## Pipelines
 
@@ -54,6 +55,7 @@ local-examples.md
 k8s-examples.md
 huggingface.md
 model-zoo.md
+multi-version.md
 pipeline-examples.md
 pipeline-to-pipeline.md
 explainer-examples.md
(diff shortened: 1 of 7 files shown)


samples/k8s-multi-version.md
```diff
@@ -0,0 +1,178 @@
...
+  storageUri: "gs://seldon-models/scv2/samples/triton_22-11/multi-version"
+  artifactVersion: 1
+  requirements:
+  - triton
+  - python
+
+```
+
+```bash
+kubectl create -f ./models/multi-version-1.yaml -n ${NAMESPACE}
```
agrski, line 53: :thought_balloon: Probably worth using `apply` here as we expect to update this later; I realise this is generated from the notebook

docs/source/contents/examples/index.md
```diff
@@ -54,6 +55,7 @@ local-examples.md
 k8s-examples.md
 huggingface.md
 model-zoo.md
+multi-version.md
```
agrski, line 58: :question: Has this file been added?

agrski (commented):
One important question and a minor suggestion, but generally looks good!

I've also checked that the other parts of the model spec are provided in the adaptor between k8s resource and scheduler spec.

docs/source/contents/examples/index.md
cliveseldon: will add!

agrski (approved):
Checked this renders fine in docs - looks good!

PR #4779, opened 2023-04-04
feat(dataflow): Add missing filter for requests without input tensors

cliveseldon:
Missing filter for joining pipelines when there are no tensor matches.


agrski (approved):
Good spot!

:question: Do we need an equivalent for outputs/responses?

PR #4785, opened 2023-04-05
Use Istio gateway in server_examples notebook

adriangonz:

It seems like Ambassador has a small delay between the pod coming up and actually start routing traffic there. Which means that sending an inference request immediately after the pod becomes ready can sometimes lead to a `404` - thus causing flakiness on some of the integration tests.

This PR changes one of the notebooks used in the tests to use the Istio gateway instead (which seems slightly more stable?). 

Fixes #

### DIFF
--- a/notebooks/server_examples.ipynb
+++ b/notebooks/server_examples.ipynb
@@ -45,7 +45,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 20,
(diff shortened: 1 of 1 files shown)


adriangonz:
Notebooks seem to pass, so hopefully this PR makes them a bit more stable

RafalSkolasinski (approved):
Nice one!

PR #4787, opened 2023-04-06
Bump MLServer to 1.3.0 (V2)

adriangonz:
Bump MLServer to `1.3.0`

Fixes #

We'll run tests first with RCs and then convert it from draft once all looks good.


samples/scripts/models/Makefile
```diff
@@ -66,7 +66,7 @@ income-xgb: env
 
 .PHONY: upload-income-xgb
 upload-income-xgb:
-	gsutil cp income-xgb/model.json gs://seldon-models/scv2/samples/${MLSERVER_FOLDER}/income-xgb/model.json
+	gsutil cp income-xgb/model.bst gs://seldon-models/scv2/samples/${MLSERVER_FOLDER}/income-xgb/model.bst
```
cliveseldon, line 69: Why the change to bst?

adriangonz: Not sure - I haven't checked why yet, but that was the artefact output by the training step. 

I'll have a deeper look once I get a chance.

adriangonz: Just to make sure we've got the answer documented here, the training script saves it as `model.bst`: 

https://github.com/SeldonIO/seldon-core/blob/3de894a7338b1aeab86c3239793318c8407294ee/samples/scripts/models/income-xgb/train.py#L72

It would still be good to save it as a `*.json` instead - which is XGBoost's newer format - but I don't want to risk more changes before the release :grimacing:

cliveseldon (approved):

PR #4791, opened 2023-04-10
Fix python grpc api build

cliveseldon:
* Creates python folder
 * Creates grpc python bindindings
 * Uses the python grpc-protoc needed for python rather than the usual protoc binary


agrski (commented):
I've tested this with:
```bash
cd apis
python3 -m venv venv
source venv/bin/activate
pip install grpcio-tools
make build-v2-python
```
I've confirmed this generates sensible-looking outputs.

---

:thought_balloon: I've made a couple of suggestions for the Makefile.

:question: Separately, elsewhere we've added the generated clients into git.  Should we do that here as well?

apis/Makefile
```diff
@@ -41,18 +41,20 @@ clean-go:
...
+
 .PHONY: build-v2-python
-build-v2-python:
+build-v2-python: python-deps
 	mkdir -p python
 	python -m grpc_tools.protoc \
 		$(PROTOC_IMPORT_PATH) \
 		$(PROTOC_PYTHON_OPTIONS) \
-		--grpc_python_out=./python \
 		--pyi_out=./python \
```
agrski, line 58: :wrench: Same here about duplicated info!

agrski (approved):
One more suggested change, but approving for quick merge thereafter

PR #4805, opened 2023-04-21
Quickfix/use tcmalloc

sakoush:

This change allows the use of `tcmalloc` in triton for memory allocations. Based on the triton docs [here](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_management.md) this is the recommendation if we do a lot of loading/unloading of models (which is the typical case in SCv2 especially with overcommit). 

Fixes #


sakoush:
tested with `samples/local-over-commit-test.ipynb` (triton) in docker-compose and kind.

cliveseldon (approved):

PR #4811, opened 2023-04-22
[WIP] feat(dataflow): Add Prometheus JMX exporter Java agent

agrski:

Fixes #

This is still a work in progress and should not be merged in its current form.  It's an early draft of the PR to enable collaboration around collecting JVM metrics.

### DIFF
--- a/scheduler/Dockerfile.dataflow
+++ b/scheduler/Dockerfile.dataflow
@@ -31,4 +31,4 @@ ADD --chmod=0664 https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus
 RUN echo -e 'rules:\n- pattern: ".*"\n' > config.yaml
 
 # TODO - inject JAR name & version from build pipeline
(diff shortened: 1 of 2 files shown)


sakoush:
btw `chmod` requires buildkit

```
{
  "features": {
    "buildkit": true
  }
}
```

agrski:
> btw `chmod` requires buildkit
> 
> ```
> {
>   "features": {
>     "buildkit": true
>   }
> }
> ```

Ah yes.  I have buildkit enabled by default, so tend to forget which bits require it vs. don't.

Any suggestions on ensuring this is clear to users and for CI?

RafalSkolasinski:
> btw chmod requires buildkit

we can just have a simple `RUN chmod ...` step to not introduce requirement for buildkit if we do not want it.

sakoush:
> > btw chmod requires buildkit
> 
> we can just have a simple `RUN chmod ...` step to not introduce requirement for buildkit if we do not want it.

actually for myself I can only set the permissions using `RUN chmod ...` maybe I have an outdated buildkit

PR #4815, opened 2023-04-24
Use a more recent jdk base image

sakoush:

Update `dataflow-engine` base jdk image to use a more recent one to sort out CVEs. 

At the time of writing:
image

Fixes #

### DIFF
--- a/scheduler/Dockerfile.dataflow
+++ b/scheduler/Dockerfile.dataflow
@@ -14,7 +14,7 @@ RUN gradle build --no-daemon --info
 ################################################################################
 
 # Some dependencies require glibc, which Alpine does not provide
-FROM registry.access.redhat.com/ubi9/openjdk-17-runtime:1.14-2
+FROM registry.access.redhat.com/ubi9/openjdk-17-runtime:1.14-2.1681917142
 
 # Run update to pickup any necessary security updates
 USER root


scheduler/Dockerfile.dataflow
```diff
@@ -14,7 +14,7 @@ RUN gradle build --no-daemon --info
 ################################################################################
 
 # Some dependencies require glibc, which Alpine does not provide
-FROM registry.access.redhat.com/ubi9/openjdk-17-runtime:1.14-2
+FROM registry.access.redhat.com/ubi9/openjdk-17-runtime:1.14-2.1681917142
```
agrski, line 17: :question: Would `latest` be preferable to the specific version?

sakoush: We have not concluded yet whether it is alright to use `latest` tag. So bumping up the version for now.

sakoush: `1.14-2` is not a rolling tag (`1.14` is).

RafalSkolasinski: I'd probably stick to `1.14` then same way as I'd use for base `ubuntu:22.04` and not specific build of it - but you are right we did not make specific decision on these yet.

agrski (approved):
We can look at changing to `1.14` in future; this is fine for now

PR #4827, opened 2023-05-03
fix(helm): fix dataflow using non-existing Helm value (brokerCa instead of brokerCaPath)

RafalSkolasinski:

Our Helm charts defines [here](https://github.com/SeldonIO/seldon-core/blob/17c37214877706f9fd485a7093c54c33cad4095b/k8s/helm-charts/seldon-core-v2-setup/values.yaml#L34)
```
security.kafka.ssl.client.brokerCaPath: /tmp/certs/kafka/broker/ca.crt
```
however Dataflow Engine seems to be using [brokerCa](https://github.com/SeldonIO/seldon-core/blob/17c37214877706f9fd485a7093c54c33cad4095b/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml#L657-L658) instead. 

Fixes #

### DIFF
--- a/ansible/playbooks/templates/values-seldon-core-v2-components.yaml.j2
+++ b/ansible/playbooks/templates/values-seldon-core-v2-components.yaml.j2
@@ -40,4 +40,4 @@ security:
         keyPath: /tmp/certs/kafka/client/user.key
         crtPath: /tmp/certs/kafka/client/user.crt
         caPath: /tmp/certs/kafka/client/ca.crt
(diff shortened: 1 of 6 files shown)


RafalSkolasinski:
Current change is for SSL configuration. Converting to draft until I also review SASL_SSL

RafalSkolasinski:
Ok, seems nothing else is required straight away.

agrski (approved):

Issue #4828, opened 2023-05-04
Enable SASL_SSL kafka protocol with PLAIN mechanism


RafalSkolasinski:
When using in example managed Kafka from Confluent Cloud, the provided config may correspond to `SASL_SSL` protocol with `PLAIN` mechanism.

```bash
# Required connection configs for Kafka producer, consumer, and admin
bootstrap.servers=<broker servers uri>:9092
security.protocol=SASL_SSL
sasl.mechanisms=PLAIN
sasl.username=<user>
sasl.password=<password>

# Best practice for higher availability in librdkafka clients prior to 1.7
session.timeout.ms=45000
```

however we currently do not support this

https://github.com/SeldonIO/seldon-core/blob/17c37214877706f9fd485a7093c54c33cad4095b/scheduler/pkg/kafka/config/tls.go#L34-L45

which leads to e.g. following errors from Pipeline Gateway microservice

```
time="2023-05-03T20:59:50Z" level=info msg="Options:prefix= validationPrefix=KAFKA_BROKER clientset=<nil>" source=CertificateStore
time="2023-05-03T20:59:50Z" level=info msg="Just looking for validation cert" source=CertificateStore
time="2023-05-03T20:59:50Z" level=info msg="Getting validation certificate for KAFKA_BROKER" source=CertificateStore
time="2023-05-03T20:59:50Z" level=fatal msg="Failed to create kafka manager" error="open /tmp/certs/kafka/broker/ca.crt: no such file or directory"
```

RafalSkolasinski:
Note: this is covering case not-covered by prior https://github.com/SeldonIO/seldon-core/issues/4415 solved by https://github.com/SeldonIO/seldon-core/pull/4416

RafalSkolasinski:
For clarification. One thing we need to include is definitely option to set `sasl.mechanism` to `PLAIN` as now it is hard-coded to `SCRAM-SHA-512`. The other thing to decide is weather we could make `ssl.client.brokerValidationSecret` optional and change base from `ubi-micro` to `ubi-minimal` such that it comes with `ca-certificates` package and can work with e.g. Confluent Kafka out of the box.

RafalSkolasinski:
[...]

PR #4831, opened 2023-05-05
feat(kafka): add support for SASL with PLAIN mechanism

RafalSkolasinski:

Adds Kafka support for SASL with PLAIN mechanism.

Fixes https://github.com/SeldonIO/seldon-core/issues/4828


RafalSkolasinski:
Tested:
- Confluent Kafka (SASL_SSL/PLAIN)
- Strimzi with kraft (mTLS)
- Strimzi with Zookeper (PLAIN + SASL_SSL/SCRAM-SHA512 + mTLS)

cliveseldon:
Looks good - can we clarify key changes as will need to update for cluster wide , is the summayr
 * name changes to envs and some new ones, particualrly mechanism
 * Kotlin changes to clean up usage for configuration
 * add SASL mechanism in a clean way to go code

cliveseldon (approved):

components/tls/pkg/tls/util.go
```diff
@@ -50,3 +55,11 @@ func GetSecurityProtocolFromEnv(prefix string) string {
 	}
 	return val
 }
+
+func GetSASLMechanismFromEnv(prefix string) string {
+	val, ok := os.LookupEnv(fmt.Sprintf("%s%s", prefix, EnvSASLMechanismSuffix))
+	if !ok {
```
agrski, line 61: :thought_balloon: We should probably check this is a supported mechanism we know how to handle

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/security/Protocols.kt
```diff
@@ -0,0 +1,9 @@
+package io.seldon.dataflow.kafka.security
```
agrski, line 1: :wrench: We need to re-run licence header generation

scheduler/pkg/kafka/config/tls.go
```diff
@@ -45,13 +47,15 @@ func AddKafkaSSLOptions(config kafka.ConfigMap) error {
 }
 
 func setupSASLSSLAuthentication(config kafka.ConfigMap) error {
-	cs, err := tls.NewCertificateStore(tls.ValidationOnly(true), tls.ValidationPrefix(EnvKafkaBrokerPrefix))
-	if err != nil {
-		return err
+	// Set the SASL mechanism
+	mechanism := tls.GetSASLMechanismFromEnv(tls.EnvSecurityPrefixKafka)
+	if (mechanism != tls.SASLMechanismPlain) && (mechanism != tls.SASLMechanismSCRAMSHA256) && (mechanism != tls.SASLMechanismSCRAMSHA512) {
```
agrski, line 52: :thought_balloon: Related to my other comment, I'd probably have the validation in the env lookup so it can return an error appropriately.  That's then consistent for all users and handles the external/domain boundary at the earliest possible convenience

scheduler/pkg/kafka/config/tls.go
```diff
@@ -63,8 +67,16 @@ func setupSASLSSLAuthentication(config kafka.ConfigMap) error {
...
 	config["sasl.password"] = ps.GetPassword()
-	// issue is that ca.pem does not work with multiple certificiates defined
+
+	// Set the TLS Certificate
+	cs, err := tls.NewCertificateStore(tls.ValidationOnly(true), tls.ValidationPrefix(EnvKafkaBrokerPrefix))
+	if err != nil {
+		return err
+	}
+	caCert := cs.GetValidationCertificate()
+	// issue is that ca.pem does not work with multiple certificates defined
```
agrski, line 77: :question: Does this comment still apply?

agrski (commented):
One change requested + a couple of suggestions

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/security/Protocols.kt
RafalSkolasinski: Which target is that?

agrski: `make update-copyright` I think, at the top level

components/tls/pkg/tls/util.go
agrski: Whatever we retrieve here; see my other comment for further clarification

RafalSkolasinski: This is currently not doing anymore than `GetSecurityProtocolFromEnv` function just next to it. 
We can include in further re-factor but for now I just have seen this one as simple helper to get the mechanism and logic for validation to be in place that we use it.

scheduler/pkg/kafka/config/tls.go
RafalSkolasinski: did not verify this specificity so I'd keep it for now

RafalSkolasinski:
there is just one more point on including the "well trusted" ca-certificates in the images under the location that we expect to find it

PR #4834, opened 2023-05-09
feat(models): Add model status message for k8s resource status

agrski:
This PR adds a `message` field for model statuses to indicate, at a high level, why a model is (not) ready.  At present, models can only be ready or not ready, and may have a `reason` describing the underlying cause of that status.  That reason is fine for humans, but is verbose and is not suitable for automated systems to parse easily.

This PR also updates the pipeline status fields to share the same semantics are for models:
* Message = short-form description of the status, generally as a term from the Protobuf contracts.
* Reason = free-form text describing the underlying cause of the status.

There are also a number of formatting changes to break long lines (>120 characters) and introduce some whitespace for logical grouping/separation.

N/A

### DIFF
--- a/operator/controllers/mlops/model_controller.go
+++ b/operator/controllers/mlops/model_controller.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	schedulerAPI "github.com/seldonio/seldon-core/apis/go/v2/mlops/scheduler"
 	mlopsv1alpha1 "github.com/seldonio/seldon-core/operator/v2/apis/mlops/v1alpha1"
 )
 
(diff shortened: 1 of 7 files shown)


operator/controllers/mlops/model_controller.go
```diff
@@ -118,8 +119,19 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
+func (r *ModelReconciler) updateStatusFromError(
+	ctx context.Context,
+	logger logr.Logger,
+	model *mlopsv1alpha1.Model,
+	canRetry bool,
+	err error,
+) {
+	modelStatus := schedulerAPI.ModelStatus_ModelFailed
+	if canRetry {
+		modelStatus = schedulerAPI.ModelStatus_ModelProgressing
```
agrski, line 131: :information_source: The equivalent pipeline code simply says the pipeline failed if there was any issue requesting it to be loaded.  Which behaviour do we want -- indicate failure early, or be optimistic as transient glitches can happen?

agrski: I've updated the pipeline status to not indicate failure from retryable errors: https://github.com/SeldonIO/seldon-core/pull/4834/commits/558b0a68cd26ae5f8188a08a4b8e36f9b796087c

cliveseldon (approved):

PR #4835, opened 2023-05-10
Fix Alibi Explainer tests

adriangonz:
Following the release of `urllib3` 2.0, the Alibi Explain Server tests have suddenly started to fail with the error below:

```
  AttributeError

  'HTTPResponse' object has no attribute 'strict'

  at ~/.local/venv/lib/python3.7/site-packages/cachecontrol/serialize.py:54 in dumps
       50│                 ),
       51│                 u"status": response.status,
       52│                 u"version": response.version,
       53│                 u"reason": text_type(response.reason),
    →  54│                 u"strict": response.strict,
       55│                 u"decode_content": response.decode_content,
       56│             }
       57│         }
       58│ 
```

This PR makes sure that the tests run with the `python-builder` image.

Fixes #

### DIFF
--- a/.github/workflows/alibiexplainer_tests.yml
+++ b/.github/workflows/alibiexplainer_tests.yml
@@ -34,7 +34,7 @@ jobs:
         uses: snok/install-poetry@v1
         with:
           version: 1.1.15
-          virtualenvs-create: false
+          virtualenvs-create: true
       - name: test-python
         run: |
           pip install --upgrade pip setuptools
--- a/.github/workflows/alibiexplainer_tests.yml
+++ b/.github/workflows/alibiexplainer_tests.yml
@@ -23,19 +23,14 @@ jobs:
 
   python-tests:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/alibiexplainer_tests.yml
```diff
@@ -24,18 +24,13 @@ jobs:
   python-tests:
 
     runs-on: ubuntu-latest
+    container: seldonio/python-builder:0.6
```
RafalSkolasinski, line 27: :wrench: if we could get away without using `seldonio/python-builder` image it'd be probably preferable and easier to maintain in long run

adriangonz: Totally agree - but that was causing some issues (probably caused by using an old version of something somewhere). Adding back `seldonio/python-builder` was the quickest way to sort it out. 

Having said so, I'm sure that if we spent some time troubleshooting it we could get it to work with the GH standard stuff.

PR #4836, opened 2023-05-10
Remove librpmio from Dockerfile.dataflow

sakoush:
remove librpmio as it is not installed anymore!

It looks like `librpmio` is not available anymore in the base image and therefore we currently fail in during `microdnf remove`. Fix is to remove it.

Fixes #


agrski (approved):

RafalSkolasinski:
Nice one, we have noticed the same just now with @agrski

PR #4837, opened 2023-05-10
Add mermaid diagrams to pipeline docs page

jklaise:
Add basic visualizations of example pipelines in the docs.

- This is a draft PR meant to discuss if the visual approach is correct. After this initial review, the intention is, if possible, to factor out these diagrams in stand-alone files and include them in the relevant places in both docs and examples.
- I've chosen to try `mermaid` as it's code based and easier to update and manage than external assets like images, produced with another tool. The trade-off is that the layout and styling options are more limited.
- There are some issues with using both `:caption:` and `:align:` directives, which may be due to the [docs theme](https://github.com/mgaitan/sphinxcontrib-mermaid/issues/90)

See relevant docs page here: https://seldon--4837.org.readthedocs.build/projects/seldon-core/en/4837/contents/pipelines/index.html

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -4,6 +4,7 @@ sphinxcontrib-applehelp==1.0.2
 sphinxcontrib-devhelp==1.0.2
 sphinxcontrib-htmlhelp==1.0.3
 sphinxcontrib-jsmath==1.0.1
+sphinxcontrib-mermaid==0.8.1
 sphinxcontrib-qthelp==1.0.3
 sphinxcontrib-serializinghtml==1.1.4
 myst-parser==0.14.0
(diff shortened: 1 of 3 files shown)


cliveseldon:
These are great - they provide an extra layer of understanding which helps

matthewlowdon:
Super! I have two minor points:

- There are a couple of implicit output joins of type 'any'. Maybe better to show them explicitly?
- tfsimple3's INPUT should be called INPUT1.

jklaise:
Examples without and with explicit `stepsJoin` (output) step:

**Model routing via tensors**
Original:
image
Explicit output join:
image

**Trigger joins**
Original:
image

Explicit output join:
image

@cliveseldon on this last example, what happens on the output if both `mul10` and `add10` are triggered?

jklaise:
Last commit also adds a diagram for the pipeline-pipeline example:
image

cliveseldon (approved):

PR #4843, opened 2023-05-14
build(install): Generate plain YAML manifests from Helm templates

### LINKED ISSUE #4842: Replace SCv2 YAML generation with 'helm template'
The current plain YAML files are stale -- they have not been kept up to date with the Helm charts with the security setup, env vars, etc.

As a simpler solution, we can generate the Helm charts as the main, expected method of installation, and use `helm template` to create a completely compatible set of plain YAML manifests.

This also works fine with Kustomize, as it can use a Helm template as an input source.

agrski:

Fixes #4842 

Tested with:
```
kind create cluster --name=issue-4842 --image=kindest/node:v1.24.7@sha256:577c630ce8e509131eab1aea12c022190978dd2f745aac5eb1fe65c0807eb315
ansible-playbook playbooks/setup-ecosystem.yaml

# To avoid re-pulling huge images from the Internet
kind load docker-image --name issue-4842 seldonio/mlserver:1.3.2
kind load docker-image --name issue-4842 nvcr.io/nvidia/tritonserver:22.03-py3

kubectl apply -f yaml/certs.yaml
kubectl apply -f yaml/crds.yaml --server-side
kubectl apply -f yaml/components.yaml 
kubectl apply -f yaml/servers.yaml 
```

### DIFF
--- a/k8s/Makefile
+++ b/k8s/Makefile
@@ -1,19 +1,12 @@
 CUSTOM_IMAGE_TAG ?= latest
 NEW_VERSION ?= 0.0.0
 
-.PHONY: create
-create: create-helm-charts create-yaml
-
-.PHONY: create-yaml
-create-yaml:
-	helm template seldon-core-v2-certs ./helm-charts/seldon-core-v2-certs > yaml/certs.yaml
-	helm template seldon-core-v2-crds ./helm-charts/seldon-core-v2-crds > yaml/crds.yaml
-	helm template seldon-core-v2-components ./helm-charts/seldon-core-v2-setup > yaml/components.yaml
-	helm template seldon-core-v2-servers ./helm-charts/seldon-core-v2-servers > yaml/servers.yaml
-
 HELM_CRD_BASE := helm-charts/seldon-core-v2-crds/templates
 HELM_COMPONENTS_BASE := helm-charts/seldon-core-v2-setup/templates
 HELM_SERVERS_BASE := helm-charts/seldon-core-v2-servers/templates
+
+.PHONY: create
+create: create-helm-charts create-yaml
 
 .PHONY: create-helm-charts
 create-helm-charts:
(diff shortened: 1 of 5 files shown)


k8s/yaml/servers.yaml
```diff
@@ -1,15 +1,20 @@
+---
+# Source: seldon-core-v2-servers/templates/seldon-v2-servers.yaml
 apiVersion: mlops.seldon.io/v1alpha1
 kind: Server
 metadata:
   name: mlserver
+  namespace: 'default'
```
agrski, line 7: :wrench: We probably want this to be templated nicely... Will need to look into why it isn't being

k8s/Makefile
```diff
@@ -57,13 +23,16 @@ create-helm-charts:
...
 set-chart-version:
-	sed -i "s/version: .*/version: ${NEW_VERSION}/" helm-charts/seldon-core-v2-certs/Chart.yaml
-	sed -i "s/appVersion: .*/appVersion: ${NEW_VERSION}/" helm-charts/seldon-core-v2-certs/Chart.yaml
-	sed -i "s/version: .*/version: ${NEW_VERSION}/" helm-charts/seldon-core-v2-crds/Chart.yaml
-	sed -i "s/appVersion: .*/appVersion: ${NEW_VERSION}/" helm-charts/seldon-core-v2-crds/Chart.yaml
-	sed -i "s/version: .*/version: ${NEW_VERSION}/" helm-charts/seldon-core-v2-setup/Chart.yaml
-	sed -i "s/appVersion: .*/appVersion: ${NEW_VERSION}/" helm-charts/seldon-core-v2-setup/Chart.yaml
-	sed -i "s/version: .*/version: ${NEW_VERSION}/" helm-charts/seldon-core-v2-servers/Chart.yaml
-	sed -i "s/appVersion: .*/appVersion: ${NEW_VERSION}/" helm-charts/seldon-core-v2-servers/Chart.yaml
+	sed -i 's/\(version\|appVersion\): .*/\1: ${NEW_VERSION}/' helm-charts/seldon-core-v2-certs/Chart.yaml
```
agrski, line 35: :information_source: For some reason, Make doesn't like use the extended-regex form for sed, at least on my machine.  When using that (via `-E` or `-r`), it complains about an invalid reference, but when running directly it's fine for me!  Perhaps a `sh` vs. `bash` difference?

agrski: Fixed in https://github.com/SeldonIO/seldon-core/pull/4843/commits/01b26c13d30c2eadff04f29d0eaf97ab6d1481e3 - apparently needed to provide `-i -r` instead of `-ir`

k8s/yaml/components.yaml
```diff
@@ -633,13 +713,39 @@ spec:
       containers:
       - env:
         - name: SELDON_KAFKA_BOOTSTRAP_SERVERS
-          value: seldon-kafka-bootstrap.seldon-mesh:9092
+          value: 'seldon-kafka-bootstrap.seldon-mesh:9092'
```
agrski, line 716: :information_source: This is a bit inconsistent.  Lots of things are put in the `default` namespace, but Kafka is expected in `seldon-mesh`

agrski: With the installation discussed in the PR body, this still works fine, although it may not when TLS certs are in use due to the different namespaces?  We could improve that in a follow-up PR

cliveseldon (approved):

PR #4844, opened 2023-05-15
Update copyright and licenses and fix missing local java sasl mechanism

cliveseldon:


scheduler/data-flow/src/main/resources/local.properties
```diff
@@ -18,4 +18,5 @@ kafka.tls.broker.secret=
 kafka.tls.endpoint.identification.algorithm=
 kafka.sasl.username=seldon
 kafka.sasl.secret=
-kafka.sasl.password.path=
\ No newline at end of file
+kafka.sasl.password.path=
+kafka.sasl.mechanism=PLAIN
```
RafalSkolasinski, line 22: :bulb: idea for later - maybe we could allow in data-flow code for this to be unset.

Also worth to note that in other places we default to SCRAM-SHA-512 but probably does not make much difference.

RafalSkolasinski (approved):

PR #4868, opened 2023-05-25
removing alpha messaging from index pages

edshee:

Fixes #


edshee:
Not sure what's causing the docs build to fail. I've only changed two lines. Was it failing before this?

cliveseldon (approved):

PR #4891, opened 2023-06-06
ci: Add CI workflow ensuring PR titles conform to Conventional Commit format

agrski:
This PR adds a check for PR titles to ensure that they are compatible with the [Conventional Commit format](https://www.conventionalcommits.org/en/v1.0.0/).  This imposes minimal burden to contributors: they need to specify the **type** of change, whether or not it is a breaking change, and optionally its scope.

This has a couple of benefits, both for human users and automation.  For maintainers, it is easier to quickly grok the scope and impact of changes due to the additional context.  Using tools like grep, it is also easy to filter for particular types of changes.  We could consider using the CC format in release notes in the future, such that changes are automatically grouped into sections, e.g. new features, bug fixes, and breaking changes.

N/A

The PR check can be seen working on this PR.  Initially it will fail as I have deliberately not met the CC format for the title.  After changing this, the check should pass.

### DIFF
--- a/.github/workflows/pr-title.yaml
+++ b/.github/workflows/pr-title.yaml
@@ -1,0 +1,23 @@
+name: lint-pr-title
+
+on:
+  pull_request:
+    branches:
+      - v2
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+
+jobs:
+  conventional-commit:
+    if: github.repository == 'SeldonIO/seldon-core' # Do not run this on forks.
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - name: Check PR title matches Conventional Commit format
+        uses: matthiashermsen/lint-pull-request-title@v1.0.0
+        with:
+          allowed-pull-request-types: build,ci,docs,feat,fix,refactor,revert,test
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-title.yaml
```diff
@@ -0,0 +1,23 @@
...
+  conventional-commit:
+    if: github.repository == 'SeldonIO/seldon-core' # Do not run this on forks.
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - name: Check PR title matches Conventional Commit format
+        uses: matthiashermsen/lint-pull-request-title@v1.0.0
+        with:
+          allowed-pull-request-types: build,ci,docs,feat,fix,refactor,revert,test
```
sakoush, line 23: if we have a developer/contribution guide section, perhaps add this to it?

agrski: We do indeed have a `CONTRIBUTING.md`.  I'll enact this shortly - good suggestion!

agrski: Fixed in https://github.com/SeldonIO/seldon-core/pull/4891/commits/adc91d0502b89383a8d5f42ab2f7122153da0632

sakoush (approved):
LGTM

PR #4901, opened 2023-06-08
Add note in v2 docs about vCPU compatibility

jganseman:
```release-note
Added note to the local install docs about vCPU compatibility
```

**What this PR does / why we need it**: Incompatible vCPUs lead to containers in Docker crashing with "glibc x86-64-v2" errors. Users deploying on their own cluster should be aware they need to create VMs with compatible vCPUs.

Fixes #


edshee (approved):
Thanks for this

PR #4908, opened 2023-06-12
fix: Remove default namespace from yaml files

sakoush:
Currently when we create the yaml files under `k8s/yaml`, we output a `namespace: 'default'` in the components metadata. This seems to cause issues when deploying to a specific namespace. 
This PR then removes these entries. Fixed also the makefile that generates these yaml files.
There seems also to be outdated settings that are now regenerated as well. 

Fixes #

### DIFF
--- a/k8s/yaml/components.yaml
+++ b/k8s/yaml/components.yaml
@@ -4,35 +4,30 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: hodometer
-  namespace: 'default'
 ---
 # Source: seldon-core-v2-setup/templates/seldon-v2-components.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: seldon-controller-manager
-  namespace: 'default'
 ---
 # Source: seldon-core-v2-setup/templates/seldon-v2-components.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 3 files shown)


k8s/yaml/components.yaml
```diff
@@ -718,30 +688,36 @@ spec:
           value: '1'
         - name: SELDON_KAFKA_PARTITIONS_DEFAULT
           value: '1'
-        - name: SELDON_CORES_COUNT
-          value: '4'
+        - name: SELDON_KAFKA_MAX_MESSAGE_SIZE_BYTES
```
sakoush, line 691: not sure if these settings are meant to be like that?

k8s/yaml/components.yaml
```diff
@@ -439,90 +424,78 @@ apiVersion: rbac.authorization.k8s.io/v1
...
   name: seldon-scheduler-rolebinding
-  namespace: 'default'
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
   name: seldon-scheduler-role
 subjects:
 - kind: ServiceAccount
   name: seldon-scheduler
-  namespace: 'default'
```
cliveseldon, line 525: rbac Role needs namespace so this is maybe problematic?

cliveseldon:
closing for now

Issue #4870, opened 2023-05-26
Use one-sided TLS for Kafka communication


SDJustus:
Following on @RafalSkolasinski work at #4828, it would be nice for client side authentication to be optional when using SSL (at the moment only SASL and mTLS are possible). 

The use case is not having to maintain client side certificates, if the security aspect is covered by other infrastructure, such as having the Cluster and the Kafka in the same VPC in an AWS Environment.

#4658 already did some work on this, so maybe this could be a supported mechanism for connecting a Kafka broker to seldon-core v2.

RafalSkolasinski:
@SDJustus thanks for opening the issue! W

Would be great to have some example Kafka configuration that illustrated desired consumer/producer config.

SDJustus:
Of course.
This would be the broker settings (AWS MSK):
![Screenshot from 2023-05-26 13-05-31](https://github.com/SeldonIO/seldon-core/assets/44165116/1d9740d7-5a55-4e9f-8356-afa1db994b4b)

Kafka Client Security Settings (i.e. from a Spring Boot Application):
```
security.protocol: SSL
ssl.endpoint.identification.algorithm: HTTPS
```

Let me know, if you need further information. Thanks for having a look on this.

RafalSkolasinski:
@SDJustus is the `ssl.endpoint.identification.algorithm: HTTPS` option just cherry on top or required for it to work properly?

SDJustus:
I just checked the confluence [page](https://docs.confluent.io/platform/current/kafka/authentication_ssl.html) and https seems to be the default.. Therefore I don't think it needs to be specified.

RafalSkolasinski:
closed via https://github.com/RafalSkolasinski/seldon-core/pull/131

RafalSkolasinski:
closed via https://github.com/SeldonIO/seldon-core/pull/4916

SDJustus:
Will take a look. Thank you very much for implementing!

PR #4916, opened 2023-06-13
feat: One side TLS for Kafka

RafalSkolasinski:
Building on top of https://github.com/SeldonIO/seldon-core/pull/4658
Closes https://github.com/SeldonIO/seldon-core/issues/4870

Tested with:
- Strimzi Plain
- Strimzi one sided TLS
- Strimzi mTLS
- Confluent Cloud

To make use of one -sided TLS simply leave `security.kafka.ssl.client.secret` value unset, e.g.
```yaml
kafka:
  bootstrap: seldon-kafka-tls-bootstrap.kafka:9093

security:
  kafka:
    protocol: SSL
    ssl:
      client:
        secret:
        brokerValidationSecret: kafka-broker-auth
        endpointIdentificationAlgorithm: https
```

Note: this also adds propagation of 
```yaml
security:
  kafka:
    ssl:
      client:
        endpointIdentificationAlgorithm: https
```
Helm value to Golang client if SSL or SASL_SSL is being used. 
Empty value will result in `none` value being set as otherwise golang client raises

> Configuration property "ssl.endpoint.identification.algorithm" cannot be set to empty value.


RafalSkolasinski:
For reference, strimzi listeners configuration for testing:
```
    listeners:
    - name: plain
      port: 9092
      tls: false
      type: loadbalancer
    - name: tls
      port: 9093
      tls: true
      type: loadbalancer
    - authentication:
        type: tls
      name: mtls
      port: 9094
      tls: true
      type: loadbalancer
```

components/tls/pkg/tls/util.go
```diff
@@ -63,3 +67,13 @@ func GetSASLMechanismFromEnv(prefix string) string {
 	}
 	return val
 }
+
+func GetEndpointIdentificationMechanismFromEnv(prefix string) string {
+	val, ok := os.LookupEnv(fmt.Sprintf("%s%s", prefix, EnvEndpointIdentificationMechanismSuffix))
+	if !ok {
+		val = EndpointIdentificationMechanismHTTPS
```
adriangonz, line 74: Should we default to `https` if the env var is not present? Or would it be better to keep it this way for backwards compatibility?

components/tls/pkg/tls/util.go
```diff
@@ -63,3 +67,13 @@ func GetSASLMechanismFromEnv(prefix string) string {
 	}
 	return val
 }
+
+func GetEndpointIdentificationMechanismFromEnv(prefix string) string {
```
adriangonz, line 71: Is there any other instance where the `prefix` is different to `KAFKA_CLIENT`?

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
```diff
@@ -75,21 +75,25 @@ private fun getSecurityProperties(params: KafkaStreamsParams): Properties {
...
-            params.security.certConfig.endpointIdentificationAlgorithm
+
+        if (certConfig.clientSecret.isNotEmpty()) {
+            val keyStoreConfig = Provider.keyStoreFromCertificates(certConfig)
+            this[SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG] = keyStoreConfig.keyStoreLocation
+            this[SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG] = keyStoreConfig.keyStorePassword
+            this[SslConfigs.SSL_KEY_PASSWORD_CONFIG] = keyStoreConfig.keyStorePassword
+        }
+
+        this[SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG] = certConfig.endpointIdentificationAlgorithm
```
adriangonz, line 96: Does the `endpointIdentificationAlgorithm` key in Kotlin follow similar rules to the config in Go? As in,

- If not present, default to `ssl`
- If present but empty, default to `plaintext`
- Otherwise, use the value

scheduler/pkg/kafka/config/tls.go
```diff
@@ -79,27 +82,46 @@ func setupSASLSSLAuthentication(config kafka.ConfigMap) error {
...
+	cs, err := tls.NewCertificateStore(
+		tls.Prefix(EnvKafkaClientPrefix),
+		tls.ValidationPrefix(EnvKafkaBrokerPrefix),
+	)
 	if err != nil {
-		return err
+		cs, err = tls.NewCertificateStore(
+			tls.ValidationOnly(true),
+			tls.ValidationPrefix(EnvKafkaBrokerPrefix),
+		)
```
adriangonz, line 101: Just to clarify, does this bit try to first set up mTLS and then falls back to one-sided TLS after erroring?

components/tls/pkg/tls/util.go
RafalSkolasinski: It may be `KAFKA_BROKER` I believe - this is following logic of other helpers.

components/tls/pkg/tls/util.go
RafalSkolasinski: That's what is happening - if not present it will be `https`. 
If it is empty string it will be `None` (which as I understand for Golang is equivalent of empty string). 

In any case I wanted to ensure backwards compatibility

scheduler/pkg/kafka/config/tls.go
RafalSkolasinski: yes - exactly

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
RafalSkolasinski: From the [documentation](https://docs.confluent.io/platform/current/kafka/authentication_ssl.html#id1)  

> The endpoint identification algorithm used by clients to validate server host name. The default value is https. Clients including client connections created by the broker for interbroker communication verify that the broker host name matches the host name in the broker’s certificate. Disable server host name verification by setting ssl.endpoint.identification.algorithm to an empty string.

The thing was that Golang client errors when you set it to an empty string, hence I believe where the value `none` comes from. 

The behaviour should therefore be consistent - worth to note that we always set the environmental variable for in our helm deployment.

scheduler/pkg/kafka/config/tls.go
adriangonz: Got it - it may be worth adding a comment to clarify that's what's going on?

scheduler/pkg/kafka/config/tls.go
```diff
@@ -79,27 +82,46 @@ func setupSASLSSLAuthentication(config kafka.ConfigMap) error {
...
 
 func setupTLSAuthentication(config kafka.ConfigMap) error {
-	cs, err := tls.NewCertificateStore(tls.Prefix(EnvKafkaClientPrefix),
-		tls.ValidationPrefix(EnvKafkaBrokerPrefix))
+	var cs *tls.CertificateStore
+	cs, err := tls.NewCertificateStore(
+		tls.Prefix(EnvKafkaClientPrefix),
+		tls.ValidationPrefix(EnvKafkaBrokerPrefix),
+	)
 	if err != nil {
```
RafalSkolasinski, line 97:
```suggestion
        // in case of error (missing client certificates) fallback to one-sided TLS
	if err != nil {
```

SDJustus:
works as expected. Thank you very much! :100:

Issue #4931, opened 2023-06-21
Scheduler panic on server restart with model loaded


cliveseldon:
Reproduced by killing mlserver after model sucessfully loaded on k8s.

```
│ panic: runtime error: invalid memory address or nil pointer dereference                                                                                                 │
│ [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a0244d]                                                                                                 │
│                                                                                                                                                                         │
│ goroutine 190 [running]:                                                                                                                                                │
│ github.com/seldonio/seldon-core/scheduler/v2/pkg/envoy/processor.(*IncrementalProcessor).updateEnvoyForModelVersion(0xc0005c1ad0, {0xc0020360a0, 0x4}, 0xc0018d5d00, 0x │
│     /build/scheduler/pkg/envoy/processor/incremental.go:264 +0xcd                                                                                                       │
│ github.com/seldonio/seldon-core/scheduler/v2/pkg/envoy/processor.(*IncrementalProcessor).addModelTraffic(0xc0005c1ad0, {0xc0020360a0, 0x4}, 0xc001849fb0, 0x1a44f18?, 0 │
│     /build/scheduler/pkg/envoy/processor/incremental.go:337 +0x652                                                                                                      │
│ github.com/seldonio/seldon-core/scheduler/v2/pkg/envoy/processor.(*IncrementalProcessor).addModel(0xc0005c1ad0, 0xc001849fb0)                                           │
│     /build/scheduler/pkg/envoy/processor/incremental.go:393 +0x305                                                                                                      │
│ github.com/seldonio/seldon-core/scheduler/v2/pkg/envoy/processor.(*IncrementalProcessor).modelUpdate(0xc0005c1ad0, {0xc0020360a0, 0x4})                                 │
│     /build/scheduler/pkg/envoy/processor/incremental.go:592 +0xa8b                                                                                                      │
│ github.com/seldonio/seldon-core/scheduler/v2/pkg/envoy/processor.(*IncrementalProcessor).handleModelEvents.func1()                                                      │
│     /build/scheduler/pkg/envoy/processor/incremental.go:183 +0x54                                                                                                       │
│ created by github.com/seldonio/seldon-core/scheduler/v2/pkg/envoy/processor.(*IncrementalProcessor).handleModelEvents                                                   │
│     /build/scheduler/pkg/envoy/processor/incremental.go:182 +0x1b8  
```

cliveseldon:
Fixed

PR #4934, opened 2023-06-21
fix: server can be return as nil when not found causing panics

cliveseldon:
FIXES: #4931 where a panic was causes when mlserver was deleted while it had a model. The envoy logic was getting the server for this model but there was none.

We were return no error when a server was not found but just a nil server. Having checked all the uses I think returning an error is best and no functionality will change. 

In future we could return a typed error for no server found if we think clients need to check for this type of error but this does not seem to be needed right now.

Have tested the use case in #4931 and the model is reloaded ok after the mlserver pod comes back.

The PR also cleans up the Envoy update to not do unnecessary calls when an error state has been hit. The control flow still needs to reach the end of the `modelUpdate` call to do batch updates.


sakoush (approved):
LGTM

PR #4935, opened 2023-06-22
adding end to end tensorflow (mlserver) example

edshee:
Need to check docs render properly on RTD quickly first.


edshee:
Docs test seems to fail due to generating the sitemap. I ran one locally but the build succeeded - is it possibly something to do with the build environment?

PR #4937, opened 2023-06-22
fix: Allow config CRD changes to cause a reconcilliation loop for dependent (but not owned) resources

cliveseldon:
At present changes to SeldonConfig do not cause a reconcile of associated SeldonRuntimes and ServerConfig changes do not cause a reconcile for Servers. This would hamper rollout of changes.

There is not a direct ownership from SeldonConfigs to SeldonRuntimes or ServerConfigs to Servers as SeldonRuntimes and Servers are created later and have a `spec` link to an associated config.

The solution is to add to the controllers watches with custom reconcile e.g.

```
Watches(
			&source.Kind{Type: &mlopsv1alpha1.SeldonConfig{}},
			handler.EnqueueRequestsFromMapFunc(r.mapSeldonRuntimesFromSeldonConfig),
		).
```

where we add 
```
// Find SeldonRuntimes that reference the changes SeldonConfig
func (r *SeldonRuntimeReconciler) mapSeldonRuntimesFromSeldonConfig(obj client.Object) []reconcile.Request {
	logger := log.FromContext(context.Background()).WithName("mapSeldonRuntimesFromSeldonConfig")
	var seldonRuntimes mlopsv1alpha1.SeldonRuntimeList
	if err := r.Client.List(context.Background(), &seldonRuntimes); err != nil {
		logger.Error(err, "error listing seldonRuntimes")
		return nil
	}

	seldonConfig := obj.(*mlopsv1alpha1.SeldonConfig)
	var req []reconcile.Request
	for _, seldonRuntime := range seldonRuntimes.Items {
		if seldonRuntime.Spec.SeldonConfig == seldonConfig.Name {
			req = append(req, reconcile.Request{
				NamespacedName: types.NamespacedName{
					Namespace: seldonRuntime.Namespace,
					Name:      seldonRuntime.Name,
				},
			})
		}
	}
	return req
}
```

### DIFF
--- a/operator/controllers/mlops/seldonruntime_controller.go
+++ b/operator/controllers/mlops/seldonruntime_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 
-
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
--- a/operator/controllers/mlops/seldonruntime_controller.go
+++ b/operator/controllers/mlops/seldonruntime_controller.go
@@ -19,8 +19,7 @@ package mlops
 import (
 	"context"
 	"fmt"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/source"
+
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
@@ -35,8 +34,10 @@ import (
 	"knative.dev/pkg/apis"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	mlopsv1alpha1 "github.com/seldonio/seldon-core/operator/v2/apis/mlops/v1alpha1"
 	"github.com/seldonio/seldon-core/operator/v2/controllers/reconcilers/common"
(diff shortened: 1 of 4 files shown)


operator/controllers/mlops/seldonruntime_controller.go
```diff
@@ -217,6 +220,30 @@ func (r *SeldonRuntimeReconciler) updateStatus(seldonRuntime *mlopsv1alpha1.Seld
 	return nil
 }
 
+// Find SeldonRuntimes that reference the changes SeldonConfig
+func (r *SeldonRuntimeReconciler) mapSeldonRuntimesFromSeldonConfig(obj client.Object) []reconcile.Request {
+	logger := log.FromContext(context.Background()).WithName("mapSeldonRuntimesFromSeldonConfig")
+	var seldonRuntimes mlopsv1alpha1.SeldonRuntimeList
+	if err := r.Client.List(context.Background(), &seldonRuntimes); err != nil {
+		logger.Error(err, "error listing seldonRuntimes")
+		return nil
```
sakoush, line 229: do we want to pass an error to the caller?

operator/controllers/mlops/server_controller.go
```diff
@@ -226,11 +229,39 @@ func (r *ServerReconciler) updateStatus(server *mlopsv1alpha1.Server) error {
 	return nil
 }
 
+// Find Servers that need reconcilliation from a change to a given ServerConfig
+func (r *ServerReconciler) mapServerFromServerConfig(obj client.Object) []reconcile.Request {
+	logger := log.FromContext(context.Background()).WithName("mapServerFromServerConfig")
+	var servers mlopsv1alpha1.ServerList
+	if err := r.Client.List(context.Background(), &servers); err != nil {
+		logger.Error(err, "error listing servers")
+		return nil
```
sakoush, line 238: ditto

sakoush (approved):
LGTM, I left a minor comment on error handling

It would also be good to add this somewhere in the docs to explain the behaviour for the user.

Also worth adding a follow up ticket to make this configurable.

operator/controllers/mlops/seldonruntime_controller.go
cliveseldon: connected to https://github.com/kubernetes-sigs/controller-runtime/issues/2235 so not sure how to handle will need to look further

operator/controllers/mlops/server_controller.go
cliveseldon: https://github.com/kubernetes-sigs/controller-runtime/issues/2235

PR #4945, opened 2023-06-27
fix: Fix test logic, get rid of k6 warning

apaleyes:
Two things:
1. Gets rid of "The headers property is deprecated, replace it with the metadata property, please." warning from k6
2. Fixes logic in "infer_constant_vu.js" scenario so that we don't skip iterations when using only one protocol for test

Fixes #

tested by running locally


sakoush (approved):
Lovely! thanks for sorting it out.

PR #4963, opened 2023-06-29
fix: remove last applied annotation in operator when adding annotations

cliveseldon:
We need to remove the annotation `kubectl.kubernetes.io/last-applied-configuration` if it appears in custom resources so they don't get pushed down to the owned resources causing all owned resources to be recreated on any change. This is most apprarent for stateful sets such as Servers where a change for `replicas` would cause always all underplying pods to be restarted even when you might need to just add or remove some replica pods.


seldondev:
* failed to create agent
* failed to load trigger config for repository SeldonIO/seldon-core for ref 
* failed to switch to ref 
* failed to fetch repository SeldonIO/seldon-core
* error fetching
* exit status 128 fatal
* unable to access 'https://github.com/SeldonIO/seldon-core/'
* Failed to connect to github.com port 443 after 130915 ms
* Operation timed out

sakoush (approved):

PR #4990, opened 2023-07-06
fix: Disable default logging for Kafka in Ansible Playbooks

vtaskow:
When running Seldon via Ansible playbooks, Kafka logs everything, which makes it difficult to inspect the logs of `seldon-pipelinegateway` when using pipelines.

### DIFF
--- a/ansible/playbooks/templates/values-seldon-core-v2-components.yaml.j2
+++ b/ansible/playbooks/templates/values-seldon-core-v2-components.yaml.j2
@@ -7,7 +7,6 @@ envoy:
     type: "{{ seldon_core_v2_type_svc_type }}"
 
 kafka:
-  debug: all
   bootstrap: "seldon-kafka-bootstrap.{{ seldon_kafka_namespace }}.svc.cluster.local:9093"
   topics:
     numPartitions: 4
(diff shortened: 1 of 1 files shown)


RafalSkolasinski (approved):

agrski (commented):
Rather than removing the config from the Helm chart, I'd instead suggest adding a Helm variable to control it.  Note that this is not config for Kafka itself, but rather for _our_ components _about_ Kafka.

Being able to modify log verbosity is useful for both developers and end users, so I'd prefer not to see utility removed!

cliveseldon:
> Rather than removing the config from the Helm chart, I'd instead suggest adding a Helm variable to control it. Note that this is not config for Kafka itself, but rather for _our_ components _about_ Kafka.
> 
> Being able to modify log verbosity is useful for both developers and end users, so I'd prefer not to see utility removed!

There is a helm value for it. This is the ansible use of it which should be configurable via allowing helm values or for now removed so its not hard wired.

agrski:
> There is a helm value for it. This is the ansible use of it which should be configurable via allowing helm values or for now removed so its not hard wired.

Ah yes, I forgot the context.  I'd still be in favour of this being configurable easily.  I don't mind if that's even just leaving it as `info` level so it's still obvious that setting is there, without having to dig around to find it.

cliveseldon (approved):

cliveseldon (approved):

PR #5014, opened 2023-07-11
feat(helm): Make SeldonConfig configurable in SeldonRuntime Helm chart

### LINKED ISSUE #5013: Be able to configure `seldonConfig` for a `SeldonRuntime`
It would be nice to be able to configure the `SeldonConfig` being used for a given `SeldonRuntime`. Otherwise the `default` will always be used unless the chart is overridden.

franco-bocci:
Makes `SeldonConfig` configurable for `SeldonRuntime`.

Fixes #5013 

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-runtime/templates/seldon-runtime.yaml
+++ b/k8s/helm-charts/seldon-core-v2-runtime/templates/seldon-runtime.yaml
@@ -4,7 +4,7 @@ metadata:
   name: seldon
   namespace: '{{ .Release.Namespace }}'  
 spec:
-  seldonConfig: default
+  seldonConfig: {{ .Values.seldonConfig }}
   disableAutoUpdate: {{ .Values.disableAutoUpdate }}
   overrides:
   - name: hodometer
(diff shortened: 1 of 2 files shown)


seldondev:
Hi @franco-bocci. Thanks for your PR.

<details>

</details>

agrski (changes requested):
Hi @franco-bocci,

Thanks for contributing a PR to fix this important oversight!

The way the Helm chart works is that it's generated using `kustomize` patches.  That means you need to update those patches, otherwise the next time someone runs this generation process, your changes will be lost!

agrski (approved):
Hi @franco-bocci,

Looks like I was wrong and this particular Helm chart is in fact hand-crafted, rather than following the kustomization approach of all the others.

Given that, these changes look correct!

franco-bocci:
> Hi @franco-bocci,
> 
> Looks like I was wrong and this particular Helm chart is in fact hand-crafted, rather than following the kustomization approach of all the others.
> 
> Given that, these changes look correct!

Thanks for the review! 🚀

Issue #5021, opened 2023-07-13
hodometer needs to be enabled for seldonConfig to be deployed inside helm chart


SDJustus:
Is there a reason, why hodometer needs to be enabled inside the setup helm chart for the default SeldonConfig to be deployed?
Also, inside the setup helm chart, there is hodometer.enable and in the runtime helmchart, there is hodometer.disable. I think it would be nice to have either enabled or disabled everywhere.

cliveseldon:
This is a typo in the docs [here](https://docs.seldon.io/projects/seldon-core/en/v2/contents/metrics/usage.html).
We will update the docs as now its only on the runtime that you need to disable components if you wish.

cliveseldon:
Fixed

PR #5025, opened 2023-07-13
fix: hodometer enable flag and docs

cliveseldon:
* Fix docs for hodometer disable
 * Remove enable flag from setup chart as hodometer can be disabled from runtime chart now

Fixes: #5021


k8s/Makefile
```diff
@@ -16,11 +16,7 @@ create-helm-charts:
 	kustomize build kustomize/helm-components-ns/ > ${HELM_COMPONENTS_BASE}/seldon-v2-components.yaml
 	kustomize build kustomize/helm-servers/ > ${HELM_SERVERS_BASE}/seldon-v2-servers.yaml
 	sed -i 's/\(.*\)\(imagePullSecrets:\).*/{{- with .Values.imagePullSecrets }}\n\1\2\n\1{{- toYaml . | nindent 8 }}\n{{- end }}/g' ${HELM_COMPONENTS_BASE}/seldon-v2-components.yaml
-	sed -n --file=add-helm-toggle-hodometer.sed ${HELM_COMPONENTS_BASE}/seldon-v2-components.yaml \
```
agrski, line 19: :wrench: As we don't need the magical `sed` script any more, that can be deleted (as much as it saddens me to see it go!)

agrski (commented):
Looks good apart from removing the now-defunct sed script

k8s/Makefile
cliveseldon: fixed in https://github.com/SeldonIO/seldon-core/pull/5025/commits/a7511fb831865eeedef8c088f4cc99cc034324da

agrski (approved):

PR #5030, opened 2023-07-17
fix: Start triton server via bash -c tritonserver instead of just tritonserver

### LINKED ISSUE #5029: Start triton server via bash -c tritonserver instead of just tritonserver
## Describe the bug
At the moment, Triton seems to have a sort of [bug](https://github.com/triton-inference-server/server/issues/3877), where it has to be started using `bash -c tritonserver` rather calling the binary directly (as it is done in the default ServerConfig, provided by the Helm Charts). This only occurs when using triton with CUDA and using i.e. Version 22.12. This is not really a seldon-core bug but it would be nice to have this different call in the default ServerConfig to not have to write an own one.
Sorry to open the bug here, since this isn't really seldon core related, but the triton dev team seems to see the start with `bash -c tritonserver` as a fix, as the last comment on the [issue](https://github.com/triton-inference-server/server/issues/3877) states.

## To reproduce
1. Deploy a Triton Server using GPU
2. Deploy a Model (i.e. torchscript or onnx model, dali works) 
3. get the error message from triton: `UNAVAILABLE: Internal: unable to create stream: the provided PTX was compiled with an unsupported toolchain`

## Expected behaviour
Models to be deployable on GPU

## Environment

* Cloud Provider: AWS EKS
* Kubernetes Cluster Version 1.24
* Deployed Seldon System Images: Helm Charts of Release 2.6.0

## Model Details 
i.e. scripted pytorch model traced within the nvcr.io/nvidia/pytorch:22.12-py3 container from NGC

SDJustus:
adjusted the components.yaml and the corresponding helm chart file, to be able to launch newer triton versions with gpu support, according to this [issue](https://github.com/triton-inference-server/server/issues/3877) in the triton repository.

Fixes #5029 

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
+++ b/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
@@ -1869,14 +1869,12 @@ spec:
       - mountPath: /mnt/tracing
         name: tracing-config-volume
     - args:
-      - --model-repository=$(SERVER_MODELS_DIR)
-      - --http-port=$(SERVER_HTTP_PORT)
-      - --grpc-port=$(SERVER_GRPC_PORT)
-      - --log-verbose=1
-      - --model-control-mode=explicit
-      - --backend-config=python,shm-default-byte-size=16777216
+      - -c
+      - tritonserver --model-repository=$(SERVER_MODELS_DIR) --http-port=$(SERVER_HTTP_PORT)
+        --grpc-port=$(SERVER_GRPC_PORT) --log-verbose=1 --model-control-mode=explicit
+        --backend-config=python,shm-default-byte-size=16777216
       command:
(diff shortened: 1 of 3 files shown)


cliveseldon:
@SDJustus As at present we create the yaml/helm from kustomize you will actually need to:
 1. change https://github.com/SeldonIO/seldon-core/blob/ea5f5d424ce8e95186530c8b116fcddd3ff34eac/operator/config/serverconfigs/triton.yaml#L118-L126
 2. Rerun https://github.com/SeldonIO/seldon-core/blob/ea5f5d424ce8e95186530c8b116fcddd3ff34eac/k8s/Makefile#L10

In future we plan for the helm to be the source of truth,

SDJustus:
Any updates on this? Maybe @cliveseldon

SDJustus:
Hi,
Any updates on this?

PR #5051, opened 2023-07-21
fix: Allow digits in topic prefix

### LINKED ISSUE #5050: Kafka topic prefix misses allowing digits
https://github.com/SeldonIO/seldon-core/blob/v2/scheduler/pkg/kafka/topics.go#L46-L59

cliveseldon:
Adds missing digits to regexp checking topic prefix.

Fixes #5050

### DIFF
--- a/scheduler/pkg/kafka/topics.go
+++ b/scheduler/pkg/kafka/topics.go
@@ -44,7 +44,7 @@ type TopicNamer struct {
 }
 
 var (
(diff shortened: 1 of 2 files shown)


agrski (approved):
This is consistent with the [Kafka definition](https://github.com/apache/kafka/blob/0.10.2/core/src/main/scala/kafka/common/Topic.scala#L29), at least in a previous version

PR #5073, opened 2023-08-08
fix(operator): consistent pod service monitors reconcilor app labels

### LINKED ISSUE #5071: Pipeline Gateway Podmonitor label selector does not match Pipeline Gateway labels
It looks like the configuration of the prometheus podmonitors were not updated correctly when the [SeldonRuntime CRD was introduced](https://github.com/SeldonIO/seldon-core/pull/4847/files#diff-608c783a8c14298e3151fc91b57a38a686e07bf59a44b19ad0caa725490ede5c).

Previously, we used labels to configure which resources our podmonitors would scrape metrics from, e.g. for the [pipelinegateway](https://github.com/SeldonIO/seldon-core/blob/79afceae01540a08cca158535540189b381f79da/prometheus/monitors/pipelinegateway-podmonitor.yaml#L8):
```
spec:
  selector:
    matchLabels:
      app: pipelinegateway
```

I can see from the the reconcilers code that we are adding the labels to the deployment spec like so
```
templateLabels := utils.MergeMaps(map[string]string{constants.KubernetesNameLabelKey: name}, labels)
```
where the `constants.KubernetesNameLabelKey` is defined as
```
const (
	AppKey                    = "app"
	KubernetesNameLabelKey    = "app.kubernetes.io/name"
```

From what I can tell, name is typically defined as `seldon-pipelinegateway` from the [operator api definition](https://github.com/SeldonIO/seldon-core/blob/79afceae01540a08cca158535540189b381f79da/operator/apis/mlops/v1alpha1/seldonconfig_types.go#L38) which results in the following label `app.kubernetes.io/name=seldon-pipelinegateway` which does not match our existing podmonitors.

This results in prometheus being unable to scrape things like the `pipelinegateway` and its [metrics](https://docs.seldon.io/projects/seldon-core/en/v2/contents/metrics/operational.html#list-of-scv2-metrics) being missing.

As far as I can tell, the server metrics happen to be fine because [they use the `constants.AppKey`](https://github.com/SeldonIO/seldon-core/blob/79afceae01540a08cca158535540189b381f79da/operator/controllers/reconcilers/server/statefulset_reconciler.go#L73) so have not changed. I'm not sure if the [envoy ServiceMonitor](https://github.com/SeldonIO/seldon-core/blob/79afceae01540a08cca158535540189b381f79da/prometheus/monitors/envoy-servicemonitor.yaml#L8) was similarly affected

## Model Details 
* Images of your model: [Output of: `kubectl get seldondeployment -n <yourmodelnamespace> <seldondepname> -o yaml | grep image:` where `<yourmodelnamespace>`]
* Logs of your model: [You can get the logs of your model by running `kubectl logs -n <yourmodelnamespace> <seldonpodname> <container>`]

michaelcheah:
The default `matchLabel` configs for Prometheus `podmonitor`s do not match what is configured by the `SeldonRuntime` operator for several resources.

This PR:
1. Fixes the labels so that they are consistent across all resources, i.e. all label use `app.kubernetes.io/name` key instead of `app` to reference seldon components
2. The component name `seldon-mesh` is outdated and is updated to `seldon-envoy`.
3. The server metrics port for the default `mlserver` is not defined (it is defined for the `triton` server). We expose it on port 8082 which is the [default](https://mlserver.readthedocs.io/en/latest/reference/settings.html#mlserver.settings.Settings.metrics_port) for `mlserver`

Fixes #5071

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
+++ b/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
@@ -1663,6 +1663,8 @@ spec:
       - containerPort: 9000
         name: server-http
         protocol: TCP
+      - containerPort: 8082
+        name: server-metrics
       readinessProbe:
         httpGet:
           path: /v2/health/live
--- a/prometheus/monitors/pipelinegateway-podmonitor.yaml
+++ b/prometheus/monitors/pipelinegateway-podmonitor.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app: pipelinegateway
+      app.kubernetes.io/name: pipelinegateway
   namespaceSelector:
     matchNames: []
     any: false
--- a/operator/controllers/reconcilers/server/statefulset_reconciler.go
+++ b/operator/controllers/reconcilers/server/statefulset_reconciler.go
@@ -70,9 +70,9 @@ func toStatefulSet(meta metav1.ObjectMeta,
 	scaling *mlopsv1alpha1.ScalingSpec,
 	labels map[string]string,
 	annotations map[string]string) *appsv1.StatefulSet {
-	labels[constants.AppKey] = constants.ServerLabelValue
-	metaLabels := utils.MergeMaps(map[string]string{constants.AppKey: constants.ServerLabelValue}, labels)
-	templateLabels := utils.MergeMaps(map[string]string{constants.ServerLabelNameKey: meta.Name, constants.AppKey: constants.ServerLabelValue}, labels)
+	labels[constants.KubernetesNameLabelKey] = constants.ServerLabelValue
+	metaLabels := utils.MergeMaps(map[string]string{constants.KubernetesNameLabelKey: constants.ServerLabelValue}, labels)
+	templateLabels := utils.MergeMaps(map[string]string{constants.ServerLabelNameKey: meta.Name, constants.KubernetesNameLabelKey: constants.ServerLabelValue}, labels)
 	ss := &appsv1.StatefulSet{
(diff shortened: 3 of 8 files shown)


cliveseldon (approved):

RafalSkolasinski (changes requested):

prometheus/monitors/pipelinegateway-podmonitor.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app: pipelinegateway
```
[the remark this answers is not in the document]
RafalSkolasinski: Old labels could be done as in my test [here](https://github.com/RafalSkolasinski/seldon-core/commit/0980b396af5621ab6f65bef2f539f37da05877d9) - I would prefer to have the old labels there as well and then rely on these for metrics.

RafalSkolasinski (approved):

k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
```diff
@@ -1663,6 +1663,8 @@ spec:
       - containerPort: 9000
         name: server-http
         protocol: TCP
+      - containerPort: 8082
```
RafalSkolasinski, line 1666: @michaelcheah did you check that mlserver indeed expose metrics under this port?

RafalSkolasinski: Ok, it does.

RafalSkolasinski: Ok, it does.

PR #5074, opened 2023-08-09
fix(scheduler): Manual trigger envoy update

sakoush:

In some cases under heavy load, periodic batched envoy model update does not get a fair chance to run due to lock contention. This fix introduces a manual trigger as well (which does not require to release the lock and therefore is guaranteed to proceed).

This PR also adds:
- increases the queue sizes of the event hub to 1000.
- do not reset servers for models that fail scheduling but have replicas still loaded on some.

Fixes #

### DIFF
--- a/scheduler/pkg/envoy/processor/incremental.go
+++ b/scheduler/pkg/envoy/processor/incremental.go
@@ -655,6 +655,7 @@ func (p *IncrementalProcessor) callVersionCleanupIfNeeded(modelName string) {
 }
 
 func (p *IncrementalProcessor) triggerModelSyncIfNeeded() bool {
+	// the first time we trigger the batch update we need to set the time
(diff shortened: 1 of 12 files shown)


scheduler/pkg/envoy/processor/incremental.go
```diff
@@ -640,10 +654,32 @@ func (p *IncrementalProcessor) callVersionCleanupIfNeeded(modelName string) {
 	}
 }
 
-func (p *IncrementalProcessor) modelSync() {
-	logger := p.logger.WithField("func", "modelSync")
+func (p *IncrementalProcessor) triggerModelSyncIfNeeded() bool {
+	if p.batchTriggerManual == nil {
```
cliveseldon, line 658: add comment on why we do this for clarity

cliveseldon (approved):

PR #5104, opened 2023-08-23
feat(docs): [SCv1] Automatically create and upload a custom HF model to seldon-models in GCS on every version release

vtaskow:
The next version of MLServer will be allowing users to create custom HuggingFace models. Users will be able to specify a location in which model artefacts will be located and later used. This will allow users to customize certain model parameters, such as tokenizers, feature extractors, image processors, etc and then build a HuggingFace model and use it, not limiting themselves to existing models on the official hub.

Automating the creation and the upload of a simple custom HuggingFace model to our collection of models in our GCS bucket - seldon-models, will pave the way to use said model in upcoming demos, showcasing how users can actually load a custom model and use it in SCv2 or in our paid products.

By having this automation, we ensure that on every new MLServer version release, the custom HF model is added to our collection of models in the bucket and demo pages are kept up-to-date.

By having this change separated from [a similar SCv2-related change](https://github.com/SeldonIO/seldon-core/pull/5103), I am ensuring that SCv1 related demo pages are using models uploaded in non-SCv2 folders in GCS. This will avoid confusion as to  why a SCv1-related demo is using a model uploaded to gs://seldon-models/scv2/... folder.

**Testing**
- [x] The resulted model was generated correctly by running `make env && make train`
- [x] The generated model was uploaded correctly to a public GCS bucket(mine in this case) by running `make upload`
- [x] The uploaded model was used correctly by creating a K8s `SeldonDeployment` resource and was able to correctly make a prediction without any additional downloads
```
apiVersion: machinelearning.seldon.io/v1alpha2
kind: SeldonDeployment
metadata:
  name: huggingface-model
spec:
  protocol: v2
  predictors:
  - graph:
      name: transformer
      implementation: HUGGINGFACE_SERVER
      modelUri: gs://viktor-models/v1.18.0-dev/huggingface/text-generation
      parameters:
      - name: task
        type: STRING
        value: text-generation
    componentSpecs:
      - spec:
          containers:
            - name: transformer
              resources:
                limits:
                  cpu: 1
                  memory: 4Gi
                requests:
                  cpu: 100m
                  memory: 3Gi
    name: default
    replicas: 1

```  
image

Fixes #

### DIFF
--- a/examples/upgrade-to-rclone/global-rclone-upgrade.ipynb
+++ b/examples/upgrade-to-rclone/global-rclone-upgrade.ipynb
@@ -10,7 +10,7 @@
     "In this documentation page we provide an example upgrade path from kfserving-based to rclone-based storage initializer. This is required due to the fact that secret format expected by these two storage initializers is different. \n",
     "\n",
     "Storage initializers are used by Seldon's pre-packaged model servers to download models binaries. \n",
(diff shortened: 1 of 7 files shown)


adriangonz (approved):
Nice one @vtaskow! Looks good to me.

Issue #5043, opened 2023-07-20
Finalisers blocking resource deletion


edfincham:
## Describe the bug
- The finaliser for models can prevent the removal of these resources. The workaround is to manually remove the finaliser from the resource
- This behaviour does not always occur. I have found that it most reliably occurs with model resources after something goes wrong with the model. In my current load-testing use case, the load causes the server pod to restart and reload the model
- To remove the model, I have to manually delete the finaliser. However, after that, when no model resources exist in the namespace, it still shows up in a `"ModelTerminating"` state when listing models via the CLI.
- When I reapply the model manifest, it does not always download the environment tarball (this may be because it still exists on the PVC? But even purging the PVC does not solve the `"ModelTerminating"` state)
- A similar issue can happen with servers, but I can't reliably reproduce it

## Expected behaviour
- `kubectl delete model/name` to remove model resources without issue

## Environment
- EKS (1.21.14)

### Resources
- Nothing fancy here:
```yaml
apiVersion: mlops.seldon.io/v1alpha1
kind: Server
metadata:
  name: mlserver-test
  namespace: seldon-v2
spec:
  serverConfig: mlserver
  replicas: 3
  podSpec:
    serviceAccountName: seldon-v2-sa
    containers:
    - name: mlserver
      env:
      - name: MLSERVER_PARALLEL_WORKERS
        value: "5"
---
apiVersion: mlops.seldon.io/v1alpha1
kind: Model
metadata:
  name: test-model
  namespace: seldon-v2
spec:
  replicas: 3
  storageUri: "s3://seldon-artifact-storage-bucket/model_version_name/model_version_number"
  secretName: "seldon-rclone-secret"
  server: mlserver-test
```

kevinnowland:
I am seeing what I think is a related issue. This is all in kind, deployed with latest helm charts this morning.

Setup:
1. Deploy a server (mlserver in this case)
2. Deploy a model that gets put onto that server

Action:
1. Attempt to (mistakenly) delete server before deleting model. The server does not delete, which is expected, as it has models deployed to it.
2. Attempt to delete model to resolve failed server delete

Result:
1. Model does not delete, server does not delete

Workaround:
1. Delete model finalizer and then model deletes, but server does not yet delete
2. Delete server finalizer and then the server deletes.

Expectation:
Deleting the model should actually delete the model and with all models removed, the sever delete should go through.

Please let me know if my expectation is out of line or my issue is not related to the original post. Thanks!

[...]

PR #5109, opened 2023-08-28
feat: Remove finalizers in V2

cliveseldon:
Work to see if we can remove finalizers for Servers and maybe Models.

Fixes: #5043


Kolajik:
@cliveseldon Any progress on this, please?

dtpryce:
Hey @cliveseldon looks like this is so close to being merged ... any chance of an update here? This small feature currently blocks us using V2 in any shape or form

cliveseldon:
I was hoping to do it for Models as well but can change to what this is just for Servers and can get a review.

PR #5119, opened 2023-09-01
feat(docs): V2D-1268 Use smaller model in SCv1 demo but still reference larger one

vtaskow:
Replacing the massive GPT2 model with a smaller one - tiny stories. The smaller one will be used in Cypress tests. The larger one will still be referenced in case clients would like to run more realistic models. The intention is to tie the Cypress tests together with the Demo pages by using the same GCS models. Tracking future problems will be easier and more visible in case something breaks.

Fixes #


doc/source/servers/huggingface.md
```diff
@@ -78,29 +78,22 @@ You can deploy a custom HuggingFace model by providing the location of the model
...
-                requests:
-                  cpu: 100m
-                  memory: 3Gi
     name: default
     replicas: 1
 ```
+
+````{note}
+As a next step, why not try running a larger-scale model? You can find one in gs://seldon-models/v1.18.0-dev/huggingface/text-gen-custom-gpt2. However, you may need to request more memory!
+````
```
adriangonz, line 99: Nice touch!

adriangonz (approved):
Looks good!

PR #5120, opened 2023-09-01
fix(podmonitor): Pipelinegateway podmonitor label fix

### LINKED ISSUE #5071: Pipeline Gateway Podmonitor label selector does not match Pipeline Gateway labels

cherrymu:
**What this PR does / why we need it**:  This will align with the correct labels that pipeline gateway pod contains and metrics will be scrapped

Fixes #5071 

### DIFF
--- a/prometheus/monitors/pipelinegateway-podmonitor.yaml
+++ b/prometheus/monitors/pipelinegateway-podmonitor.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app.kubernetes.io/name: seldon-pipelinegateway
+      app.kubernetes.io/name: pipelinegateway
   namespaceSelector:
     matchNames: []
     any: false
--- a/prometheus/monitors/pipelinegateway-podmonitor.yaml
+++ b/prometheus/monitors/pipelinegateway-podmonitor.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app.kubernetes.io/name: pipelinegateway
+      app.kubernetes.io/name: seldon-pipelinegateway
   namespaceSelector:
     matchNames: []
     any: false


ChewieSC (approved):
I tested this in Deploy - this fixes the requests not showing for Core v2 pipelines on the dashboard page (plus the graphs below were also empty).
But mb wait with merge for @RafalSkolasinski to approve as well.

RafalSkolasinski (approved):

PR #5129, opened 2023-09-12
refactor(crds): Use built-in OpenAPI validation

jesse-c:
Part of a progressive change to use the CRD validation that's available via kubebuilder [1].

This has the benefits of us:

- Not having to write the code for the validation, since it's generated/handled via OpenAPI
- It's structurally/programatically available in the spec
- Less repetition

This uses the generate `suite_test.go` for now [2].

[1] https://book.kubebuilder.io/reference/markers/crd-validation.html
[2] https://book.kubebuilder.io/cronjob-tutorial/writing-tests

### DIFF
--- a/operator/apis/mlops/v1alpha1/pipeline_types.go
+++ b/operator/apis/mlops/v1alpha1/pipeline_types.go
@@ -74,9 +74,6 @@ const (
 
 type PipelineStep struct {
 	// Name of the step
-	// +kubebuilder:validation:MinLength=1
-	// +kubebuilder:validation:MaxLength=253
-	// +kubebuilder:validation:Pattern=`^[a-z0-9][a-z0-9\-\.]+[a-z0-9]+$`
 	Name string `json:"name"`
 
 	// Previous step to receive data from
--- a/operator/apis/mlops/v1alpha1/pipeline_types.go
+++ b/operator/apis/mlops/v1alpha1/pipeline_types.go
@@ -25,83 +25,6 @@ import (
 	"github.com/seldonio/seldon-core/apis/go/v2/mlops/scheduler"
 )
 
-// PipelineSpec defines the desired state of Pipeline
-type PipelineSpec struct {
-	// External inputs to this pipeline, optional
-	Input *PipelineInput `json:"input,omitempty"`
-	// The steps of this inference graph pipeline
-	Steps []PipelineStep `json:"steps"`
-	// Synchronous output from this pipeline, optional
-	Output *PipelineOutput `json:"output,omitempty"`
-}
-
-type JoinType string
-
-const (
-	JoinTypeInner JoinType = "inner"
-	JoinTypeOuter JoinType = "outer"
-	JoinTypeAny   JoinType = "any"
-)
-
-type PipelineStep struct {
-	// Name of the step
-	Name string `json:"name"`
-	// Previous step to receive data from
-	Inputs []string `json:"inputs,omitempty"`
-	// msecs to wait for messages from multiple inputs to arrive before joining the inputs
-	JoinWindowMs *uint32 `json:"joinWindowMs,omitempty"`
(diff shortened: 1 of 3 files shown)


operator/apis/mlops/v1alpha1/pipeline_types.go
```diff
@@ -25,42 +25,77 @@ import (
 	"github.com/seldonio/seldon-core/apis/go/v2/mlops/scheduler"
 )
 
+//+kubebuilder:object:root=true
```
jesse-c, line 28: **explain:** More conventional kubebuilder annotations to have this for both `struct`s.

operator/apis/mlops/v1alpha1/pipeline_types.go
```diff
@@ -25,42 +25,77 @@ import (
...
 	Steps []PipelineStep `json:"steps"`
+
 	// Synchronous output from this pipeline, optional
 	Output *PipelineOutput `json:"output,omitempty"`
 }
 
+// +kubebuilder:validation:Enum=inner;outer;any
 type JoinType string
 
 const (
```
jesse-c, line 66: **explain:** It's not ideal that it's repeated here, but this approach seems common [1].

[1] https://github.com/search?q=%2Bkubebuilder%3Avalidation%3AEnum&type=code — e.g. Grafana, Cilium, etc.

operator/apis/mlops/v1alpha1/pipeline_types.go
```diff
@@ -25,42 +25,77 @@ import (
...
 	JoinTypeOuter JoinType = "outer"
-	JoinTypeAny   JoinType = "any"
+	// first data input that arrives will be forwarded
+	JoinTypeAny JoinType = "any"
 )
 
 type PipelineStep struct {
 	// Name of the step
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
```
cliveseldon, line 78: why is the max 253. Probably makes sense but as an internal string does it need to be?

jesse-c: I think you're right that it may not need to be!

My understanding last week when I first went through this was that there was some way that the spec field could end up being used somewhere, where k8s would be unhappy with the length.

That was most likely just the pipeline name, and not a pipeline step name though! I'll revert the validations on this field to avoid being overly strict!

jesse-c: ✅ Done!

cliveseldon (approved):

operator/controllers/mlops/suite_test.go
```diff
@@ -78,3 +86,201 @@ var _ = AfterSuite(func() {
...
+				ErrStatus: metav1.Status{
+					TypeMeta: metav1.TypeMeta{Kind: "", APIVersion: ""},
+					ListMeta: metav1.ListMeta{
+						SelfLink:           "",
+						ResourceVersion:    "",
+						Continue:           "",
+						RemainingItemCount: nil,
+					},
+					Status:  "Failure",
+					Message: fmt.Sprintf("Pipeline.mlops.seldon.io \"%s\" is invalid: [metadata.name: Invalid value: \"%s\": must be no more than 253 characters, metadata.name: Invalid value: \"%s\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.steps: Required value]", pipelineName, pipelineName, pipelineName),
```
jesse-c, line 119: **explain:*** https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

PR #5158, opened 2023-11-02
fix(envoy): add readiness probe to Envoy

### LINKED ISSUE #5157: (envoy): Envoy startup probe is not set
## Describe the bug
I have deployed seldon core v2 Release 2.6.0 and have scaled the model-gateway, pipeline-gateway, envoy and dataflow pods to 2 via the SeldonRuntime. All of these pods do have a Pod Disruption Budget of minAvailible: 1 as well, which should not matter here. When I restart an Envoy Pod, I get an GRPC Exception (StatusCode.UNAVAILABLE).

I think this is due to the health endpoints of the envoy pod, which return ready/live despite the envoy pod being not ready. The traffic the newly created pod then receives, results in the failure.

The envoy is ready after the followling log line is printed:
`[info][config] [source/extensions/listener_managers/listener_manager/listener_manager_impl.cc:858] all dependencies initialized. starting workers`

## To reproduce

Scale up the envoy pod to >1 and restart one while doing consistent requests against seldon-mesh
## Expected behaviour

readiness and liveness probes to be configured.

## Environment

* Cloud Provider: AWS
* Kubernetes Cluster Version: 1.25
* Deployed Seldon System Images: 2.6.0
## Model Details 
not relevant

SDJustus:
this PR adds a readiness prove to envoy to prevent it from getting traffic while not being able to handle the requests. Primarily this is needed, if the pod starts/restarts.

Fixes #5157

I was first thinking about adding a startup probe for this, but the kubernetes docs state, that this is mostly done for slow starting services and I don't think the envoy is one. On my cluster, the envoy restart/start takes between 10 and 15 seconds so the initial delay of 10 seconds and 3 retries every 5 seconds should be sufficient. 
I couldn't see any problems with controller after the health check was added.

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
+++ b/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
@@ -1353,6 +1353,13 @@ spec:
           name: http
         - containerPort: 9003
           name: envoy-admin
+        readinessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /ready
+            port: envoy-admin
+          initialDelaySeconds: 10
+          periodSeconds: 5
         resources:
           limits:
             memory: '{{ .Values.envoy.resources.memory }}'
--- a/k8s/yaml/components.yaml
+++ b/k8s/yaml/components.yaml
@@ -976,6 +976,13 @@ spec:
           name: http
         - containerPort: 9003
           name: envoy-admin
+        readinessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /ready
+            port: envoy-admin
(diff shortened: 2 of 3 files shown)


seldondev:
Hi @SDJustus. Thanks for your PR.

<details>

</details>

agrski (approved):
I've done two things to test this, having checked out the PR:
* I've tried running `make -C operator create` and confirmed there are no changes.
* I've tried installing the updated chart into an existing `kind` cluster, confirmed the probe is now set, and confirmed that k8s doesn't mark the pod as ready instantly.  In fact, when port-forwarding the admin port I was able to hit the `/ready` endpoint before it was ready and confirmed the statuses being sent then and after readiness (503 and 200 respectively).

PR #5234, opened 2024-01-26
fix(deps): Upgrade to go.1.20.13 and fix high level CVEs

sakoush:
This PR upgrades go to 1.20.13 and various go packages to fix high level CVEs.

The list of changes are can be checked in the commit log.

Fixes #


sakoush:
ran successfully `samples/pipeline-tests.sh` using a `docker-compose` setup.

apis/go/go.mod
```diff
@@ -3,16 +3,15 @@ module github.com/seldonio/seldon-core/apis/go/v2
 go 1.18
 
 require (
-	google.golang.org/grpc v1.49.0
+	google.golang.org/grpc v1.56.3
```
lc525, line 6: Latest grpc version is 1.61.0 (released 5 days ago) vs 1.56.3 (released 10 Oct 2023). Looking at https://github.com/grpc/grpc-go/releases it seems there have been some security updates & bugfixes in the meantime. I assume 56.3 is what dependabot proposed? 

We could just choose what dependabot offers to keep things simple

apis/go/go.mod
```diff
@@ -3,16 +3,15 @@ module github.com/seldonio/seldon-core/apis/go/v2
...
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.2.0
-	google.golang.org/protobuf v1.28.1
+	google.golang.org/protobuf v1.30.0
```
lc525, line 8: Similarly here, latest protobuf is 1.32.0 : https://pkg.go.dev/google.golang.org/protobuf?tab=versions

hodometer/Dockerfile.hodometer
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20.11-bullseye as builder
+FROM golang:1.20.13-bullseye as builder
```
lc525, line 1: Just wondering out loud here, whether it would make sense to remove very precise tags like these from Dockerfiles, similarly to how you've changed things in the github workflows from 1.20.11 to just 1.20. The equivalent image here would be `golang:1.20-bullseye`. 

This would give us some upgrades for free when rebuilding images, but it does add some risk that things break. In general I think go is better than others in maintaining compatibility, but it's a discussion worth having (whether to take on that risk). 

However, irrespective of the choice, I would apply the same strategy both here and in the github actions to maintain exact compatibility between the tests that github runs and the docker images being built.

hodometer/Dockerfile.receiver
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20.11-alpine AS builder
+FROM golang:1.20.13-alpine AS builder
```
lc525, line 1: Apologies for the long list of comments on rather trivial changes, but I'm lacking context here. I see `golang:*-bullseye`, `golang:*-alpine` and `golang:*` all used in those dockerfiles. We should probably aim to move to a single base unless there are good reasons to keep things different (just to elliminate possible divergences in the compiled binaries)

Generally this should not present a problem with go as we get staticly linked binaries, but if we're using CGO anywhere there will be a dependency on glibc which might differ between debian (bullseye) and redhat/alpine

scheduler/Dockerfile.dataflow
```diff
@@ -14,7 +14,7 @@ RUN gradle build --no-daemon --info
 ################################################################################
 
 # Some dependencies require glibc, which Alpine does not provide
-FROM registry.access.redhat.com/ubi9/openjdk-17-runtime:1.16-3
+FROM registry.access.redhat.com/ubi9/openjdk-17-runtime:1.17-1.1705573249
```
lc525, line 17: This change is related to the dataflow libraries, might be worth to remove it from this PR as I'll have it my PR.

In my changes, I've updated this to `registry.access.redhat.com/ubi9/openjdk-17-runtime:1.17` following the idea I've mentioned above. In fact, I can see dependabot pinning us to even more specific versions here than what we had before.

lc525 (approved):
I added some minor points that might be worth thinking about, if not within this PR at least as a strategy for updating moving forward.

apis/go/go.mod
sakoush: yes this is what dependabot **security** workflow suggested. I still have to go over the dependabot **version** upgrade PRs. I will keep it as is for now as I havent changed it manually.

apis/go/go.mod
sakoush: same as above, will follow dependabot security updates for now.

hodometer/Dockerfile.hodometer
sakoush: For now I pinned it to the same versions. Lets discuss if we can relax it as suggested later. Hopefully this will all be done by dependabot.
I did relax it in github workflow initially but I was not sure completely if they are available in lock steps.

hodometer/Dockerfile.receiver
sakoush: Some images require glibc, which alpine does not provide (based on some comments in the code). I think also the aim was to keep the image as small as possible. We probably can move them all to the same base image if the size difference is not significant. I will leave it to a future PR.

scheduler/Dockerfile.dataflow
sakoush: ok moved to `openjdk-17-runtime:1.17` in [348301f](https://github.com/SeldonIO/seldon-core/pull/5234/commits/348301f94878a160711144c4d624c8504e582ac8)

hodometer/Dockerfile.hodometer
sakoush: as discussed we will relax go patch version in docker images.

Issue #4918, opened 2023-06-15
A mirror in pipeline experiments has a different request ID from the default route 


jklaise:
## Describe the bug
Running an experiment with a mirror between pipelines and inspecting each pipeline after sending a single request reveals that the request IDs between the default route and the mirror are different.

## To reproduce
Follow the "pipeline mirror" example here: https://docs.seldon.io/projects/seldon-core/en/v2/contents/examples/local-experiments.html#pipeline-mirror-experiment

After sending a single request, inspect both pipelines, e.g.:
```bash
seldon pipeline inspect pipeline-add10
```
```
seldon.default.model.add10.inputs	ci5cr6d8gs9c73b1tea0	{"inputs":[{"name":"INPUT", "datatype":"FP32", "shape":["4"], "contents":{"fp32Contents":[1, 2, 3, 4]}}]}
seldon.default.model.add10.outputs	ci5cr6d8gs9c73b1tea0	{"modelName":"add10_1", "modelVersion":"1", "outputs":[{"name":"OUTPUT", "datatype":"FP32", "shape":["4"], "contents":{"fp32Contents":[11, 12, 13, 14]}}]}
seldon.default.pipeline.pipeline-add10.inputs	ci5cr6d8gs9c73b1tea0	{"inputs":[{"name":"INPUT", "datatype":"FP32", "shape":["4"], "contents":{"fp32Contents":[1, 2, 3, 4]}}]}
seldon.default.pipeline.pipeline-add10.outputs	ci5cr6d8gs9c73b1tea0	{"outputs":[{"name":"OUTPUT", "datatype":"FP32", "shape":["4"], "contents":{"fp32Contents":[11, 12, 13, 14]}}]}
```
```bash
seldon pipeline inspect pipeline-mul10
```
```
seldon.default.model.mul10.inputs	ci5cr6d8gs9c73b1teag	{"inputs":[{"name":"INPUT", "datatype":"FP32", "shape":["4"], "contents":{"fp32Contents":[1, 2, 3, 4]}}]}
seldon.default.model.mul10.outputs	ci5cr6d8gs9c73b1teag	{"modelName":"mul10_1", "modelVersion":"1", "outputs":[{"name":"OUTPUT", "datatype":"FP32", "shape":["4"], "contents":{"fp32Contents":[10, 20, 30, 40]}}]}
[...]

PR #5275, opened 2024-01-30
fix(pipelinegateway): Use composite key for Kafka when x-request-id header is specified

KengoA:

## Issue
Currently, there's a bug in the shadow deployment where sending a request to the main pipeline with a `x-request-id` header returns a 504 timeout error. 

In https://github.com/SeldonIO/seldon-core/blob/v2/samples/local-experiments.md, the following inference request to the main pipeline (which is then replicated to the shadow deployment) works if the `x-request-id` is not specified i.e.

```bash
seldon pipeline infer pipeline-add10 -i 1 --inference-mode grpc \
 '{"model_name":"add10","inputs":[{"name":"INPUT","contents":{"fp32_contents":[1,2,3,4]},"datatype":"FP32","shape":[4]}]}'
```

but fails if it's specified
```bash
seldon pipeline infer pipeline-add10 -i 1 --inference-mode grpc --header x-request-id=12345 \
 '{"model_name":"add10","inputs":[{"name":"INPUT","contents":{"fp32_contents":[1,2,3,4]},"datatype":"FP32","shape":[4]}]}'
```
This is due to the requests being dropped from the consumer based on the specified key, which is the same as the request id in the current implementation.

Specifying request ids is a common use case for shadow deployments as suggested in  https://github.com/SeldonIO/seldon-core/issues/4918#issuecomment-1592643099.

## Fix

This PR fixes this issue by creating a composite key for multi topic consumers with the pipeline name (i.e. `add10-pipeline.12345` for default pipeline and `mul10-pipeline.12345` for the shadow pipeline such that the requests that still need to be processed are not dropped.

## Steps to reproduce
```
$ cd samples
$ seldon model load -f ./models/add10.yaml
$ seldon model load -f ./models/mul10.yaml
$ seldon pipeline load -f ./pipelines/add10.yaml
$ seldon pipeline load -f ./pipelines/mul10.yaml
$ seldon experiment start -f ./experiments/addmul10-mirror.yaml 
$ seldon pipeline infer pipeline-add10 -i 1 --inference-mode grpc --header x-request-id=12345 \
 '{"model_name":"mul10","inputs":[{"name":"INPUT","contents":{"fp32_contents":[1,2,3,4]},"datatype":"FP32","shape":[4]}]}'
```

### DIFF
--- a/scheduler/pkg/kafka/pipeline/kafkamanager.go
+++ b/scheduler/pkg/kafka/pipeline/kafkamanager.go
@@ -202,7 +202,7 @@ func (km *KafkaManager) Infer(
 		km.mu.RUnlock()
 		return nil, err
 	}
-	key := requestId
+	key := fmt.Sprintf("%s.%s", resourceName, requestId)
 	request := &Request{
 		active: true,
 		wg:     new(sync.WaitGroup),
@@ -228,7 +228,7 @@ func (km *KafkaManager) Infer(
 	}
 
 	ctx, span := km.tracer.Start(ctx, "Produce")
-	span.SetAttributes(attribute.String(util.RequestIdHeader, key))
+	span.SetAttributes(attribute.String(util.RequestIdHeader, requestId))
 	// Add trace headers
 	carrier := splunkkafka.NewMessageCarrier(msg)
 	otel.GetTextMapPropagator().Inject(ctx, carrier)
@@ -246,9 +246,9 @@ func (km *KafkaManager) Infer(
 		span.End()
 	}()
 	km.mu.RUnlock()
-	logger.Debugf("Waiting for response for key %s", key)
+	logger.Debugf("Waiting for response for request id %s", requestId)
 	request.wg.Wait()
-	logger.Debugf("Got response for key %s", key)
+	logger.Debugf("Got response for request id %s", requestId)
 	return request, nil
 }
 
--- a/scheduler/pkg/kafka/pipeline/kafkamanager.go
+++ b/scheduler/pkg/kafka/pipeline/kafkamanager.go
@@ -202,27 +202,28 @@ func (km *KafkaManager) Infer(
 		km.mu.RUnlock()
 		return nil, err
 	}
-	key := getCompositeKey(resourceName, requestId, ".")
+	// Use composite key to differentiate multiple piplines (i.e. mirror) using the same message
+	compositeKey := getCompositeKey(resourceName, requestId, ".")
 	request := &Request{
 		active: true,
 		wg:     new(sync.WaitGroup),
-		key:    key,
+		key:    compositeKey,
 	}
-	pipeline.consumer.requests.Set(key, request)
-	defer pipeline.consumer.requests.Remove(key)
+	pipeline.consumer.requests.Set(compositeKey, request)
+	defer pipeline.consumer.requests.Remove(compositeKey)
 	request.wg.Add(1)
 
 	outputTopic := km.topicNamer.GetPipelineTopicInputs(resourceName)
 	if isModel {
 		outputTopic = km.topicNamer.GetModelTopicInputs(resourceName)
 	}
-	logger.Debugf("Produce on topic %s with key %s", outputTopic, key)
+	logger.Debugf("Produce on topic %s with key %s", outputTopic, compositeKey)
 	kafkaHeaders := append(headers, kafka.Header{Key: resources.SeldonPipelineHeader, Value: []byte(resourceName)})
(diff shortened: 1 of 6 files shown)


scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -202,7 +202,7 @@ func (km *KafkaManager) Infer(
 		km.mu.RUnlock()
 		return nil, err
 	}
-	key := requestId
+	key := getCompositeKey(resourceName, requestId, ".")
```
sakoush, line 205: nit: just for legibility could you rename `key` to `compositeKey` and also add a note why we are doing it?

scheduler/pkg/kafka/pipeline/multi_topic_consumer.go
```diff
@@ -156,12 +156,17 @@ func (c *MultiTopicsKafkaConsumer) pollAndMatch() error {
...
 			if val, ok := c.requests.Get(string(e.Key)); ok {
-				// Add tracing span
 				ctx := context.Background()
 				carrierIn := splunkkafka.NewMessageCarrier(e)
 				ctx = otel.GetTextMapPropagator().Extract(ctx, carrierIn)
+
+				// Add tracing span
 				_, span := c.tracer.Start(ctx, "Consume")
-				span.SetAttributes(attribute.String(util.RequestIdHeader, string(e.Key)))
+				requestId := getRequestIdFromKafkaHeaders(e.Headers)
```
sakoush, line 165: maybe add a note that the key is composite and cannot be used as the request id?

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -228,7 +228,7 @@ func (km *KafkaManager) Infer(
 	}
 
 	ctx, span := km.tracer.Start(ctx, "Produce")
-	span.SetAttributes(attribute.String(util.RequestIdHeader, key))
+	span.SetAttributes(attribute.String(util.RequestIdHeader, requestId))
```
sakoush, line 231: do we want to differentiate between the span of the main and shadow pipeline? I am thinking that the shadow pipeline might be using a faster model perhaps?

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -246,9 +246,9 @@ func (km *KafkaManager) Infer(
 		span.End()
 	}()
 	km.mu.RUnlock()
-	logger.Debugf("Waiting for response for key %s", key)
+	logger.Debugf("Waiting for response for request id %s", requestId)
```
sakoush, line 249: do we want to log `resourceName` as well?

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -246,9 +246,9 @@ func (km *KafkaManager) Infer(
...
 	request.wg.Wait()
-	logger.Debugf("Got response for key %s", key)
+	logger.Debugf("Got response for request id %s", requestId)
```
sakoush, line 251: ditto

sakoush (approved):
I left some minor comments, but it looks fine to me once they are addressed.

This fix might have also addressed potentially another bug when we have joins in the mirror experiment as the message in kafka cannot be similarly differentiated per pipeline if the requestid is set

scheduler/pkg/kafka/pipeline/kafkamanager.go
KengoA: Added in https://github.com/SeldonIO/seldon-core/pull/5275/commits/6dc56b2cea9c532879e0f4c2bf6e77c812999411

scheduler/pkg/kafka/pipeline/kafkamanager.go
KengoA: Yeah that's a good point, but in that case we need to set a new attribute for the resource name and that affects other places as well. @lc525 What do you think?

KengoA: There will be two separate paths in the tracing tree for shadow and main pipeline which have the same request id as an attribute. This will be followed up if we do need a different attribute key to differentiate those.

KengoA: Note that this logic is now applied to spans in gateway https://github.com/SeldonIO/seldon-core/pull/5275/commits/cc0e9a6d937a90bcbf347e84e6c9b8ab0a951d9b

sakoush (approved):

PR #5277, opened 2024-01-31
fix: Add sleep to pipeline-tests script

sakoush:

Adding a sleep (defaulting to 5 secs) between different tests in the pipeline script to let it teardown the resources.

Fixes #


lc525 (approved):
lgtm

Issue #5189, opened 2024-01-19
otel-collector-1 container not starting: error decoding 'exporters': unknown type: "jaeger" for id: "jaeger"


avolts:
## Describe the bug
After following the installation guide for Docker Compose for SeldenCoreV2, the otel-collector-1 container cannot start due to the following error:
error decoding 'exporters': unknown type: "jaeger" for id: "jaeger" (valid values: [opensearch prometheus awsxray azuredataexplorer azuremonitor carbon cassandra mezmo sentry file honeycombmarker opencensus sumologic sapm skywalking alertmanager alibabacloud_logservice awscloudwatchlogs dataset googlecloud pulsar datadog f5cloud loadbalancing logicmonitor logzio logging clickhouse loki debug otlp awsemf coralogix instana tencentcloud_logservice dynatrace elasticsearch googlecloudpubsub prometheusremotewrite signalfx otlphttp awskinesis awss3 googlemanagedprometheus influxdb kafka splunk_hec syslog zipkin])

## To reproduce
1. Follow the installation instructions for Docker Compose [https://docs.seldon.io/projects/seldon-core/en/v2/contents/getting-started/docker-installation/index.html](url) and set CUSTOM_IMAGE_TAG=2.6.0
2. Check running containers in Docker Desktop
3.  See that otel-collector-1 cannot start 

## Expected behaviour
I expect that the otel-collector-1 container is starting without issues after local deployment.
Replacing jaeger with otlp at the marked places in the otel-collector-config.yaml solves the issue.
"jaeger" is no longer in the exporters values.
![image](https://github.com/SeldonIO/seldon-core/assets/22810956/329a1a41-1141-413a-90e3-0e6dd1ff1c99)

## Environment

- SeldonCoreV2 Installation on WSL Ubuntu-22.04
- Docker Desktop v4.24.2
- Windows Windows 10 Enterprise, 22H2

sakoush:
A potential fix is described [here](https://github.com/SeldonIO/seldon-core/pull/5170)

a workaround for now is to use an older version of otel/jeager via setting `JAEGER_IMG=jaegertracing/all-in-one:1.49`

lc525:
This is believed to be fixed via #5291

PR #5170, opened 2023-12-12
fix(scheduler): update otel-collector-config after jaeger endpoint does not supported

ducminhle:
**What this PR does / why we need it**:
With existing config: otel-collector-config.yaml, I got an error when try to run the `make deploy-local`, `otel-collector` container crashed with logs:

```
Error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'exporters': unknown type: "jaeger" for id: "jaeger" (valid values: [file prometheusremotewrite skywalking splunk_hec cassandra dataset influxdb kafka awss3 otlp azuremonitor googlecloudpubsub loadbalancing zipkin logging datadog elasticsearch sapm sentry sumologic tanzuobservability tencentcloud_logservice debug carbon clickhouse mezmo prometheus azuredataexplorer awsxray f5cloud googlecloud pulsar syslog alibabacloud_logservice coralogix logzio loki opensearch awskinesis awscloudwatchlogs awsemf dynatrace googlemanagedprometheus honeycombmarker instana logicmonitor otlphttp signalfx opencensus])
2023/12/11 06:42:49 collector server run finished with error: failed to get config: cannot unmarshal the configuration: 1 error(s) decoding:

* error decoding 'exporters': unknown type: "jaeger" for id: "jaeger" (valid values: [file prometheusremotewrite skywalking splunk_hec cassandra dataset influxdb kafka awss3 otlp azuremonitor googlecloudpubsub loadbalancing zipkin logging datadog elasticsearch sapm sentry sumologic tanzuobservability tencentcloud_logservice debug carbon clickhouse mezmo prometheus azuredataexplorer awsxray f5cloud googlecloud pulsar syslog alibabacloud_logservice coralogix logzio loki opensearch awskinesis awscloudwatchlogs awsemf dynatrace googlemanagedprometheus honeycombmarker instana logicmonitor otlphttp signalfx opencensus])
```

I found the problem: `opentelemetry-collector-contrib` has removed jaeger and jaegerthrifthttp exporters from this commit: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/26546. We need to update the endpoint to fix this issue (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/15291#issuecomment-1826734333)

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

PR #5291, opened 2024-02-05
fix(logging): Update otel and tracing config

lc525:
- fixes dataflow engine logging errors by specifying the otel exporter protocol

the updates are needed because the `opentelemetry-java-instrumentation` library requires a http(s) URI for the otel collector endpoint, regardless of the actual protocol. As the default for us is grpc, explicitly set the OTEL_EXPORTER_OTLP_PROTOCOL environment variable on dataflow pods

- fixes otel-collector config to remove deprecated jagger exporter (jagger now supports otel directly).

- add the OTEL_EXPORTED_OTLP_PROTOCOL key to the seldon-tracing configMap and update the operator, crds and helm charts to support getting the value for this key from tracing config, similarly to how OTEL_EXPORTER_OTLP_PROTOCOL is fetched

- update versions used by ansible for jagger and opentelemetry-operator
- port 14250 no longer needs to be exposed under any config
- fix dependency ordering for dataflow/gradle.
  - previous ordering caused kafka-streams not to be able to find the slf4j logging provider
  - this lead to logs produced by kafka-streams not being recorded

Fixes #
Internal issue references:
- #INFRA-568 Jagger latest is crashing
- #INFRA-464 Otel is not able to parse its config (deprecated exporters)

Public issues:
- #5189
- related public PR: #5170 

- helm chart updates generated via makefile; in order to ensure results consistent to previous helm charts (i.e descriptions that don't disappear), I had to add some docs / update the makefile; those changes are unrelated to the primary goal of the PR.
- k8s/yamls/* - auto generated via makefile


k8s/helm-charts/seldon-core-v2-runtime/templates/seldon-runtime.yaml
```diff
@@ -55,4 +55,5 @@ spec:
     tracingConfig:
       disable: {{ .Values.config.tracingConfig.disable }}
       otelExporterEndpoint: {{ .Values.config.tracingConfig.otelExporterEndpoint }}
+      otelExporterProtocol: {{ .Values.config.tracingConfig.otelExporterProtocol }}
```
lc525, line 58: @RafalSkolasinski following this change, should we update any of the helm chart versions? I don't know how helm chart version updates are handled, or if there is a policy on updating them. Atm, for example, I've seen that the appVersion has nothing to do with the Seldon Core version. Happy to follow what you suggest here.

operator/config/crd/bases/mlops.seldon.io_pipelines.yaml
lc525, line 1: The descriptions I've added here were in the current helm-charts but not in the sources that generate them via kustomize (i.e. this file). I've added them because otherwise those descriptions would have disappeared on re-generation via the Makefile.

scheduler/otel-collector-config.yaml
```diff
@@ -9,14 +9,14 @@ exporters:
     const_labels:
       provider: seldon
 
-  logging:
+  debug:
```
lc525, line 12: the logger exporter is deprecated and replaced by debug

sakoush (approved):
lgtm

k8s/Makefile
```diff
@@ -25,7 +25,7 @@ create-helm-charts:
 
 .PHONY: create-yaml
 create-yaml:
-	helm template seldon-core-v2-certs ./helm-charts/seldon-core-v2-certs | grep -v "namespace:" > yaml/certs.yaml
+	helm template -n seldon-mesh seldon-core-v2-certs ./helm-charts/seldon-core-v2-certs | grep -v "namespace:" > yaml/certs.yaml
```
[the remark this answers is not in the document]
lc525: Concretely, I don't know. However, the _currently generated_ certs helm chart (and the .yaml files) have seldon-mesh, so not adding the namespace here would have changed lots of dnsNames from `seldon-mesh` to `default` on re-generation. I've simply added this here to avoid _any_ changes to the generated certs.yaml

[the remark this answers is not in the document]
lc525: So, I've checked. In all the other ones `{{ .Release.Namespace }}` only appears as a value of the `namespace` key, which is removed by the `grep -v`, so the resulting yamls do not end up containing any hardcoded values. In that sense they are generic.

k8s/yaml/crds.yaml
lc525, line 1: file generated from helm charts via Makefile (`make create-yaml`)

PR #5292, opened 2024-02-05
fix(dataflow, controller): Connection retries when scheduler restarts

sakoush:

In some cases during  testing, `dataflow-engine` didnt reconnect to the new scheduler pod after a rolling update. The `controller` also suffers from the same issue. The fix is to retry (for now indefinitely) which is the same pattern used in `agent` ([link](https://github.com/seldonio/seldon-core/blob/acc55f09330f1a40564f7955ba34c8b1bad42840/scheduler/pkg/agent/client.go#L217)), `model-gateway` ([link](https://github.com/seldonio/seldon-core/blob/v2/scheduler/pkg/kafka/gateway/client.go)) and `pipeline-gateway` ([link](https://github.com/seldonio/seldon-core/blob/v2/scheduler/pkg/kafka/pipeline/status/client.go)).

Summary of changes:
- Add indefinite retry to `subscribePipelines` (dataflow)
- Add indefinite retry to `startEventHanders` (controller)

Fixes INFRA-713 (internal)

### DIFF
--- a/operator/scheduler/client.go
+++ b/operator/scheduler/client.go
@@ -106,7 +106,7 @@ func (s *SchedulerClient) RemoveConnection(namespace string) {
 	}
 }
 
-// A smoke test allows us to quicky check if we actually have a functional grpc connection to the scheduler
+// A smoke test allows us to quickly check if we actually have a functional grpc connection to the scheduler
 func (s *SchedulerClient) smokeTestConnection(conn *grpc.ClientConn) error {
 	grcpClient := scheduler.NewSchedulerClient(conn)
 
--- a/operator/scheduler/client.go
+++ b/operator/scheduler/client.go
@@ -44,7 +44,7 @@ type SchedulerClient struct {
 	mu               sync.Mutex
 }
 
-//  connect on demand by add getConnection(namespace) which if not existing calls connect to sheduler.
+//  connect on demand by add getConnection(namespace) which if not existing calls connect to scheduler.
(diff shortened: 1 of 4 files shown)


operator/scheduler/client.go
```diff
@@ -66,30 +67,68 @@ func getSchedulerHost(namespace string) string {
...
 
+// startEventHanders starts the grpc stream connections to the scheduler for the different resources we care about
+// we also add a retry mechanism to reconnect if the connection is lost, this can happen if the scheduler is restarted
+// or if the network connection is lost. We use an exponential backoff to retry the connection.
+// note that when the scheduler is completely dead we will be not be able to reconnect and these go routines will retry forever
+// TODO: add a max retry count and report back to the caller.
 func (s *SchedulerClient) startEventHanders(namespace string, conn *grpc.ClientConn) {
+	retryFn := func(fn func(context context.Context, conn *grpc.ClientConn) error, context context.Context, conn *grpc.ClientConn) error {
+		logFailure := func(err error, delay time.Duration) {
+			s.logger.Error(err, "Scheduler not ready")
```
lc525, line 78: One of the issues here is that we'll probably be seeing each error 4 times when the scheduler is down (once from each of the coroutines below), but I'm not sure if we can sort that out easily.

lc525 (approved):
lgtm; This clearly fixes a large class of the restart issues we've been seeing. I'm still not entirely sure about the grpc.Dial and failures there, but if anything pops up it can be solved in a different PR, without delaying this one further.

PR #5295, opened 2024-02-06
fix(docs): Update docs w latest K8s compatibility

paulb-seldon:
**What this PR does / why we need it**: Update the docs to show what version of K8s core 1.18 is compatible with


RafalSkolasinski (approved):
LGTM

PR #5298, opened 2024-02-08
fix: Pipeline state during disconnects

sakoush:

This PR fixes issues with regards to pipeline state inconsistency in cases of components failures (e.g. dataflow-engine, scheduler).

Fixes INFRA-716 (internal)

## Changes:
- Do not remove a pipeline from the scheduler local persistence store upon the user deleting this particular pipeline. This is required because if the scheduler restarts and before the pipeline delete is propagated and acknowledged by the different services (e.g. dataflow-engine and controller) then there is a risk that this state will become inconsistent. This has the drawback though that the internal state of the scheduler indefinitely increasing so we need to recycle it in a follow up work. For now we prioritise consistency over storage as we expect storage of protobuf control plan messages of these pipelines to be low.
- Fix a bug in dataflow-engine to report success when the pipeline topology is already running.
- Fix a bug in controller to not remove the finaliser of a pipeline if its state is `PipelineTerminate` (i.e. still to terminate).
- Return `PipelineTerminating` pipelines when calling `GetAllRunningPipelineVersions`, which allow us to handle the following case.
- For `PipelineTerminating` pipelines and no currently available dataflow-engines, set them to `PipelineTerminated`.
- If receiving an event for a pipeline with state `PipelineTermiante` or `PipelineTerminating` and no currently available dataflow-engines, set them to `PipelineTerminated`.
- We always send all pipelines (event deleted ones) when the controller connects as it might be that some messages have been missed during the disconnect.
## Testing

- Induced a 2 min delay in datafllow pipeline create and delete (to allow time to kill components).
```
    private suspend fun handleDelete(metadata: PipelineMetadata) {
        logger.info("Delete pipeline ${metadata.name} version: ${metadata.version} id: ${metadata.id}")
        Thread.sleep(120_000)

    private suspend fun handleCreate(
        metadata: PipelineMetadata,
        steps: List<PipelineStepUpdate>,
        kafkaConsumerGroupIdPrefix: String,
        namespace: String,
    ) {
        logger.info("Create pipeline ${metadata.name} version: ${metadata.version} id: ${metadata.id}")
        Thread.sleep(120_000)
```
- Killing dataflow-engine, controller and scheduler pods, them make sure state is eventually consistent.

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
@@ -138,7 +138,7 @@ class PipelineSubscriber(
             }
         } else {
             pipelineStarted = true
-            logger.warn("pipeline ${metadata.id} already exists")
+            logger.warn("pipeline ${metadata.name} with id ${metadata.id} already exists")
         }
 
         client.pipelineUpdateEvent(
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
@@ -53,9 +53,10 @@ class PipelineSubscriber(
     private val pipelines = ConcurrentHashMap<PipelineId, Pipeline>()
 
     suspend fun subscribe() {
-        logger.info("will connect to ${upstreamHost}:${upstreamPort}")
         while (true) {
+            logger.info("will connect to ${upstreamHost}:${upstreamPort}")
             retry(binaryExponentialBackoff(50..5_000L)) {
+                logger.debug("retrying to connect to ${upstreamHost}:${upstreamPort}")
                 subscribePipelines(kafkaConsumerGroupIdPrefix, namespace)
             }
         }
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
@@ -137,6 +137,7 @@ class PipelineSubscriber(
                 updateEventReason = "kafka topic error"
             }
         } else {
+            pipelineStarted = true
             logger.warn("pipeline ${metadata.id} already exists")
         }
 
--- a/scheduler/pkg/kafka/dataflow/server.go
+++ b/scheduler/pkg/kafka/dataflow/server.go
@@ -396,7 +396,7 @@ func (c *ChainerServer) handlePipelineEvent(event coordinator.PipelineEventMsg) 
 			logger.WithField("pipeline", event.PipelineName).Warn(errMsg)
 
 			status := pv.State.Status
-			// if not dataflow engines available then we think we can terminate. however it might be a networking glitch
+			// if no dataflow engines available then we think we can terminate. however it might be a networking glitch
 			if pv.State.Status == pipeline.PipelineTerminating {
 				status = pipeline.PipelineTerminated
 			}
--- a/scheduler/pkg/kafka/dataflow/server.go
+++ b/scheduler/pkg/kafka/dataflow/server.go
@@ -325,8 +325,10 @@ func contains(slice []string, val string) bool {
 
 func (c *ChainerServer) rebalance() {
 	logger := c.logger.WithField("func", "rebalance")
+	// note that we are not retrying PipelineFailed pipelines, consider adding this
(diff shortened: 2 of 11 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -136,6 +137,7 @@ class PipelineSubscriber(
                 updateEventReason = "kafka topic error"
             }
         } else {
+            pipelineStarted = true
             logger.warn("pipeline ${metadata.id} already exists")
```
lc525, line 141: Should we put log the pipeline name here instead of id? or perhaps the pair (name, id). Just to make it more readable

scheduler/pkg/kafka/dataflow/server.go
```diff
@@ -393,12 +400,17 @@ func (c *ChainerServer) handlePipelineEvent(event coordinator.PipelineEventMsg)
 			errMsg := "no dataflow engines available to handle pipeline"
 			logger.WithField("pipeline", event.PipelineName).Warn(errMsg)
 
-			err := c.pipelineHandler.SetPipelineState(pv.Name, pv.Version, pv.UID, pv.State.Status, errMsg, sourceChainerServer)
+			status := pv.State.Status
+			// if no dataflow engines available then we think we can terminate. however it might be a networking glitch
```
lc525, line 404: do we deal with the network glitch case?

scheduler/pkg/kafka/dataflow/server.go
```diff
@@ -393,12 +401,18 @@ func (c *ChainerServer) handlePipelineEvent(event coordinator.PipelineEventMsg)
 			errMsg := "no dataflow engines available to handle pipeline"
 			logger.WithField("pipeline", event.PipelineName).Warn(errMsg)
 
-			err := c.pipelineHandler.SetPipelineState(pv.Name, pv.Version, pv.UID, pv.State.Status, errMsg, sourceChainerServer)
+			status := pv.State.Status
+			// if no dataflow engines available then we think we can terminate pipelines.
+			// TODO: however it might be a networking glitch and we need to handle this better in future
```
sakoush, line 406: I think in the case of a networking glitch, then no the pipelines are going to remain in dataflow engine and not removed. We could repay all pipeline control plane messages up to a specific time and therefore could deal with glitches but will leave it to a follow up PR.

scheduler/pkg/kafka/dataflow/server.go
```diff
@@ -420,7 +434,7 @@ func (c *ChainerServer) handlePipelineEvent(event coordinator.PipelineEventMsg)
 			if err != nil {
 				logger.WithError(err).Errorf("Failed to set pipeline state to terminating for %s", pv.String())
 			}
-			msg := c.createPipelineMessage(pv)
+			msg := c.createPipelineMessage(pv) // note pv is a copy and does not include the new change to terminating state
```
lc525, line 437: This is not an issue right? I see it could lead to duplicate logs with status = pipeline.PipelineTerminated, but not sure of other implications

scheduler/pkg/store/pipeline/store.go
```diff
@@ -293,7 +291,8 @@ func (ps *PipelineStore) GetAllRunningPipelineVersions() []coordinator.PipelineE
 	for _, p := range ps.pipelines {
 		pv := p.GetLatestPipelineVersion()
 		switch pv.State.Status {
-		case PipelineCreate, PipelineCreating, PipelineReady:
+		// we consider PipelineTerminating as running as it is still active
+		case PipelineCreate, PipelineCreating, PipelineReady, PipelineTerminating:
```
lc525, line 295: Would it be useful to document the meaning of the Terminating states?
The way I see it now:
- Terminate: Received signal that pipeline should be terminated
- Terminating: In the process of doing cleanup/housekeeping for pipeline termination.
- Terminated: All finished

Would that be an accurate description? How long do you think the "Terminating" state would last?

lc525 (approved):
lgtm; left some minor observations/comments

scheduler/pkg/store/pipeline/store.go
sakoush: done in [ac56141](https://github.com/SeldonIO/seldon-core/pull/5298/commits/ac5614142e8cf43948478a75dbe02fe847567acd)

scheduler/pkg/kafka/dataflow/server.go
sakoush: It is not an issue in this particular case as we send the event (`PipelineTerminate`) to dataflow-engine and mark it in the scheduler to be `PipelineTerminating` until dataflow responds back with success or failure.

PR #5307, opened 2024-02-12
fix: Add kubectl option to pipeline-tests script

sakoush:

Adding the option to run `pipeline-tests.sh` with kubectl and therefore smoke testing the controller.

Fixes #

### DIFF
--- a/samples/pipeline-tests.sh
+++ b/samples/pipeline-tests.sh
@@ -32,6 +32,9 @@ function load() {
       elif [ $1 == "pipeline" ]
       then
             seldon pipeline load -f $2
+      elif [ $1 == "experiment" ]
+      then
+            seldon experiment start -f $2
       fi
   fi
 }
@@ -47,6 +50,9 @@ function unload() {
       elif [ $1 == "pipeline" ]
       then
             seldon pipeline unload $2
+      elif [ $1 == "experiment" ]
+      then
+            seldon experiment stop $2
       fi
   fi
 }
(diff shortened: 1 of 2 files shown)


samples/pipeline-tests.sh
```diff
@@ -1,110 +1,196 @@
...
+else
+      namespace=$3
+fi
+
+function load() {
+  if [ $cmd == "kubectl" ]
+  then
+      kubectl apply -f $2 -n $namespace
+  else
+      if [ $1 == "model" ]
```
lc525, line 29: Should those be $2 as $1 is the sleepTime?

samples/pipeline-tests.sh
```diff
@@ -1,110 +1,196 @@
...
+fi
+
+if [ -z "$3" ]
+then
+      namespace="seldon-mesh"
+else
+      namespace=$3
+fi
+
+function load() {
```
lc525, line 24: Is the goal here to not need to set `seldon` environment variables pointing to the k8s svc? If yes perhaps we should also have the equivalent of `seldon status` commands using kubectl.

sakoush: good point, I will add it.

samples/pipeline-tests.sh
```diff
@@ -1,110 +1,196 @@
...
 seldon pipeline infer trigger-joins --inference-mode grpc  '{"model_name":"pipeline","inputs":[{"name":"ok3","contents":{"fp32_contents":[1]},"datatype":"FP32","shape":[1]},{"name":"INPUT","contents":{"fp32_contents":[1,2,3,4]},"datatype":"FP32","shape":[4]}]}'
-seldon pipeline unload trigger-joins
-seldon model unload mul10
-seldon model unload add10
+unload pipeline trigger-joins ./pipelines/trigger-joins.yaml
+unload model mul10 ./models/mul10.yaml
+unload model add10 ./models/add10.yaml
+
+
+# MLServer
```
lc525, line 165: Perhaps we should rename the file to "smoke-tests.sh"? (given that now we also have MLServer and Experiments)

samples/pipeline-tests.sh
```diff
@@ -1,110 +1,196 @@
...
+  '{"inputs": [{"name": "predict", "shape": [1, 4], "datatype": "FP32", "data": [[1, 2, 3, 4]]}]}'
+seldon model infer iris --show-headers \
+  '{"inputs": [{"name": "predict", "shape": [1, 4], "datatype": "FP32", "data": [[1, 2, 3, 4]]}]}'
+seldon model infer iris -s -i 50 \
+  '{"inputs": [{"name": "predict", "shape": [1, 4], "datatype": "FP32", "data": [[1, 2, 3, 4]]}]}' 
+seldon model infer iris --inference-mode grpc -s -i 50\
+   '{"model_name":"iris","inputs":[{"name":"input","contents":{"fp32_contents":[1,2,3,4]},"datatype":"FP32","shape":[1,4]}]}'
+unload experiment experiment-sample ./experiments/ab-default-model.yaml
+unload model iris ./models/sklearn1.yaml
+unload model iris2 ./models/sklearn2.yaml
```
lc525, line 196: nit; github complains that there's no newline at end of file

lc525 (approved):
lgtm, just added some small suggestions and an idea of improving this further in the future.

samples/pipeline-tests.sh
sakoush: done

samples/pipeline-tests.sh
sakoush: done in [70285e0](https://github.com/SeldonIO/seldon-core/pull/5307/commits/70285e0ae1d4a3a992ea4090659fbf78e79d8871)

PR #5318, opened 2024-02-12
build(ansible): use of local dev images in kind cluster

lc525:
Introduce the ability to declaratively specify, via an ansible-playbook extra vars file, sets of components that should be built locally from the workdir and pushed into kind.

The kind Seldon Core v2 install is then configured to use those dev images.

Full documentation included in `README.dev.md`


ansible/playbooks/vars/set-custom-images.yaml
```diff
@@ -1,86 +1,239 @@
+# Configure your custom seldon deployment via ansible
```
RafalSkolasinski, line 1: Should potentially a simple readme.md work better with few, ready to copy-paste and modify examples?

ansible/roles/seldon/defaults/main.yaml
```diff
@@ -8,9 +8,83 @@ seldon_install_runtime: true
 seldon_install_servers: true
 
 seldon_force_install_crds: false
+seldon_force_install_components: false
```
RafalSkolasinski, line 11: In other Ansible playbooks we tend to use a single flag
```yaml
helm_force_install: false
```
that is used for every Helm installation - unless we need to be more granular potentially single flag is easier?

ansible/roles/seldon/defaults/main.yaml
```diff
@@ -8,9 +8,83 @@ seldon_install_runtime: true
...
+
+seldon_default_dev:
+  dockerhub_user: "seldonio"
+  install_kind_images: true
+  skip_build_images: false
+  save_helm_components_overrides: false
+  save_helm_components_overrides_file: "{{ ansible_env.HOME }}/seldon_helm_comp_values.yaml"
+
+seldon_default_image_config:
+  registry: "{{ omit }}"
```
RafalSkolasinski, line 27: Does `{{ omit }}` carry any special meaning?

lc525: It acts as the key hasn't been set (in this case, the registry key will not appear in the helm chart config unless overwritten by the extra vars file). As opposed to it being set to an empty value.

ansible/roles/seldon/defaults/main.yaml
lc525: In the `seldon` role here (ansible for core v2) there was only one "force" option available, namely `seldon_force_install_crds`. For the others, no option was available, and nothing was being passed to the corresponding task. 

I typically just force the components when doing incremental deployments (i.e running the ansible playbook over an existing kind cluster), hence the finer granularity.

But I can set them all by default to a common`helm_force_install` and let the dev choose

ansible/playbooks/vars/set-custom-images.yaml
lc525: I'll add that, would indeed be better.

ansible/roles/seldon/defaults/main.yaml
RafalSkolasinski: hmm.... what we also tend to have in other place and may be useful here are flags to disable installation of all components except one, we have for example 
```yaml
full_install: yes

install_istio: "{{ full_install }}"
install_nginx: "{{ full_install }}"
install_knative_eventing: "{{ full_install }}"
install_knative_serving: "{{ full_install }}"

...
```

and one can do `-e full_install=no -e install_istio=yes`

lc525: This is the same here. This PR retains all the normal core v2 ansible install features (which has those options) and just adds additional config capabilities via an extra config vars file. 

In fact, providing some of those `install_...` options within this extra vars file helps express everything declaratively in one place. For example, an extra vars file that spins up a cluster connected to Confluent Kafka would also have `install_kafka: false` inside so as to not use strimzi.

RafalSkolasinski: Ah yes, it is just `full_install` being defined in entrypoint playbook and not deeper in the role like in other place, indeed...

Then single `helm_force_install` should suffice and help overall keep number of variables lower

lc525: I've added the `helm_force_install` top-level option. The individual `seldon_force...` variables now default to its value (similar to the `full_install` case)

ansible/roles/seldon/tasks/set_value_overrides.yaml
```diff
@@ -53,3 +125,7 @@
            {'imagePullSecrets': [{ 'name': custom_image_pull_secrets.name }]},
            recursive=true, list_merge="append")}}
   when: custom_image_pull_secrets is defined
+
+# - name: Debug
```
sakoush, line 129: remove if not required.

sakoush (approved):
great!

ansible/README.dev.md
```diff
@@ -0,0 +1,630 @@
...
+```{warning}
+Please install Ansible as described in README.md before following the instructions in this file
+```
+
+In all the cases described below, save your configuration into a file (e.g.
+`~/seldon-configs/my-dev-config.yaml`) then pass this file as an extra configuration variables
+file when running ansible playbooks. Please note the `@` before the file path, which is required:
+
+```bash
+ansible-playbook playbooks/seldon-all.yaml -e @<path-to-custom-images-config.yaml>
```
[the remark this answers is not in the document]
lc525: Yes. For GCP and dev images, we would only need the addition of a task to push locally-built images into a GCP-project artifact registry. For available non-dev images, this should already be supported as long as kubectl context points to the right cluster and one doesn't run the `seldon-all` playbook (which creates a kind cluster and sets that as the new context) but instead runs the other individual playbooks while passing the custom config file via -e.

ansible/roles/seldon/defaults/main.yaml
```diff
@@ -7,10 +7,87 @@ seldon_install_components: true
...
+    image_repository: "seldon-rclone"
+  grafana:
+    helm_config_type: standalone
+    src: "{{ repo_root_dir }}/scheduler"
+    build_target: "docker-build-grafana"
+    image_repository: "grafana"
+  mlserver:
+    helm_config_type: serverConfig
+    src: "{{ seldon_mlserver_src | default('') }}"
+    build_target: "build"
```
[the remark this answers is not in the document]
lc525: yes, but would fail if the `seldon_mlserver_src` variable is not set. I'll look into handling that gracefully

PR #5321, opened 2024-02-15
bump(librdkafka): from v1.9.2 to v2.3.0

lc525:
Update needed for automatic re-authentication of kafka connections via SASL (used when connecting to Confluent Cloud via OAUTHBEARER)

- smoke tested using pipeline-tests.sh in local kind cluster (with local Kafka)


sakoush (approved):
lgtm. assuming jvm side is a follow up PR?

lc525:
indeed, jvm side is a follow-up. Converting this to a draft as I have seen model/pipeline gateway start issuing some certificate errors. I need to evaluate the impact of those.

PR #5329, opened 2024-02-19
fix(agent): set context deadline for grpc model server control plane

sakoush:

Previously we have a `Background` context used in the control plane operations (load/unload) issued from the agent to the model servers (mlserver and triton). The background context will wait indefinitely if the model servers do not respond back and we have seen cases (on unload) that mlserver did not respond back to the unload (not sure why). 

Anyway in the case that mlserver did not respond back the agent will be stuck indefinitely on this particular model and therefore the scheduler state of this model is also stuck.

To get around that we use a `Deadline` context now on these control plane operations as follows:

- `GRPCModelServerLoadTimeout`   = 30  Minutes 
- `GRPCModelServerUnloadTimeout` = 2 Minutes
- `GRPCControlPlaneTimeout` = 1 Minutes (for other control plane ops)

We currently use the hardocded defaults and perhaps in a future PR we can make these configurable (per model?).

Fixes INFRA-751 (internal)

- Introduced `GRPCModelServerLoadTimeout`, `GRPCModelServerUnloadTimeout` and `GRPCControlPlaneTimeout`
- Refactor code to define a `V2Config` where these parameters and others can be set and used (so that in the future can be configured)
- Added unit test

### DIFF
--- a/scheduler/pkg/agent/client_test.go
+++ b/scheduler/pkg/agent/client_test.go
@@ -797,7 +797,8 @@ func TestAgentStopOnSubServicesFailure(t *testing.T) {
 
 			time.Sleep(50 * time.Millisecond)
 
-			v2Client := oip.NewV2Client("", backEndGRPCPort, log.New())
+			v2Client := oip.NewV2Client(
+				oip.GetV2ConfigWithDefaults("", backEndGRPCPort), log.New())
 
 			modelRepository := FakeModelRepository{}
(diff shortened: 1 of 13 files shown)


lc525 (approved):
lgtm; can take another look if needed when you add the timeouts for the other ops

sakoush:
> lgtm; can take another look if needed when you add the timeouts for the other ops

now added

PR #5348, opened 2024-02-19
fix(adserver): adserver returns cloudevents compatible response

michaelcheah:

The `adserver` is meant to be deployed as a knative serving service. However, it is currently not setting the appropriate headers in its http response in the POST handler for the response to be considered a CloudEvent message. As can be seen by the [spec](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#required-attributes), the following attributes must be present in the http message:
* id
* source
* specversion
* type

Previously, none of these attributes were set in the response (though they were set when the message was forwarded to the `replyUrl`)

This PR does the following:
1. Fixes some type checks
2. Adds extra assertions in the `test_server.py` file to check that responses are CE compatible
3. Refactors the server's POST handler to return CE messages

### DIFF
--- a/components/alibi-detect-server/adserver/cm_model.py
+++ b/components/alibi-detect-server/adserver/cm_model.py
@@ -21,7 +21,7 @@ SELDON_MODEL_ID = DEFAULT_LABELS["model_name"]
 SELDON_PREDICTOR_ID = DEFAULT_LABELS["predictor_name"]
 
 
-def _load_class_module(module_path: str) -> str:
+def _load_class_module(module_path: str):
     components = module_path.split(".")
     mod = __import__(".".join(components[:-1]))
     for comp in components[1:]:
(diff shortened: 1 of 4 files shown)


components/alibi-detect-server/adserver/cm_model.py
```diff
@@ -21,7 +21,7 @@
 SELDON_PREDICTOR_ID = DEFAULT_LABELS["predictor_name"]
 
 
-def _load_class_module(module_path: str) -> str:
+def _load_class_module(module_path: str):
```
RafalSkolasinski, line 24: Why? It still seems to return `str`, or not?

RafalSkolasinski:
@michaelcheah please mark as `Ready for review` - I will hold my full review until then

components/alibi-detect-server/adserver/cm_model.py
michaelcheah: It actually returns a `Module`. It was raising type hint warnings because it's used like so in L77-79

```python
            # Load from locally available models
            MetricsClass = _load_class_module(self.storage_uri)
            self.model = MetricsClass()
```

RafalSkolasinski: so potentially `-> "Module"` (or other appropriate way to annotate the return type) could be good?

RafalSkolasinski: probably you'd need to import appropriate type

RafalSkolasinski (approved):
LGTM :rocket:

PR #5350, opened 2024-02-20
fix(scheduler): Send server statuses on controller reconnect

sakoush:

Similar to models, pipelines and experiments, we want to send the server statuses to the controller on connection to the scheduler server updates stream.

Fixes INFRA-794 (internal)


lc525 (approved):
lgtm

PR #5353, opened 2024-02-22
fix(dataflow): set default OTEL_EXPORTER_OTLP_PROTOCOL in compose setup

sakoush:

We need to provide a value to dataflow-engine for OTEL_EXPORTER_OTLP_PROTOCOL. This was missing from the docker-compose setup.

Fixes #


lc525 (approved):
lgtm; I've missed this in my otel PR, thanks for catching it up!

PR #5355, opened 2024-02-22
fix(scheduler): Update Inactive check to include LoadFailed

sakoush:

The check for inactive model replicas is missing `LoadFailed` which caused some issues with rolling upgrades to a new version.

Fixes INFRA-806 (internal)


lc525 (approved):
lgtm

PR #5358, opened 2024-02-23
fix(dataflow): handle pipeline errors and clear kafka streams state

lc525:
Because we're now continuously retrying the subscription to the scheduler when the subscription terminates, we need to also clean the kafka streams state of failed pipelines when the subscription is terminated with an error.

This commit also introduces a new way of handling errors and the changes in pipeline status. It moves us towards being able to inspect the status of the underlying kafka streams from outside the Pipeline object. The end-goal (not archieved here) is to react to pipelines reaching error states during their operation, rather than just when they are created for the first time (what's currently implemented). However, this PR does get us to a point where the pipeline creation status, including possible errors/exceptions are reported to the scheduler and can show up either in the seldon CLI or in k8s.

Instead of propagating exceptions, the idea is to return and handle common errors in a more similar way to golang.

- Fixes #INFRA-755 (internal): Exceptions in pipeline creation/deletion are unrecoverable because of uncleaned Kafka Streams state
- Progress on #INFRA-617 (internal): Investigate high priority unhandled errors
- Progress on #INFRA-648 (internal): Handle dataflow [...] exceptions and present meaningful error messages

** PR Merge Sequencing GROUP 663 **
This is part of a sequence of PRs building on each-other, but split in order to simplify reviewing. 
This PR has sequence no: G663/1

Next to review in sequence: https://github.com/lc525/seldon-core/pull/47 : fix(dataflow): wait for kafka topic creation

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
@@ -11,7 +11,7 @@ package io.seldon.dataflow
 
 import io.klogging.Level
 import io.klogging.config.loggingConfiguration
-import io.klogging.rendering.RENDER_ANSI
+import io.klogging.rendering.RENDER_ISO8601
 import io.klogging.sending.STDOUT
 
 object Logging {
@@ -20,7 +20,7 @@ object Logging {
     fun configure(appLevel: Level = Level.INFO, kafkaLevel: Level = Level.WARN) =
         loggingConfiguration {
             kloggingMinLogLevel(appLevel)
-            sink(stdoutSink, RENDER_ANSI, STDOUT)
+            sink(stdoutSink, RENDER_ISO8601, STDOUT)
             logging {
                 fromLoggerBase("io.seldon")
                 toSink(stdoutSink)
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
@@ -9,7 +9,7 @@ the Change License after the Change Date as each is defined in accordance with t
 
 package io.seldon.dataflow.kafka
 
-import io.seldon.dataflow.kafka.security.KafkaSaslMechanisms
+import io.klogging.noCoLogger
 import io.seldon.dataflow.kafka.security.SaslConfig
 import io.seldon.dataflow.mtls.CertificateConfig
 import io.seldon.dataflow.mtls.K8sCertSecretsProvider
@@ -23,7 +23,12 @@ import org.apache.kafka.common.config.SslConfigs
 import org.apache.kafka.common.config.TopicConfig
 import org.apache.kafka.common.security.auth.SecurityProtocol
 import org.apache.kafka.streams.StreamsConfig
+import org.apache.kafka.streams.errors.DeserializationExceptionHandler
+import org.apache.kafka.streams.errors.ProductionExceptionHandler
+import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler
 import java.util.*
+
+const val KAFKA_UNCAUGHT_EXCEPTION_HANDLER_CLASS_CONFIG = "default.processing.exception.handler"
 
 data class KafkaStreamsParams(
     val bootstrapServers: String,
@@ -49,6 +54,8 @@ val kafkaTopicConfig = { maxMessageSizeBytes: Int ->
         TopicConfig.MAX_MESSAGE_BYTES_CONFIG to maxMessageSizeBytes.toString(),
     )
 }
+
+private val logger = noCoLogger(Pipeline::class)
 
 fun getKafkaAdminProperties(params: KafkaStreamsParams): KafkaAdminProperties {
     return getSecurityProperties(params).apply {
@@ -156,10 +163,11 @@ fun getKafkaProperties(params: KafkaStreamsParams): KafkaProperties {
         this[StreamsConfig.NUM_STREAM_THREADS_CONFIG] = 1
         this[StreamsConfig.SEND_BUFFER_CONFIG] = params.maxMessageSizeBytes
         this[StreamsConfig.RECEIVE_BUFFER_CONFIG] = params.maxMessageSizeBytes
+        // tell Kafka Streams to optimize the topology
+        this[StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG] = StreamsConfig.OPTIMIZE
 
         // Testing
         this[StreamsConfig.REPLICATION_FACTOR_CONFIG] = params.replicationFactor
-        this[StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG] = 0
         this[StreamsConfig.COMMIT_INTERVAL_MS_CONFIG] = 10_000
 
         this[ConsumerConfig.AUTO_OFFSET_RESET_CONFIG] = "latest"
@@ -201,7 +209,20 @@ fun KafkaProperties.withStreamThreads(n: Int): KafkaProperties {
     val properties = KafkaProperties()
 
     properties.putAll(this.toMap())
-    this[StreamsConfig.NUM_STREAM_THREADS_CONFIG] = n
+    properties[StreamsConfig.NUM_STREAM_THREADS_CONFIG] = n
+
+    return properties
+}
+
+fun KafkaProperties.withErrorHandlers(deserializationExceptionHdl: DeserializationExceptionHandler?,
+                                      streamExceptionHdl: StreamsUncaughtExceptionHandler?,
+                                      productionExceptionHdl: ProductionExceptionHandler?): KafkaProperties {
+    val properties = KafkaProperties()
+    properties.putAll(this.toMap())
+
+    deserializationExceptionHdl?.let  { properties[StreamsConfig.DEFAULT_DESERIALIZATION_EXCEPTION_HANDLER_CLASS_CONFIG] = it::class.java }
+    streamExceptionHdl?.let           { properties[KAFKA_UNCAUGHT_EXCEPTION_HANDLER_CLASS_CONFIG] = it::class.java                        }
+    productionExceptionHdl?.let       { properties[StreamsConfig.DEFAULT_PRODUCTION_EXCEPTION_HANDLER_CLASS_CONFIG] = it::class.java      }
 
     return properties
 }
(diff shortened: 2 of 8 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
```diff
@@ -51,31 +89,38 @@ class Pipeline(
...
             return
         }
         // CREATED, REBALANCING and RUNNING (with the latter one already handled above)
         // are the only non-error states. Everything else indicates an error or shutdown
         // and we should release the lock on which start() awaits and return an error.
         // see: https://kafka.apache.org/28/javadoc/org/apache/kafka/streams/KafkaStreams.State.html
-        if (newState != KafkaStreams.State.CREATED && newState != KafkaStreams.State.REBALANCING) {
-            errorOrShutdown = true
+        if (newState != State.CREATED && newState != State.REBALANCING) {
+            status = PipelineStatus.Error(newState).withMessage("pipeline data streams error: kafka streams state: $newState")
```
lc525, line 123: **Review note:**

At the moment, the status updates happening in the kafka streams state update callback are not used anywhere. However, this lays the groundwork for being able to access them via the Pipeline.state property.

We'll need to be careful in selecting which of those changes should be propagated to the scheduler, in order to let it take global decisions about the fate of the pipeline (i.e re-init, stop across other components like {pipeline, model}gateway, etc)

lc525:
Moved to draft as I've realised a couple more improvements can/should be made before merging

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
```diff
@@ -11,7 +11,7 @@ package io.seldon.dataflow
 
 import io.klogging.Level
 import io.klogging.config.loggingConfiguration
-import io.klogging.rendering.RENDER_ANSI
+import io.klogging.rendering.RENDER_ISO8601
```
lc525, line 14: **Review note**
Moving to non-color logger as it produces noisy escape sequences when viewing the logs outside k9s

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
```diff
@@ -156,10 +163,11 @@ fun getKafkaProperties(params: KafkaStreamsParams): KafkaProperties {
         this[StreamsConfig.NUM_STREAM_THREADS_CONFIG] = 1
         this[StreamsConfig.SEND_BUFFER_CONFIG] = params.maxMessageSizeBytes
         this[StreamsConfig.RECEIVE_BUFFER_CONFIG] = params.maxMessageSizeBytes
+        // tell Kafka Streams to optimize the topology
+        this[StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG] = StreamsConfig.OPTIMIZE
 
         // Testing
         this[StreamsConfig.REPLICATION_FACTOR_CONFIG] = params.replicationFactor
-        this[StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG] = 0
```
lc525, line 162: Setting deprecated

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
```diff
@@ -201,7 +209,20 @@ fun KafkaProperties.withStreamThreads(n: Int): KafkaProperties {
     val properties = KafkaProperties()
 
     properties.putAll(this.toMap())
-    this[StreamsConfig.NUM_STREAM_THREADS_CONFIG] = n
```
lc525, line 204: previously, this had no effect (setting the config on "this" does not set it on the returned copy -- the `properties` variable)

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
```diff
@@ -0,0 +1,93 @@
...
+    class Started() : PipelineStatus(null, false) {
+        override var message: String? = "pipeline data streams: ready"
+    }
+
+    data class Error(val errorState: KafkaStreams.State?): PipelineStatus(errorState,true)
+
+    override var exception: Exception? = null
+    override var message: String? = null
+
+}
```
sakoush, line 93: add newline

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
```diff
@@ -0,0 +1,93 @@
+package io.seldon.dataflow.kafka
+
+import io.klogging.Klogger
+import io.klogging.Level
+import io.klogging.NoCoLogger
+import io.seldon.dataflow.DataflowStatus
+import kotlinx.coroutines.runBlocking
+import org.apache.kafka.streams.KafkaStreams
+
+open class PipelineStatus(val state: KafkaStreams.State?, var isError: Boolean) : DataflowStatus {
```
sakoush, line 10: question: where `state` be set outside of errors?

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
```diff
@@ -93,12 +137,33 @@ class Pipeline(
...
-            return Pipeline(metadata, topology, streamsApp, kafkaDomainParams, numSteps)
+            var streamsApp : KafkaStreams? = null
+            var pipelineError: PipelineStatus.Error? = null
+            try {
+                streamsApp = KafkaStreams(topology, pipelineProperties)
+            } catch (e: StreamsException) {
+                pipelineError = PipelineStatus.Error(null)
+                    .withException(e)
+                    .withMessage("failed to initialize kafka streams app")
+                return null to pipelineError
```
sakoush, line 151: question: lacking kotlin knowledge but assuming this returns `(null, pipelineError)`

sakoush (approved):
Great change! I have added comments / questions to the extent I can read kotlin.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Configuration.kt
```diff
@@ -156,10 +163,11 @@ fun getKafkaProperties(params: KafkaStreamsParams): KafkaProperties {
         this[StreamsConfig.NUM_STREAM_THREADS_CONFIG] = 1
         this[StreamsConfig.SEND_BUFFER_CONFIG] = params.maxMessageSizeBytes
         this[StreamsConfig.RECEIVE_BUFFER_CONFIG] = params.maxMessageSizeBytes
+        // tell Kafka Streams to optimize the topology
+        this[StreamsConfig.TOPOLOGY_OPTIMIZATION_CONFIG] = StreamsConfig.OPTIMIZE
```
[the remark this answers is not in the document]
lc525: I think setting OPTIMIZE to on is safe, and shouldn't significantly impact startup times. One of the areas where it can help is in reducing the need to rebalance consumers. At the moment the level of optimisations added is really small, and would not affect a wide array of cases. 

Only two cases are covered: 
1. reusing the source topic as the kafka-streams changelog topic (rather than creating a new one) and 
2. colapsing multiple repartition topics into a single repartition topic (this happens when changing the key that gets processed within a topology from one step to the next). 

Essentially some repartitions are lazy, waiting for a whole pass through the topology before doing one rebalance rather than eagerly doing rebalancing during each operation.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
lc525: Indeed, that's the relatively strange but terse way of building a pair. It gets slightly less strange if you think of the pair as mapping one element **to** another one.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
```diff
@@ -93,12 +137,33 @@ class Pipeline(
...
+            }
+
+            val uncaughtExceptionHandlerClass = pipelineProperties[KAFKA_UNCAUGHT_EXCEPTION_HANDLER_CLASS_CONFIG] as? Class<StreamsUncaughtExceptionHandler>?
+            uncaughtExceptionHandlerClass?.let{
+                logger.info("Setting custom Kafka streams uncaught exception handler")
```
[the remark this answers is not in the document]
lc525: I can move it to `debug`, it makes sense.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
```diff
@@ -93,12 +137,33 @@ class Pipeline(
...
+                streamsApp.setUncaughtExceptionHandler(it.getDeclaredConstructor().newInstance())
```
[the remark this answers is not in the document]
lc525: It's a rather strange API:

Two of the exception handlers can be set as configuration values (`DeserializationExceptionHandler` and `ProductionExceptionHandler`)

The third one (the `StreamsUncaughtExceptionHandler`, which your comment refers to) can only be set via the direct call to `streamsApp.setUncaughtExceptionHandler`.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
```diff
@@ -0,0 +1,93 @@
...
+            if (prev is StreamStopped) {
+                this.prevState = prev.prevState
+            }
+            this.isError = this.prevState?.isError ?: false
+        }
+
+        override fun getDescription() : String? {
+            val exceptionMsg = this.exception?.message
+            var statusMsg = this.message
+            val prevStateDescription = prevState?.getDescription()
```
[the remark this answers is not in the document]
lc525: they are equivalent, I'll change to `this.prevState` for consistency and readability

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
lc525: For example `StreamStarted` and `StreamStopped` are not errors by default (the Pipeline transitions to `StreamStopped` after the stop() call, which would also happen when the pipeline gets deleted, for example).

PR #5359, opened 2024-02-26
fix(gateway): wait for kafka topic creation

lc525:

Kafka topic creation happens asynchronously. This means that even when the
return value from createTopics(...) indicates that the topic has been created
successfuly, the topic can not be immediately subscribed to.

We retry DescribeTopics until all of the topics for the pipeline can be
described successfully. This indicates that the topic has been fully created at
least on one broker, and can now be subscribed to.

Fixes gateway component for #INFRA-663 (internal): Pipeline creation goes into ERROR state

### DIFF
--- a/scheduler/pkg/kafka/gateway/infer.go
+++ b/scheduler/pkg/kafka/gateway/infer.go
@@ -11,7 +11,6 @@ package gateway
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -265,7 +264,7 @@ func (kc *InferKafkaHandler) ensureTopicsExist(topicNames []string) error {
 
 	for _, topicDescription := range topicsDescResult.TopicDescriptions {
 		if topicDescription.Error.Code() != kafka.ErrNoError {
-			return fmt.Errorf("topic description failure: %s", topicDescription.Error.Error)
+			return fmt.Errorf("topic description failure: %s", topicDescription.Error.Error())
 		}
 	}
 
--- a/scheduler/pkg/kafka/gateway/infer.go
+++ b/scheduler/pkg/kafka/gateway/infer.go
@@ -197,7 +197,7 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
 	logger := kc.logger.WithField("func", "createTopics")
 	if kc.adminClient == nil {
 		logger.Warnf("no kafka admin client, can't create any of the following topics: %v", topicNames)
-		return nil
+		return fmt.Errorf("no kafka admin client")
 	}
 	t1 := time.Now()
 
@@ -212,7 +212,7 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
 	results, err := kc.adminClient.CreateTopics(
 		context.Background(),
 		topicSpecs,
-		kafka.SetAdminOperationTimeout(time.Minute),
+		kafka.SetAdminOperationTimeout(TopicCreateTimeout),
 	)
 	if err != nil {
 		return err
@@ -220,11 +220,14 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
 
 	// Wait for topic creation
 	logFailure := func(err error, delay time.Duration) {
-		logger.WithError(err).Errorf("waiting for topic creation")
+		logger.WithError(err).Errorf("still waiting for all topics to be created...")
 	}
 
-	logger.Infof("waiting for topic creation...")
-	retryPolicy := backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 60)
+	logger.Infof("waiting for kafka topic creation")
+	retryPolicy := backoff.WithMaxRetries(
+		backoff.NewConstantBackOff(TopicDescribeRetryDelay),
+		TopicDescribeMaxRetries,
+	)
 	err = backoff.RetryNotify(
 		func() error {
 			return kc.ensureTopicsExist(topicNames)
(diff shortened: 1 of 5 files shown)


scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -216,16 +218,57 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
...
+	// Wait for topic creation
+	logFailure := func(err error, delay time.Duration) {
+		logger.WithError(err).Errorf("waiting for topic creation")
+	}
+
+	logger.Infof("waiting for topic creation...")
+	retryPolicy := backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 60)
+	err = backoff.RetryNotify(
+		func() error {
+			return kc.ensureTopicsExist(topicNames)
```
sakoush, line 230: do you want to add logging?

scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -194,7 +196,7 @@ func (kc *InferKafkaHandler) GetNumModels() int {
 func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
 	logger := kc.logger.WithField("func", "createTopics")
 	if kc.adminClient == nil {
-		logger.Warnf("Can't create topics %v as no admin client", topicNames)
+		logger.Warnf("no kafka admin client, can't create any of the following topics: %v", topicNames)
```
sakoush, line 199: return error?

scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -216,16 +218,57 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
...
+
+	if err != nil {
+		logger.WithError(err).Errorf("some topics not created, giving up")
+	} else {
+		logger.Infof("all topics created")
+	}
+
 	for _, result := range results {
-		logger.Debugf("Topic result for %s", result.String())
+		logger.Debugf("topic result for %s", result.String())
```
sakoush, line 242: return error?

scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -216,16 +218,57 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
...
 	}
 
 	t2 := time.Now()
-	logger.Infof("Topic created in %d millis", t2.Sub(t1).Milliseconds())
+	logger.Infof("kafka topics created in %d millis", t2.Sub(t1).Milliseconds())
```
sakoush, line 246: move to debug?

scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -216,16 +218,57 @@ func (kc *InferKafkaHandler) createTopics(topicNames []string) error {
...
 
 	return nil
 }
 
+func (kc *InferKafkaHandler) ensureTopicsExist(topicNames []string) error {
+		ctx, cancel := context.WithTimeout(context.Background(), 2 * time.Second)
```
sakoush: extract 2s as constant? is it 1s on the dataflow PR?

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -167,12 +167,12 @@ func (km *KafkaManager) loadOrStorePipeline(resourceName string, isModel bool) (
 		if err != nil {
 			return nil, err
 		}
-		pipeline.wg.Add(1) // wait set to allow consumer to say when started
 
 		val, loaded := km.pipelines.LoadOrStore(key, pipeline)
 		if loaded { // we can still have a race condition where multiple "create" are happening, we have to store the first one
 			pipeline = val.(*Pipeline)
 		} else {
+			pipeline.wg.Add(1) // wait set to allow consumer to say when started
```
sakoush, line 175: I think that there might be a race condition if put it here:
if we have 2 infer requests at the same time for a new pipeline: r1 and r2.
r1 is going to create the pipeline and r1 will get the reference at line 171. r2 is faster and will proceed until `pipeline.wg.Wait()` without waiting for r1 to start the consume at line 175.?

sakoush (commented):
I am assuming that we do not need to do anything for pipelinegateway (waiting for topic creation - pipeline input and pipeline output) as this is is done by dataflow engine and it will wait for the these topics to be created (with the other PR).

scheduler/pkg/kafka/gateway/infer.go
lc525: logging exists, the passed `logFailure` function is called after each failed attempt.

scheduler/pkg/kafka/gateway/infer.go
lc525: error now returned, line 240

scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -241,7 +287,7 @@ func (kc *InferKafkaHandler) AddModel(modelName string) error {
 	kc.subscribedTopics[inputTopic] = true
 	err := kc.subscribeTopics()
 	if err != nil {
-		kc.logger.WithError(err).Warn("Failed to subscribe to topics")
+		kc.logger.WithError(err).Errorf("failed to subscribe to topics")
 		return nil
```
lc525, line 291: we should investigate more on why all of these errors are being swallowed in `infer.go` instead of being returned

sakoush (approved):
Lgtm

scheduler/pkg/kafka/gateway/infer.go
lc525: It appears that returning an error here fails the go tests

PR #5375, opened 2024-02-26
fix(dataflow): wait for kafka topic creation

lc525:
This is identical to the reviewed (and approved) https://github.com/lc525/seldon-core/pull/47 but rebased against current seldonio/v2 for merging

**What this PR does / why we need it:**
Kafka topic creation happens asynchronously. This means that even when the return value from `createTopics(...)` indicates that the topic has been created successfuly, the topic can not be immediately subscribed to.

Instead of verifying the status of the topic from the `createTopics` return value, here we're repeatedly calling `describeTopics` until all of the topics for the pipeline can be described successfully. This indicates that the topic has been fully created _at least_ on one broker, and can now be subscribed to.

**Which issue(s) this PR fixes:**
Fixes dataflow component for #INFRA-663 (internal): Pipeline creation goes into ERROR state

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
@@ -47,6 +47,12 @@ object Cli {
     val brokerSecret = Key("kafka.tls.broker.secret", stringType)
     val endpointIdentificationAlgorithm = Key("kafka.tls.endpoint.identification.algorithm", stringType)
 
+    // Kafka waiting for topic creation
+    val topicCreateTimeoutMillis = Key("topic.create.timeout.millis", intType)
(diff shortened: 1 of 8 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
```diff
@@ -75,6 +81,10 @@ object Cli {
             clientSecret,
             brokerSecret,
             endpointIdentificationAlgorithm,
+            topicCreateTimeoutMillis,
+            topicDescribeTimeoutMillis,
+            topicDescribeRetries,
+            topicDescribeRetryDelayMillis,
```
sakoush, line 87: Lgtm

sakoush (approved):
Lgtm

PR #5381, opened 2024-02-28
fix(dataflow): be more strict about error states on pipeline creation

lc525:
Only return success = true if the pipeline is running

A concurrency issue remains because if running the code on multiple coroutines these may interleave at suspension points. This is not an issue atm, as we are executing on one coroutine. If/when we introduce more parallelism, this will need to be dealt with.

**Fixed issues**:
- #INFRA-830 (internal): pipeline might be marked ready before is actually running

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
@@ -186,12 +186,11 @@ class PipelineSubscriber(
             }
         }
 
-        // There is a small chance that pipeline.start() returned a status of PipelineState.StreamStopped(),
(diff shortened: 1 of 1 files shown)


sakoush (approved):
lgtm. although we probably need to test if the state on the scheduler side can be moved to ready (once it has started) beyond this point.

For now as we are effectively running serial operations this case cannot happen.

seldondev:
* failed to create agent
* failed to load trigger config for repository SeldonIO/seldon-core for ref 
* failed to switch to ref 
* failed to merge repository SeldonIO/seldon-core
* failed to run command in dir /var/tmp/gitrepo3773906085
* git, [merge FETCH_HEAD]
* Updating cfe54e21..6a45b2b0
error
* Your local changes to the following files would be overwritten by merge:
	.github/workflows/images.yml
Please commit your changes or stash them before you merge.
Aborting
* exit status 1

PR #5403, opened 2024-03-05
fix: Use mlserver 1.5.0

sakoush:

This change upgrades the "default" images for mlserver to point to `1.5.0` as it is now released.
Note that some models are still using the artifacts generated from mlserver `1.3.5` but they work with the new version.
I didnt regenerate the notebooks for now but happy to do if required.

Fixes #

Model artifacts are generated and uploaded to [seldon public cloud storage](https://console.cloud.google.com/storage/browser/seldon-models/scv2/samples/mlserver_1.5.0?project=seldon-pub)

using `~/samples$ make train-all upload-all`

### DIFF
--- a/samples/scripts/models/Makefile
+++ b/samples/scripts/models/Makefile
@@ -4,11 +4,12 @@ MLSERVER_FOLDER = mlserver_${MLSERVER_VERSION}
 TRITON_VERSION=$(shell grep 'nvidia/tritonserver' ../../../scheduler/Makefile | cut -d':' -f2 | cut -d'-' -f1 | tr '.' '-')
 TRITON_FOLDER=triton_${TRITON_VERSION}
 
+# note that income-lgb fails with new version of lightgbm
 .PHONY: train-all
-train-all: iris huggingface-text-gen-gpt2 huggingface-text-gen-tiny-stories moviesentiment income income-xgb income-lgb download-mnist-onnx download-cifar10-tensorflow wine-mlflow mnist-pytorch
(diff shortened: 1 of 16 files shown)


samples/scripts/models/Makefile
```diff
@@ -4,11 +4,12 @@ MLSERVER_FOLDER = mlserver_${MLSERVER_VERSION}
 TRITON_VERSION=$(shell grep 'nvidia/tritonserver' ../../../scheduler/Makefile | cut -d':' -f2 | cut -d'-' -f1 | tr '.' '-')
 TRITON_FOLDER=triton_${TRITON_VERSION}
 
+# note that income-lgb fails with new version of lightgbm
 .PHONY: train-all
-train-all: iris huggingface-text-gen-gpt2 huggingface-text-gen-tiny-stories moviesentiment income income-xgb income-lgb download-mnist-onnx download-cifar10-tensorflow wine-mlflow mnist-pytorch
```
lc525, line 8: Is income-lgb no longer needed? (I see its corresponding make rule has remained in the Makefile)
lgtm otherwise.

lc525 (approved):
a minor question related to income-lgb, mostly for context

samples/scripts/models/Makefile
sakoush: It is still part of the examples we provide, however building a new model with the new lgb library fails because of some api changes in the new library. Given that the old version still works then I thought we can fix this issue later in a follow up PR.

PR #5405, opened 2024-03-07
fix(dataflow): catch all exceptions when creating a KafkaStreams object

lc525:
Previously, we only caught a StreamsException. However, the creation might fail for many reasons (for example, incorrect configuration).

We want to catch any exception so that we mark the pipeline creation as failed and we don't stop the connection to the scheduler.

Previously, on configuration errors, the exception would be bubbled to the PipelineSubscriber event loop, and the connection to the scheduler would be broken. We would try to reconnect, but on reconnect the scheduler would try to re-init the problematic pipeline (with the same id). This then led to an error about existing uncleaned KafkaStreams state in /tmp. This latter error was being handled cleanly (i.e not breaking the connection to the scheduler anymore), but would mask the real reason for the failure when looking at the pipeline status (via k8s or seldon cli).


sakoush (approved):
lgtm

PR #5406, opened 2024-03-07
fix(scheduler): Increase event hub buffer inline with other components

sakoush:

This change increases the pending queues in the event hub (coordinator) for server and model events delivery. There is going to be a tradeoff, with a bigger buffer we allow more work to pile up if the producer is faster than the consumer. However in practice setting it to `1000` has worked fined in other cases, check [here](https://github.com/SeldonIO/seldon-core/blob/f8df070fb201aee625cb1abf764f4122ebb84b05/scheduler/pkg/envoy/processor/incremental.go#L37).

It also reduces the risk of deadlocks as producers will block otherwise when trying to publish on a queue that it full.

Fixes #

- changing all pending queues sizes to 1000
- fix a missing unlock call from reviewing code around this area

### DIFF
--- a/scheduler/pkg/envoy/processor/incremental.go
+++ b/scheduler/pkg/envoy/processor/incremental.go
@@ -588,6 +588,7 @@ func (p *IncrementalProcessor) modelUpdate(modelName string) error {
 			logger.Debugf("sync: No server - removing for %s", modelName)
 			if err := p.removeRouteForServerInEnvoyCache(modelName); err != nil {
 				logger.WithError(err).Errorf("Failed to remove model route from envoy %s", modelName)
+				p.modelStore.UnlockModel(modelName)
 				return err
 			}
 			modelRemoved = true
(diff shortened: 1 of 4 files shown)


scheduler/pkg/envoy/processor/incremental.go
```diff
@@ -588,6 +588,7 @@ func (p *IncrementalProcessor) modelUpdate(modelName string) error {
 			logger.Debugf("sync: No server - removing for %s", modelName)
 			if err := p.removeRouteForServerInEnvoyCache(modelName); err != nil {
 				logger.WithError(err).Errorf("Failed to remove model route from envoy %s", modelName)
+				p.modelStore.UnlockModel(modelName)
```
lc525, line 591: nice catch!

lc525 (approved):
lgtm

scheduler/pkg/envoy/processor/incremental.go
lc525: we might be able to convert this into a defer, as we always unlock before return

sakoush: there is a for loop that goes over multiple models and therefore we want to lock/unlock per iteration. maybe we should refactor this logic per iteration in a separate function so that we can use defer.

sakoush: It looks like we need to release the model lock before running further logic at the end of this method

Issue #5032, opened 2023-07-18
Model don'nt get redeployed after server and scheduler are killed together


SDJustus:
## Describe the bug
As of Release version 2.6.0, there was a change, where the seldon-controller can be installed clusterwide. With this new feature, there now is a bug (with the seldon-controller still installed namespaced), that when the scheduler and a server are killed together, the models, that have been on the killed server, will not load again, since, in difference to the releases before, the controller doesn't restart itself when the scheduler gets killed. Thus, no redeployment of the models happens

## To reproduce
1. install release 2.6.0 with controller not clusterwide installed (still namespaced): following the helm install [doc](https://docs.seldon.io/projects/seldon-core/en/v2/contents/getting-started/kubernetes-installation/helm.html)
2. deploy a server
3. deploy models on the server
4. kill the scheduler and the server together
-> No models redeployed and if you run kubectl describe on the models, there are still marked as being ready

## Expected behaviour
For all models to be redeployed.

## Environment

* Cloud Provider: AWS EKS
* Kubernetes Cluster Version: 1.24
* Deployed Seldon System Images: Helm installation of release 2.6.0

## Model Details 
Not relevant.

Kolajik:
We have this issue as well. The model is not loaded again when a Server is restarted for any reason and the Model object still has status Ready.

It does help to remove the Model (after you remove finalizers manually per bug #5043), restart the scheduler statefulset, restart the Server once more and then redeploy the Model object again.

I believe there is something wrong in the scheduler. Maybe it keeps an incorrect state of the Model or does not reflect the Server restart at all?

SDJustus:
I think this particular issue came with the controller being able to be installed clusterwide... Before that, the Controller got restarted everytime something happened with the scheduler.
I don't know, if this is already fixed in some of the pre Releases though..

Kolajik:
I am using the `latest` images of scheduler, controller and envoy. And that version should be higher than `2.6.0`, so I don't think it was fixed.

SDJustus:
Can you point out if you are aware of this and can perhaps reproduce it? This seems like a critical scenario. Maybe @sakoush?

sakoush:
A fix is now merged https://github.com/SeldonIO/seldon-core/pull/5411 @SDJustus

PR #5411, opened 2024-03-11
fix(controller): reload models upon reconnect to the scheduler

sakoush:

In Core 2, the scheduler state of the models are not persisted to local storage and the system relies on the model servers to keep this state. In the case the scheduler restarts, the models servers will reconnect and then announce to the scheduler the models that they have loaded, which allows the scheduler to recover this state.

However this suffers from the issue that if the model server AND the scheduler restart, then the models states that were handled by this model server is effectively lost.

In this case there is a mismatch between between the state of the scheduler (models from the restarted model server are gone) and the controller (models from the restarted model server are ready).

We rely on k8s etcd state to recover this state loss on the scheduler side and in this PR we reload models that are marked in k8s on the controller reconnecting to the scheduler.

Note that we need to holistically think about this issue in the future but for now we decided we recover the state from k8s.

Fixes:
- #5032
- internal: INFRA_620

### DIFF
--- a/operator/scheduler/model.go
+++ b/operator/scheduler/model.go
@@ -31,6 +31,10 @@ import (
 	"github.com/seldonio/seldon-core/operator/v2/pkg/utils"
 )
 
+// LoadModel loads a model to the scheduler
+// If the connection is not provided, get a new one
+// In the case of errors we check if the error is retryable and return a boolean to indicate if the error is retryable
+// For the cases we think we should retry, check logic in `checkErrorRetryable`
 func (s *SchedulerClient) LoadModel(ctx context.Context, model *v1alpha1.Model, conn *grpc.ClientConn) (bool, error) {
 	logger := s.logger.WithName("LoadModel")
 	retryableError := false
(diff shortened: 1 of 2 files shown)


operator/scheduler/model.go
```diff
@@ -300,7 +315,39 @@ func (s *SchedulerClient) handlePendingDeleteModels(
 				// if the model exists in the scheduler so we wait until we get the event from the subscription stream
 				s.logger.Info("Unload model called successfully, not removing finalizer", "model", model.Name)
 			}
-			break
```
sakoush, line 303: this is a bug from previous work, we should not break as we need to go over the entire list of models.

lc525 (approved):
Had a look and this looks perfectly sensible.

Let's have a discussion on how this could be re-architected moving forward. As I see it, the issue is that we deal with models, pipelines and experiments very differently, and it can become confusing. Having k8s as the source of truth makes sense as otherwise we'll probably have to run our own cluster consensus process. On the other hand, "actual" state might be different from what k8s knows about, in terms of component/model status.

We should probably at least document / be explicit about the current behaviour on how we synchronise state between k8s, controller and scheduler so that we don't get surprised / forget about it as we're building more functionality.

PR #5427, opened 2024-03-12
fix(modelgateway): kafka topics with correct number of partitions

lc525:
The number of partitions with which model-specific kafka topics is created needs to be the same as the number of partitions for other topics (pipelines/experiments/etc). This is a Kafka Streams constraint in order for joins to work correctly.

Until now, an incorrect environment variable was being passed to modelgateway, which meant that it was always creating topics with 1 partition. This in turn led to dataflow processing getting stuck whenever other topics were created with multiple partitions

**Fixed issues**:
- INFRA-822: Pipelines get stuck on triggers/joins

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-setup/values.yaml.template
+++ b/k8s/helm-charts/seldon-core-v2-setup/values.yaml.template
@@ -252,7 +252,7 @@ serverConfig:
       pullPolicy: IfNotPresent
       registry: docker.io
       repository: seldonio/mlserver
-      tag: 1.3.5
+      tag: 1.5.0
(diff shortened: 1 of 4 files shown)


k8s/helm-charts/seldon-core-v2-setup/values.yaml.template
```diff
@@ -252,7 +252,7 @@ serverConfig:
       pullPolicy: IfNotPresent
       registry: docker.io
       repository: seldonio/mlserver
-      tag: 1.3.5
+      tag: 1.5.0
```
lc525, line 255: without this change, on automated re-generation of helm-charts & yamls, 1.3.5 was being introduced instead of 1.5.0

sakoush: nice catch

sakoush (approved):
lgtm

PR #5434, opened 2024-03-14
fix(scheduler): Introduce `sendMsg` timeout for grpc streams (scheduler -> controller)

sakoush:
This change adds a timeout for sending events to the controller from the scheduler. These events are sent on server side (scheduler) grpc streams (4x for each scheduler) and we have noticed that in some cases (pending more investigations of the root cause), `sendMsg` blocks indefinitely because of the lack of control flow. We double checked the controller and it is also waiting on receive so we are for now ruling out any issues with slow consumers.

The default timeout we added is 30s for each send and in this case we break the context and therefore the controller will detect a stream disconnect, and reestablishing the connection with new a stream to unblock events.

Note that we reply events so there is no risk of losing message with this logic.

logic inspired from: https://github.com/grpc/grpc-go/issues/1229

This PR also adds:
- More unit test coverage for this part of the codebase
- `stress-tests` bash script to create multiple models of the same type in parallel (currently only 2 model types supported)
- tidy go modules for the different modules of the project
- align gprc and proto go libraries for all modules

Fixes INFRA-686 (internal)


sakoush:
looks like the fix is working as I got  on scheduler
```
time="2024-03-14T17:25:56Z" level=error msg="Failed to send model status event to seldon manager for tfsimple1:1" error="rpc error: code = DeadlineExceeded desc = Failed to send event within timeout" func=sendModelStatusEvent source=SchedulerServer
```

samples/stress-tests.sh
```diff
@@ -0,0 +1,102 @@
+# usage: ./pipeline-tests.sh [count] [kubectl|seldon] [namespace]
```
lc525, line 1: nit: stress-tests.sh in the usage

sakoush: done in 
ccbd930

scheduler/pkg/server/utils.go
```diff
@@ -0,0 +1,35 @@
...
+package server
+
+import (
+	"time"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func sentWithTimeout(f func() error, d time.Duration) (bool, error) {
```
lc525, line 19: I propose we rename this to `sendWithTimeout`

scheduler/pkg/server/experiment_status.go
```diff
@@ -10,15 +10,25 @@ the Change License after the Change Date as each is defined in accordance with t
...
 
 	"github.com/seldonio/seldon-core/scheduler/v2/pkg/coordinator"
+	"github.com/seldonio/seldon-core/scheduler/v2/pkg/store/experiment"
 )
 
 func (s *SchedulerServer) SubscribeExperimentStatus(req *pb.ExperimentSubscriptionRequest, stream pb.Scheduler_SubscribeExperimentStatusServer) error {
 	logger := s.logger.WithField("func", "SubscribeExperimentStatus")
 	logger.Infof("Received subscribe request from %s", req.GetSubscriberName())
 
+	err := s.sendCurrentExperimentStatuses(stream)
```
sakoush, line 26: lets also send experiment updates at startup similar to models and pipelines. This is something that we missed doing from previous work.

scheduler/pkg/server/utils.go
sakoush: done in [b90056a](https://github.com/SeldonIO/seldon-core/pull/5434/commits/b90056a15731f316d43d718cdaec388f1f3b34ee)

scheduler/pkg/server/experiment_status_test.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+		{
+			name: "timeout",
+			loadReq: &experiment.Experiment{
+				Name: "foo",
+				Mirror: &experiment.Mirror{
+					Name: "bar",
+				},
+			},
+			timeout: 1 * time.Millisecond,
```
lc525, line 125: will this test always time-out, or is there a risk of it being flaky?

lc525 (approved):
lgtm. Just really minor nits, I like the added tests.

PR #5469, opened 2024-03-21
fix(dataflow): Do not skip events to remove old pipeline versions

sakoush:
In https://github.com/SeldonIO/seldon-core/pull/5080 there is a change to skip events coming from `chainer-server`, this essentially has stopped the flow of removing old version of pipelines during a rolling update. The reason being is that the logic that triggers the removal of old versions of the pipeline ([here](https://github.com/SeldonIO/seldon-core/blob/7d39456b5a38e24715f0dee002dbbd9e56d8f366/scheduler/pkg/store/pipeline/store.go#L371-L373)) is eventually coming from setting the status of the new version of the pipeline from an event sent from the chainer-server.

This PR then makes the change to not set the source of the events for deleting old version and therefore they can be processed further by chainer `handlePipelineEvent`.

This seems not very intuitive but I am not sure if there is a better way at the minute without a big refactor.

### DIFF
--- a/scheduler/pkg/store/pipeline/store.go
+++ b/scheduler/pkg/store/pipeline/store.go
@@ -342,20 +342,19 @@ func (ps *PipelineStore) terminateOldUnterminatedPipelinesIfNeeded(pipeline *Pip
 func (ps *PipelineStore) SetPipelineState(name string, versionNumber uint32, uid string, status PipelineStatus, reason string, source string) error {
 	logger := ps.logger.WithField("func", "SetPipelineState")
 	logger.Debugf("Attempt to set state on pipeline %s:%d status:%s", name, versionNumber, status.String())
-	evts, err := ps.setPipelineStateImpl(name, versionNumber, uid, status, reason)
+	evts, err := ps.setPipelineStateImpl(name, versionNumber, uid, status, reason, source)
 	if err != nil {
 		return err
 	}
 	if ps.eventHub != nil {
 		for _, evt := range evts {
-			evt.Source = source
 			ps.eventHub.PublishPipelineEvent(setStatusPipelineEventSource, *evt)
 		}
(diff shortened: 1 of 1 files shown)


scheduler/pkg/store/pipeline/store.go
```diff
@@ -342,20 +342,19 @@ func (ps *PipelineStore) terminateOldUnterminatedPipelinesIfNeeded(pipeline *Pip
 func (ps *PipelineStore) SetPipelineState(name string, versionNumber uint32, uid string, status PipelineStatus, reason string, source string) error {
 	logger := ps.logger.WithField("func", "SetPipelineState")
 	logger.Debugf("Attempt to set state on pipeline %s:%d status:%s", name, versionNumber, status.String())
-	evts, err := ps.setPipelineStateImpl(name, versionNumber, uid, status, reason)
+	evts, err := ps.setPipelineStateImpl(name, versionNumber, uid, status, reason, source)
```
KengoA, line 345: I'm a bit confused, this line adds `source` which is not used in `setPipelineStateImpl`? Would passing this arg alone signal to the chainer that this needs to be processed?

lc525 (approved):
This set of changes does indeed look a bit strange, but thanks to your PR description it became easy to follow the logic. Makes perfect sense to me.

However, we have to be careful not to create any infinite loops in the future, as it seems it would be quite easy to do (now, there are some events without a source coming from "chainer-server")

LGTM

KengoA (approved):
LGTM!

For context the line from #5080 where events from chainer-server are ignored: 
https://github.com/agrski/seldon-core/blob/c09abdd3ea897b9100dd779a9f0ddf72b19a82c1/scheduler/pkg/kafka/dataflow/server.go#L383

PR #5483, opened 2024-03-27
fix(deps): bump kotlin-retry 1.0.9 to 2.0.0 in /scheduler/data-flow

lc525:
The version bump required small changes in the way we use the API.


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/KafkaAdmin.kt
```diff
@@ -105,5 +105,6 @@ class KafkaAdmin(
 
     companion object {
         private val logger = coLogger(KafkaAdmin::class)
+        private val noCoLogger = noCoLogger(KafkaAdmin::class)
```
lc525, line 108: we need this because in the new version of kotlin-retry, the retry policy is no longer executed in coroutine context.

sakoush (approved):
lgtm

PR #5504, opened 2024-04-08
fix: Handle unload too quick after load

sakoush:
In the cases where we have a model load that is followed by unload/delete too quickly, there is an edge case where the unloading logic is confused because there is no server attached yet (while the model is being loaded from the first cmd). This change then adds another check to make sure that we consider cases where the model is still loading.

fixes INFRA-943 (internal)
Notes:
- sometimes in the model server agent we get this error now 
```
level=error msg="Failed to handle load model tfsimple1:1" Name=Client error="rpc error: code = Internal desc = State mismatch for tfsimple1:1 expected state Loading but was Unloading when trying to move to state Loaded"
```
This means that the state has changed beyond the expected flow, which is probably fine in this situation.

### DIFF
--- a/samples/stress-tests.sh
+++ b/samples/stress-tests.sh
@@ -101,7 +101,7 @@ function status() {
 # The following functions are the `task` options that can be called by the user.
 
 function tfsimple() {
-      echo ${1}
+      echo $i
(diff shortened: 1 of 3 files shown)


lc525 (approved):
lgtm

PR #5527, opened 2024-04-19
fix: Deal with multi model env vars

sakoush:
In our k6 tests the use can specify `MODEL_TYPE` as a comma separated list. This change allows envars to be constructed based on this list without having to be explicitly specified by the user.

The main issue we are fixing with this PR is really `MODEL_NUM_REPLICAS` so it is defaults to 1s.

FIXES INFRA_944 (internal)


lc525 (approved):
lgtm

PR #5550, opened 2024-04-25
fix(dataflow): do not create new KafkaStreams app for existing pipelines

lc525:
This fixes a dataflow-engine bug triggered when the scheduler sends/re-sends pipeline creation messages after a restart (because it's not aware of their status across various components).

Previous behaviour:
Dataflow-engine, on receiving a command to create a pipeline, would first create a new Kafka Streams application for this pipeline, before checking if one already exists and it's running.

Because of this, triggering a control-plane restart of the scheduler would result in dataflow errors for pipelines that kept internal state (mostly pipelines making use of triggers/joins). Kafka Streams would complain about an existing application using the same state directory, fail the newly created pipeline and inform the scheduler about this.

However, in actuality the old pipeline, if it was previously running ok, would continue doing so inside dataflow. This meant that a disconnect between the state of dataflow-engine and what the scheduler knew about it was being created

New behaviour:
The introduced changes mean that dataflow-engine first checks if a pipeline with the same id is already running. If its state is ok, dataflow simply informs the scheduler that the pipeline is created, without taking further action.

If a pipeline with the same id already exists but is in a failed state, it is first stopped (local Kafka Streams state is cleaned), then an attempt is made to re-create it, with the corresponding status being sent to the scheduler.

**Which issue(s) this PR fixes:**
- INFRA-978 (internal issue) Pipelines fail to start because Kafka Streams state hasn't been cleared

- Tested by restarting scheduler while a pipeline using state (a join) was running on dataflow engine
- Ran pipeline smoke-tests to confirm the change hasn't introduced unexpected behaviour

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
@@ -141,12 +141,57 @@ class PipelineSubscriber(
         kafkaConsumerGroupIdPrefix: String,
         namespace: String,
     ) {
-        logger.info(
-            "Create pipeline {pipelineName}  version: {pipelineVersion} id: {pipelineId}",
-            metadata.name,
-            metadata.version,
-            metadata.id,
-        )
+        // If a pipeline with the same id exists, we assume it has the same name & version
+        // If it's in an error state, try re-creating.
+        //
+        // WARNING: at the moment handleCreate is called sequentially on each update in
+        // Flow<PipelineUpdateMessage> from subscribePipelines(). This allows us to sidestep issues
+        // related to race conditions on `pipelines.containsKey(...)` below. If we ever move to
+        // concurrent creation of pipelines, this needs to be revisited.
+        if (pipelines.containsKey(metadata.id)) {
+            val previous = pipelines[metadata.id]!!
+            if (!previous.status.isError) {
+                client.pipelineUpdateEvent(
+                    makePipelineUpdateEvent(
+                        metadata = metadata,
+                        operation = PipelineOperation.Create,
+                        success = true,
+                        reason = previous.status.getDescription() ?: "pipeline created",
+                    ),
+                )
+                logger.debug(
+                    "response to scheduler: pipeline {pipelineName} continues to run normally; " +
+                        "pipeline version: {pipelineVersion}, id: {pipelineId}",
+                    metadata.name,
+                    metadata.version,
+                    metadata.id,
+                )
+                return
+            } else { // pipeline exists but in failed state; cleanup state and re-create
+                logger.info(
+                    "Recreating failed pipeline {pipelineName} version: {pipelineVersion}, id: {pipelineId}",
+                    metadata.name,
+                    metadata.version,
+                    metadata.id,
+                )
+                logger.debug(
+                    "Previous state for failed pipeline {pipelineName} version: {pipelineVersion}, id: {pipelineId}: {pipelineStatus}",
+                    metadata.name,
+                    metadata.version,
+                    metadata.id,
+                    previous.status.getDescription(),
+                )
+                previous.stop()
+            }
+        } else { // pipeline doesn't exist
+            logger.info(
+                "Creating pipeline {pipelineName} version: {pipelineVersion} id: {pipelineId}",
+                metadata.name,
+                metadata.version,
+                metadata.id,
+            )
+        }
+
         val (pipeline, err) =
             Pipeline.forSteps(
                 metadata,
@@ -184,28 +229,19 @@ class PipelineSubscriber(
             return
         }
 
-        val previous = pipelines.putIfAbsent(metadata.id, pipeline)
-        var pipelineStatus: PipelineStatus
-        if (previous == null) {
-            val errTopics = kafkaAdmin.ensureTopicsExist(steps)
-            if (errTopics == null) {
-                pipelineStatus = pipeline.start()
-            } else {
-                pipelineStatus =
-                    PipelineStatus.Error(null)
-                        .withException(errTopics)
-                        .withMessage("kafka streams topic creation error")
-                pipeline.stop()
-            }
+        // This overwrites any previous pipelines with the same id. We can only get here if those previous pipelines
+        // are in a failed state and they are being re-created by the scheduler.
+        pipelines[metadata.id] = pipeline
+        val pipelineStatus: PipelineStatus
+        val errTopics = kafkaAdmin.ensureTopicsExist(steps)
+        if (errTopics == null) {
+            pipelineStatus = pipeline.start()
         } else {
-            pipelineStatus = previous.status
-            logger.warn("pipeline {pipelineName} with id {pipelineId} already exists", metadata.name, metadata.id)
-            if (pipelineStatus.isError) {
-                // do not try to resuscitate an existing pipeline if in a failed state
-                // it's up to the scheduler to delete it & reinitialize it, as it might require
-                // coordination with {model, pipeline}gateway
-                previous.stop()
-            }
+            pipelineStatus =
+                PipelineStatus.Error(null)
+                    .withException(errTopics)
+                    .withMessage("kafka streams topic creation error")
+            pipeline.stop()
         }
 
         // We don't want to mark the PipelineOperation.Create as successful unless the
@@ -215,7 +251,7 @@ class PipelineSubscriber(
         if (pipelineStatus !is PipelineStatus.Started) {
             pipelineStatus.isError = true
         }
-        pipelineStatus.log(logger, Level.INFO)
+        pipelineStatus.log(logger, Level.DEBUG)
         client.pipelineUpdateEvent(
             makePipelineUpdateEvent(
                 metadata = metadata,
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
@@ -182,6 +182,10 @@ class PipelineSubscriber(
                     metadata.id,
                     previous.status.getDescription(),
                 )
+                // Calling stop() here may be superfluous (depending on the state in which the pipeline is in),
+                // but we want to ensure that we clean up the KafkaStreams state of the pipeline because
+                // otherwise we have issues in re-starting it.
+                // Calling stop() on an already stopped pipeline is safe.
                 previous.stop()
             }
         } else { // pipeline doesn't exist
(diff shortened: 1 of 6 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -141,12 +141,57 @@ class PipelineSubscriber(
...
+        // concurrent creation of pipelines, this needs to be revisited.
+        if (pipelines.containsKey(metadata.id)) {
+            val previous = pipelines[metadata.id]!!
+            if (!previous.status.isError) {
+                client.pipelineUpdateEvent(
+                    makePipelineUpdateEvent(
+                        metadata = metadata,
+                        operation = PipelineOperation.Create,
+                        success = true,
+                        reason = previous.status.getDescription() ?: "pipeline created",
```
sakoush, line 159: given that this code is used also elsewhere (line 255 in this file) should we at least make `pipeline created` message as a constant?

sakoush: also this can be in a rebalance state / creating state? ie if the dataflow in loading the pipeline, the scheduler restarted and resending the load request to the same pipeline.

or the assumption as you mentioned that it is single threaded so we guarantee that the first load cmd has finished (it is not clear 100% for me).

sakoush (approved):
lgtm. I left a comment on a potential state inconsistency that is not clear to me.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
sakoush: Thanks for the detailed response, what happens if the scheduler disconnects while dataflow is processing `handleCreate`?

lc525: The results of tests here indicate reasonable behaviour: if the scheduler goes away while dataflow is in `handleCreate`, the pipeline creation logic continues to work correctly (the pipeline gets created, waited for to get into "Running" state, etc). Now, we have two cases, both ending up with the similar behaviour:

1. If the scheduler is not yet up when `handleCreate` calls the first grpc function (to update the scheduler state)
 - a io.grpc.StatusException is thrown from `handleCreate()`, which is caught in `subscribePipelines()`, which finishes by reporting the error
 - the functionality to retry connecting to the scheduler in `subscribe()` kicks in, and continues until a scheduler becomes available
 - when the new scheduler becomes available, dataflow connects to it, receives new "Create" messages, and responds with the actual status of the pipelines (which we failed to transmit initially). If the pipeline ended up in an error state, we retry creating it at this point
2. If a new scheduler instance has come up in the meantime
- the io.grpc.StatusException still gets thrown, because the previous grpc connection transport has dropped; a new connection is established via the retry mechanism in `subscribe()` and everything behaves as in case 1. above.

A note on how this is tested:
 - if one restarts dataflow-engine, the scheduler re-sends pipeline create messages to the new instance, therefore `handleCreate()` gets called. However, the creation process takes a long time (tens of seconds), because Kafka Streams needs to sync with remote state from Kafka topics. This gives us time to kill the scheduler while `handleCreate()` waits for the pipeline to start and transition to a `Running` state.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -141,12 +141,58 @@ class PipelineSubscriber(
...
+                    metadata.id,
+                )
+                logger.debug(
+                    "Previous state for non-active pipeline {pipelineName} version: {pipelineVersion}, id: {pipelineId}: {pipelineStatus}",
+                    metadata.name,
+                    metadata.version,
+                    metadata.id,
+                    previous.status.getDescription(),
+                )
+                previous.stop()
```
sakoush, line 185: maybe add a note in the code why do we need to call explicitly `stop`

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
```diff
@@ -104,4 +104,31 @@ open class PipelineStatus(val state: KafkaStreams.State?, var isError: Boolean)
 
     override var exception: Exception? = null
     override var message: String? = null
+
+    fun isActive(): Boolean {
+        return when {
+            this is StreamStarting || this is Started -> true
+            this.state in
```
sakoush, line 111: whats the difference between these two conditions?  is the second one happening at runtime perhaps?

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
```diff
@@ -104,4 +104,31 @@ open class PipelineStatus(val state: KafkaStreams.State?, var isError: Boolean)
...
+                    KafkaStreams.State.RUNNING,
+                )
+            -> true
+            else -> false
+        }
+    }
+
+    fun isError(): Boolean {
+        return when {
+            this is Error -> true
```
sakoush, line 124: ditto

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                    "StreamStopped(prevState=StreamStopping)",
+                    !IS_ACTIVE,
+                    !IS_ERROR,
+                    PipelineStatus.StreamStopped(PipelineStatus.StreamStopping()),
+                ),
+                arguments(
+                    "StreamStopped(prevState=nested StreamStopped without error)",
+                    !IS_ACTIVE,
+                    !IS_ERROR,
+                    PipelineStatus.StreamStopped(
```
sakoush, line 64: can we have this condition for real? multiple restarts perhaps to the scheduler?

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                        ),
+                    ),
+                ),
+                arguments(
+                    "StreamStopping",
+                    !IS_ACTIVE,
+                    !IS_ERROR,
+                    PipelineStatus.StreamStopping(),
+                ),
+                arguments(
```
sakoush, line 76: seems a dup of the one before?

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
```diff
@@ -0,0 +1,172 @@
...
+            that(statusIsError).isEqualTo(expectedIsError)
+        }
+    }
+
+    companion object {
+        private const val IS_ACTIVE = true
+        private const val IS_ERROR = true
+
+        @JvmStatic
+        fun checkState(): Stream<Arguments> =
```
sakoush, line 46: is there a way to parametrize choices as opposed to listing them?

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                    "PipelineStatus(state=non-error,active state, hasError=false)",
+                    IS_ACTIVE,
+                    !IS_ERROR,
+                    PipelineStatus(KafkaStreams.State.CREATED, false),
+                ),
+                arguments(
+                    "PipelineStatus(state=non-error,active state, hasError=true)",
+                    IS_ACTIVE,
+                    IS_ERROR,
+                    PipelineStatus(KafkaStreams.State.CREATED, true),
```
sakoush, line 156: how can it be created and has error?

sakoush (approved):
lgtm, thanks for adding tests. I left some questions but it is good to go.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/PipelineStatus.kt
lc525: The pipeline status is controlled both internally within the `Pipeline` object (for example when Kafka Streams transitions from one state to another), but can also be set externally (i.e if one determines via application logic external to the pipeline that an error has occured: for example, if we couldn't create all the Kafka topics required by the pipeline). 

`PipelineStatus.Error` is a way for objects external to the `Pipeline` to declare those kinds of errors. The other error states are controlled internally within the `Pipeline`.

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
lc525: The test is already parameterised (it is written once in the checkState function above -- the one outside the companion object). The following lists the states and the expected results for the given state. If one were to automate the way the expected results are created then you would need to test that the automation produces the expected expected results and so on... So I think further parameterising here would not help, but perhaps I'm misunderstanding your point

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
lc525: indeed, will remove

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
lc525: We don't typically encounter this, but the goal of "hasError" is to mark that there has been some exception/error message recorded together with the state. It could be that it's not a fatal error (the pipeline still got created), but there was an exception and we recovered.

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
lc525: Yes, in the sense that: A pipeline is in state `PipelineStatus.Started()`. It receives a stop, and then it has state `s1=PipelineStatus.StreamStopped(PipelineStatus.Started())`. (I'm simplifying here to show the nesting limitation idea, typically on stop there would be a transition from `Started()` to `StreamStopping()`)

Now, say the pipeline receives another stop. The new state will be initialised as `s2=PipelineStatus.StreamStopped(s1)`. `s2` is essentially what gets tested above. However, because in the implementation of `StreamStopped` we take measures not to deeply nest `StreamStopped` states,
`s2 == s1`

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/PipelineStatusTest.kt
lc525: After our discussion: it would be indeed worthwhile to think of ways of ensuring that all possible combinations have been covered; Those type of combinatorial tests are possible, but not for the case where the expected values need to be manually provided per test case.

PR #5556, opened 2024-04-30
fix: Upgrade go 1.21 and golint to 1.57

sakoush:
This PR moves to:

- go 1.21 as some deps are requiring this version eg. https://github.com/SeldonIO/seldon-core/pull/5554
- golint to 1.57 as the current version has memory issues locally

## TODO:

- [x] smoke test on kind


lc525 (approved):
lgtm

PR #5609, opened 2024-05-16
test(k6): Add k8s-based control plane tests 

lc525:

This PR adds a k6 test that arbitrarily creates/updates/deletes models in each VU. Starting from the scenario defined in `scenarios/core2_qa_control_plane_ops.js`, we can run a variety of control plane tests to check that Core 2 responds as expected to a wide array of commands/situations, including concurrent operations.

New configuration options introduced for k6 control-plane scenarios:
- `modelMaxReplicas` (set via environment var  MODEL_MAX_REPLICAS, array)
    each model type may be updated during the test to a maximum of replicas defined here
- `maxMemUpdateFraction` (set via environment var MAX_MEM_UPDATE_FRACTION, float)
     each model type may be updated by having, in the worst case, this much memory relative to 
     its default setting either added or removed. Example: For a model being assigned 8M in 
     `MODEL_MEMORY_BYTES`, a value of 0.5 in `maxMemUpdateFraction` means that the model
    may be created, at random with a memory requirement between 4M and 12M.
- `maxNumModelsHeadroom` (set via environment var MAX_NUM_MODELS_HEADROOM, array)
    models beyond `maxNumModels` (which are all created during the setup phase) which are allowed
    during control-plane testing
- `createUpdateDeleteBias` (set via environment var  MODEL_CREATE_UPDATE_DELETE_BIAS, array)
     3 element array defining the relative frequency with which CREATE/UPDATE/DELETE operations
     happen during control-plane tests
- `maxCreateOpsPerVU` (set via environment var  MAX_CREATE_OPS_PER_VU, int)
     because each VU runs independently, we set aside a namespace of this size for model IDs created
     by that VU. For example, if maxCreateOpsPerVU is 100, preloaded models get ids [0, 99], models
     created by VU 1 get ids [100, 199], by VU 2 get ids [200, 299], etc.
- `k8sDelaySecPerVU` (set via environment var  K8S_DELAY_SECONDS_PER_VU, int)
     the following delay is observed by each VU before starting a new iteration

Minor k6 tests fixes:
- allow 0 as an element of MAX_NUM_MODELS
- check model/pipeline status via k8s rather than relying on connection to scheduler
- fix kustomize warnings

- INFRA-1002 (internal issue) Control-plane tests

Tested by running control-plane tests in kind

### DIFF
--- a/tests/k6/components/k8s.js
+++ b/tests/k6/components/k8s.js
@@ -8,6 +8,7 @@ import { seldonObjectType } from '../components/seldon.js'
 import { sleep } from 'k6';
 
 const seldon_target_ns = getConfig().namespace;
+const MAX_RETRIES = 60;
 var kubeclient = null;
 var schedulerClient = null;
 
@@ -46,14 +47,31 @@ function seldonObjExists(kind, name, ns) {
     }
 }
 
-export function getModelsWithNamePrefix(modelNamePrefix) {
+/**
+ * getModels() can be used to get the models currently loaded in the configured
+ * namespace.
+ *
+ * When passing prefix/suffix constrains, a filtered list is returned.
+ *
+ * With suitable model names (i.e consistent naming per type, distinct prefixes
+ * amongst types), passing a namePrefix is used to filter models of a given type
+ *
+ * If each VU creating models uses a consistent naming scheme, appending an ID
+ * from a range assigned to that specific VU, nameSuffix may be used to retrieve
+ * all the models created by the current VU.
+ *
+ * In combination, it is suggested that namePrefix and nameSuffix may be used
+ * together to retrieve all models of a given type that are managed by a given
+ * VU
+ */
+export function getModels(namePrefix="", nameSuffix="") {
     try {
         const modelList = kubeclient.list(seldonObjectType.MODEL.description, seldon_target_ns)
         const modelNames = modelList.map((modelCR) => modelCR.metadata.name)
-        if(modelNamePrefix === ""){
+        if(namePrefix === "" && nameSuffix === ""){
             return modelNames
         }
-        const filteredModels = modelNames.filter((s) => s.startsWith(modelNamePrefix))
+        const filteredModels = modelNames.filter((s) => s.startsWith(namePrefix) && s.endsWith(nameSuffix))
         return filteredModels
     } catch (error) {
         console.log("K8S List Models Error:" + error)
@@ -66,8 +84,8 @@ export function loadModel(modelName, data, awaitReady=true) {
         kubeclient.apply(data)
         let created = kubeclient.get(seldonObjectType.MODEL.description, modelName, seldon_target_ns)
         if ('uid' in created.metadata) {
-            if (awaitReady && schedulerClient != null) {
-                awaitStatus(modelName, "ModelReady")
+            if (awaitReady) {
+                return awaitStatus(modelName, "ModelReady")
             }
         }
         return true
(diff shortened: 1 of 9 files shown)


tests/k6/scenarios/core2_qa_control_plane_ops.js
```diff
@@ -0,0 +1,206 @@
...
+    // Pick random operation in CREATE/UPDATE/DELETE, amongst available ones.
+    // Each operation is added multiple times in the selection array as
+    // configured in the `createUpdateDeleteBias` array. This defines the
+    // probability ratios between the operations. For example, the [1,4,3]
+    // createUpdateDeleteBias array makes an Update four times more likely then
+    // a Create, and a Delete 3 times more likely than the Create.
+    //
+    // Because each VU picks the operation independently, it's possible to
+    // temporarily get more models than MAX_NUM_MODELS for a given model type.
+    let availableOps = Array(config.createUpdateDeleteBias[1]).fill(seldonOpType.UPDATE)
```
sakoush, line 185: Can the list of ops contain zero items from a specific operation? Ie delete?

tests/k6/scenarios/core2_qa_control_plane_ops.js
```diff
@@ -0,0 +1,206 @@
+/**
+ * This test aims to verify that Core 2 remains in an error-free state after
```
sakoush, line 2: Nit: this scenario focuses only on models so far, should this reflected in its name or should this will be extended for pipelines and experiments as well?

sakoush (approved):
Nice one!

tests/k6/scenarios/core2_qa_control_plane_ops.js
lc525: I was thinking I will extend this for pipelines/experiments. But happy to have them as separate scenarios if we want to split things up. At the moment the pipelines tests are quite weak (just single model pipelines).

tests/k6/scenarios/core2_qa_control_plane_ops.js
lc525: Yes, any of the operations can be suppressed from the createUpdateDeleteBias array. It has three elements: `[a, b, c]`, where `a` corresponds to the "create bias", `b` corresponds to the "update bias" and `c` corresponds to the "delete bias". Setting one or some of them to 0 will completely supress the corresponding operation.

tests/k6/scenarios/core2_qa_control_plane_ops.js
```diff
@@ -0,0 +1,206 @@
...
+    if (opOk === true) {
+        switch(op) {
+            case seldonOpType.CREATE:
+                createCounter.add(1)
+                break;
+            case seldonOpType.UPDATE:
+                updateCounter.add(1)
+                break;
+            case seldonOpType.DELETE:
+                deleteCounter.add(1)
```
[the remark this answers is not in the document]
lc525: Indeed, although I was thinking of possible checks/invariants we could add to confirm that all is well.
We could have:
- 0 models (pipelines/experiments) remaining after teardown
- deleteCounter < createCounter
- all remaining models in an available state before teardown

tests/k6/components/k8s.js
```diff
@@ -31,52 +41,142 @@ function seldonObjExists(kind, name, ns) {
...
+    }
+    if ('status' in modelObj) {
+        let status = modelObj.status
+        if ('conditions' in status) {
+            for (let i = 0; i < status.conditions.length; i++) {
+                if(status.conditions[i].type == targetStatus) {
+                    return true
+                }
+            }
+            return false
```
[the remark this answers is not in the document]
lc525: Yes. Those ended up like this through incremental code changes, I'll clean things up.

PR #5636, opened 2024-05-23
feat: Add pipelines to core2 qa control plane tests

sakoush:

This PR updates k6 scenario `tests/k6/scenarios/core2_qa_control_plane_ops.js` with pipelines. This scenario is then usable with tests such as the one triggered by `make deploy-kpipeline-test`.

There are assumptions to note with the current change:
- The test that we currently have is for a single model pipelines
- To update a pipeline we induce a change to the pipeline CR that has no real effect (by setting `batch` to a random value)
- We delete pipelines 50% of the time when a delete of a model is required, to simulate a pipeline that is not available due to issues with models.

### DIFF
--- a/tests/k6/scenarios/core2_qa_control_plane_ops.js
+++ b/tests/k6/scenarios/core2_qa_control_plane_ops.js
@@ -178,19 +178,19 @@ function handleCtlOp(config, op, modelTypeIx, existingModels) {
         case seldonOpType.CREATE:
         case seldonOpType.UPDATE:
             opOk = k8s.loadModel(modelName, modelCRYaml, true)
-            if (config.isLoadPipeline) {
-                opOk = k8s.loadPipeline(generatePipelineName(modelName), pipelineCRYaml, true) && opOk
+            if (opOk === seldonOpExecStatus.OK && config.isLoadPipeline) {
+                opOk = k8s.loadPipeline(generatePipelineName(modelName), pipelineCRYaml, true)
             }
             break;
         case seldonOpType.DELETE:
             opOk = k8s.unloadModel(modelName, true)
-            if (config.isLoadPipeline) {
+            if (opOk === seldonOpExecStatus.OK && config.isLoadPipeline) {
                 // a model can go away while a pipeline is still loaded, we then simulate this behavior
                 // by not unloading the pipeline in 50% of the cases
                 // TODO: make it an environment variable?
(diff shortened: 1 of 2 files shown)


tests/k6/scenarios/core2_qa_control_plane_ops.js
```diff
@@ -127,8 +143,27 @@ function handleCtlOp(config, op, modelTypeIx, existingModels) {
...
+                            "steps": steps,
+                            "output": pipeline.spec.output
+                        }
+                    }
+                    pipelineCRYaml = yamlDump(newPipelineCRYaml)
+                }
+            } catch (err) {
                 // just continue test, another VU might have deleted the chosen model
+                console.log(`Failed to update model ${modelName}: ${err}`)
                 return false
```
lc525, line 167: Just a note that I need to change this return in my PR (now this function returns a seldonOpExecStatus rather than bool)

lc525: I can make that change in my follow-up PR though

lc525 (approved):
lgtm; there will be a follow-up PR from my end

tests/k6/scenarios/core2_qa_control_plane_ops.js
```diff
@@ -143,9 +178,21 @@ function handleCtlOp(config, op, modelTypeIx, existingModels) {
...
             break;
         case seldonOpType.DELETE:
             opOk = k8s.unloadModel(modelName, true)
+            if (config.isLoadPipeline) {
+                // a model can go away while a pipeline is still loaded, we then simulate this behavior
+                // by not unloading the pipeline in 50% of the cases
+                // TODO: make it an environment variable?
+                let unloadPipeline = Math.random() < 0.5 ? 0 : 1
+                if (unloadPipeline) {
+                    opOk = k8s.unloadPipeline(generatePipelineName(modelName), true) && opOk
```
lc525, line 193: opOk is not bool but seldonOpExecStatus

sakoush: fixed in https://github.com/SeldonIO/seldon-core/pull/5636/commits/afbdd3b3a3ec358ce02c4198e14e191b78e0b1cc

sakoush:
Tested locally in a kind cluster and  it looks fine.

PR #5638, opened 2024-05-23
fix(ansible): allow multiple custom secrets without namespaces

lc525:
Fix a bug where defining multiple custom secrets in an extra-vars file without explicit namespaces would add the default namespace only to the last one, and fail the playbook.

Previously, each iteration of the modified set_fact would overwrite the `seldon_custom_secrets` variable entirely rather than updating it (appending the default namespace to each secret defined in `custom_secrets`).


sakoush (approved):
lgtm

PR #5639, opened 2024-05-23
fix(deps): major version updates in /scheduler/data-flow

lc525:
- bump kotlin language version 1.8.20 -> 1.9.21
- bump kotlinx-coroutines-core 1.7.3 -> 1.8.1
- bump io.grpc:grpc-* 1.63.0 -> 1.64.0

The only remaining update that we would like to make to dataflow is bumping JDK 17 -> JDK 21. However, that update is currently blocked by the requirements of kafka-streams; The confluent platform libraries that we're using (kafka-streams:7.6.1-ccs) have only been tested up to JDK 17 according to: https://docs.confluent.io/platform/current/installation/versions-interoperability.html#java

- INFRA-721 (Internal): Bump dataflow dependency versions

- Tested in kind (pipelines smoke test). Same behaviour as before the updates.
- The unrelated code updates just touch the code used by a test, in order to eliminate some compilation warnings

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/TypeExtensions.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/TypeExtensions.kt
@@ -10,6 +10,7 @@ the Change License after the Change Date as each is defined in accordance with t
 package io.seldon.dataflow
 
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.DEFAULT_CONCURRENCY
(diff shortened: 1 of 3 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/TypeExtensions.kt
```diff
@@ -10,14 +10,15 @@ the Change License after the Change Date as each is defined in accordance with t
 package io.seldon.dataflow
 
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
```
sakoush, line 13: question: curious what is this adding?

sakoush (approved):
lgtm

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/TypeExtensions.kt
lc525: Some parallel operations on Flows (i.e parallel processing of elements in a flow, like the one we want to do for pipelines) are marked as experimental, with semantics that might change. We have a test that uses some of those to see how they behave.

PR #5703, opened 2024-06-20
fix(scheduler): Pipeline terminating fix on rebalance

sakoush:
This PR fixes a bug when a pipeline is in a terminating state gets recreated in the case of a dataflow subscription causing a rebalance. In this case we keep marking the pipeline as terminating and issue a delete operation to `dataflow-engine`.

Also added testing coverage for this part of the code base.

Fixes INFRA-1023 (internal)

### DIFF
--- a/scheduler/pkg/kafka/dataflow/server.go
+++ b/scheduler/pkg/kafka/dataflow/server.go
@@ -358,8 +358,14 @@ func (c *ChainerServer) rebalance() {
 			for server, subscription := range c.streams {
 				if contains(servers, server) {
 					msg.Op = chainer.PipelineUpdateMessage_Create
-					if err := c.pipelineHandler.SetPipelineState(pv.Name, pv.Version, pv.UID, pipeline.PipelineCreating, "Rebalance", sourceChainerServer); err != nil {
-						logger.WithError(err).Errorf("Failed to set pipeline state to creating for %s", pv.String())
+					// we do not need to set pipeline state to creating if it is already in terminating state, and we need to delete it
+					if pv.State.Status == pipeline.PipelineTerminating {
+						msg.Op = chainer.PipelineUpdateMessage_Delete
+					} else {
+						pipelineState := pipeline.PipelineCreating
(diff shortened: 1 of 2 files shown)


scheduler/pkg/kafka/dataflow/server_test.go
```diff
@@ -475,6 +475,165 @@ func TestPipelineEvents(t *testing.T) {
...
+			g.Expect(err).To(BeNil())
+
+			//to allow events to propagate
+			time.Sleep(500 * time.Millisecond)
+
+			err = s.pipelineHandler.SetPipelineState(test.loadReq.Name, test.loadReq.Version, test.loadReq.Uid, test.status, "", sourceChainerServer)
+			g.Expect(err).To(BeNil())
+
+			//to allow events to propagate
+			time.Sleep(500 * time.Millisecond)
```
lc525, line 605: Are these the ugly sleeps you were telling me about? Not sure they are avoidable given the nature of the local (async) event hub. But (note to self) we should probably do some concurrent operation tests on a single object to make sure the state remains consistent. I believe we could even do that under the current consistency checking k6 test, by having max 1 model/pipeline.

scheduler/pkg/kafka/dataflow/server.go
```diff
@@ -358,8 +358,14 @@ func (c *ChainerServer) rebalance() {
 			for server, subscription := range c.streams {
 				if contains(servers, server) {
 					msg.Op = chainer.PipelineUpdateMessage_Create
-					if err := c.pipelineHandler.SetPipelineState(pv.Name, pv.Version, pv.UID, pipeline.PipelineCreating, "Rebalance", sourceChainerServer); err != nil {
-						logger.WithError(err).Errorf("Failed to set pipeline state to creating for %s", pv.String())
+					// we do not need to set pipeline state to creating if it is already in terminating state, and we need to delete it
+					if pv.State.Status == pipeline.PipelineTerminating {
```
lc525, line 362: Noted that we can't be in the "Terminated" state here because we only fetch running pipelines initially (with "Terminating" being considered one of the running states).

lc525 (approved):
lgtm

PR #5715, opened 2024-06-24
Update license.yml

sakoush:
Update license workflow to include a schedule as we have it defined in the v2 branch check [here](https://github.com/SeldonIO/seldon-core/blob/d825d00ae0ce089e282b69c15e09402d1fac7a31/.github/workflows/license.yml#L4-L6)


lc525 (approved):
lgtm

PR #5720, opened 2024-06-26
Revert "Update license.yml"

sakoush:
Reverts SeldonIO/seldon-core#5715

As this is only running the schedule for the `master` branch and not triggering v2.


sakoush:
will do a separate PR instead

PR #5722, opened 2024-06-26
Changes to license.yml required for v2 license check

sakoush:
This is required for v2 license checks and follows the pattern from https://github.com/SeldonIO/seldon-core/blob/master/.github/workflows/security_tests_v2.yml


lc525 (approved):
lgtm

PR #5726, opened 2024-06-27
fix: Deal with deleted experiments when restoring from cache

sakoush:
This PR fixes a bug that re-loads deleted experiments after scheduler restarts. This is further complicated by the fact that these reloaded experiments are only visible from the scheduler state and not from kubernetes state. 

The underlying cause was that we didn't check experiments state (whether they are deleted) when restoring from disk on scheduler restarts. We also didnt persist `Delete` status in the local embedded db state.

This PR adds a `Delete` field in the embedded db for experiments, which allows for this check on scheduler restarts.

We also need to consider migration path for the existing db:
 
For migration of the experiment local embedded db to the new version, as we didnt have a delete field to mark whether the experiment is deleted therefore the migration path was to delete all records for experiments (scheduler state) and allow the operator to reload experiments from the records stored in k8s `etcd`.

We added this recovery path to pipelines as well so that if the local db is deleted for any reasons pipelines can be recovered from `etcd`.

This PR also skips loading experiments if they fail validation. Importantly it  will not fail the scheduler from starting if this happens at validating a particular experiment fail. This is something that got exposed by this bug.

Note that for pipelines we do not have a validation step when restoring from disk and it has already a `Deleted`  field  in the pipeline embedded db that can be checked.

### Implementation 

- Added `ExperimentSnapshot` proto in `mlops/scheduler/storage.proto`, that has an extra field `Deleted`  to the experiment protos that we persist on disk so that on restore we can check whether the experiment is deleted. 
- Added also `get` helper from DB (badgerdb) so that we can tests whats stored on disk. I also increased tests coverage while working on this area of the codebase.
- Added controller logic to reload experiments and pipelines if there are no state in scheduler for them (if status grpc call returns with 0 items in the list).
- Added a special record in the embedded db to mark the version so that we can in the future add migration logic more explicitly.
- Increased test coverage for the controller `scheduler` sub-package, more test coverage to come in future PRs.

fixes: INFRA-1055 (internal)

TODO: 

- [x] Add migration path for experiments DB.
- [x] Test in kind setup the migration path.
- [x] Add similar logic for pipelines.


scheduler/pkg/store/experiment/db.go
```diff
@@ -73,12 +75,10 @@ func (edb *ExperimentDBManager) restore(startExperimentCb func(*Experiment) erro
 					return err
 				}
 				experiment := CreateExperimentFromRequest(&snapshot)
-				if err != nil {
-					return err
-				}
 				err = startExperimentCb(experiment)
 				if err != nil {
-					return err
+					// skip restoring the experiment if the callback returns an error
```
lc525, line 80: May I suggest we slightly alter the comment to avoid confusion: the code doesn't skip anything, it simply swallows the error and logs a warning. If we would have bubbled the error up, that would end up stopping the scheduler with a Fatal error in main().

I was thinking of something like: "If the callback fails, do not bubble the error up but simply log it as a warning. The experiment restore is skipped instead of the scheduler failing due to the returned error."

lc525 (approved):
lgtm, just a minor comment relative to the comment describing the code change.

scheduler/pkg/store/experiment/db.go
```diff
@@ -73,13 +76,46 @@ func (edb *ExperimentDBManager) restore(startExperimentCb func(*Experiment) erro
 					return err
 				}
 				experiment := CreateExperimentFromRequest(&snapshot)
+				if experiment.Deleted {
```
sakoush, line 79: this is the crux of the change. we store now `deleted` and on restoring we just add the experiment to the in-memory store without (re)starting it.

apis/go/mlops/agent/agent.pb.go
```diff
@@ -7,11 +7,10 @@ Use of this software is governed BY
 the Change License after the Change Date as each is defined in accordance with the LICENSE file.
 */
 
-
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.10
+// 	protoc-gen-go v1.34.2
+// 	protoc        v5.27.2
```
sakoush, line 13: @lc525 fyi

operator/scheduler/client.go
```diff
@@ -253,7 +253,7 @@ func (s *SchedulerClient) checkErrorRetryable(resource string, resourceName stri
 }
 
 func retryFn(
-	fn func(context context.Context, conn *grpc.ClientConn, namespace string) error,
+	fn func(context context.Context, grcpClient scheduler.SchedulerClient, namespace string) error,
```
sakoush, line 256: this change of interface to facilitate testing

operator/scheduler/model.go
```diff
@@ -271,87 +269,6 @@ func (s *SchedulerClient) SubscribeModelEvents(ctx context.Context, conn *grpc.C
 	return nil
 }
 
-func (s *SchedulerClient) handlePendingDeleteModels(
```
sakoush, line 274: moved to `utils.go`

scheduler/pkg/store/utils/utils.go
```diff
@@ -0,0 +1,60 @@
+/*
```
sakoush, line 1: note these helpers are tests in experiment and pipeline `db_test.go` via their wrappers.

operator/scheduler/experiment_test.go
```diff
@@ -0,0 +1,414 @@
...
+						g.Expect(req.GetName()).To(Equal(grpcClient.requests_experiments[idx].Experiment.GetName()))
+						activeResources++
+					}
+				}
+				g.Expect(len(grpcClient.requests_experiments)).To(Equal(activeResources))
+			} else {
+				g.Expect(len(grpcClient.requests_experiments)).To(Equal(0))
+			}
+
+			// check state is correct for each pipeline
```
lc525, line 350: nit: .."for each experiment"

operator/scheduler/experiment_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+					// we check if the experiement is not in k8s (existing_resources) then we should not act on it
+					if _, ok := isBeingDeleted[r.ExperimentName]; !ok {
+						g.Expect(err).ToNot(BeNil())
+					} else {
+						g.Expect(err).To(BeNil())
+					}
+
+					if r.CandidatesReady && r.Active && r.MirrorReady {
+						g.Expect(experiment.Status.IsReady()).To(BeTrue())
```
lc525, line 371: I think for those tests is perfectly fine, but in general I propose we move towards `BeTrueBecause` and `BeFalseBecause`, which can print more informative data on test failure. (On failure, `BeTrue` just prints "true" != "false"). [docs](https://onsi.github.io/gomega/#betruebecausereason)

lc525 (approved):
First, thank you for implementing the Experiment db migration in a way that is not disruptive on cluster updates. This sets us up nicely to be able to do such migrations cleanly in the future as well, so I think it was quite important to get right!

The added testing will improve our life considerably, and I suspect that the effort of adding testing at this stage will pay off (I know more can be done to increase coverage, but... let's take it incrementally).

Most (all?) of my comments are nits, so please feel free to ignore if you don't agree with some of them.

operator/scheduler/server.go
```diff
@@ -64,9 +63,8 @@ func (s *SchedulerClient) ServerNotify(ctx context.Context, server *v1alpha1.Ser
 }
 
 // note: namespace is not used in this function
-func (s *SchedulerClient) SubscribeServerEvents(ctx context.Context, conn *grpc.ClientConn, namespace string) error {
+func (s *SchedulerClient) SubscribeServerEvents(ctx context.Context, grcpClient scheduler.SchedulerClient, namespace string) error {
 	logger := s.logger.WithName("SubscribeServerEvents")
-	grcpClient := scheduler.NewSchedulerClient(conn)
 
 	stream, err := grcpClient.SubscribeServerStatus(
```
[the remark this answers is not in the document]
sakoush: the only caller `retryFn` is creating the grpcClient, however I think perhaps it should be tested. Similarly for the other  `Subscribe*Events` in `startEventHanders`. I will leave it as a future PR when we add test coverage for this part of the codebase?

operator/scheduler/utils_test.go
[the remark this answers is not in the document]
sakoush: yes I agree, I will leave it as a future PR if thats ok?

operator/scheduler/server.go
```diff
@@ -64,9 +63,8 @@ func (s *SchedulerClient) ServerNotify(ctx context.Context, server *v1alpha1.Ser
 }
 
 // note: namespace is not used in this function
-func (s *SchedulerClient) SubscribeServerEvents(ctx context.Context, conn *grpc.ClientConn, namespace string) error {
+func (s *SchedulerClient) SubscribeServerEvents(ctx context.Context, grcpClient scheduler.SchedulerClient, namespace string) error {
```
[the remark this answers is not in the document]
sakoush: did a search and replace for it.

operator/scheduler/experiment.go
```diff
@@ -26,31 +26,38 @@ import (
...
 		req,
 		grpc_retry.WithMax(SchedulerConnectMaxRetries),
 		grpc_retry.WithBackoff(grpc_retry.BackoffExponential(SchedulerConnectBackoffScalar)),
 	)
-	return err, s.checkErrorRetryable(experiment.Kind, experiment.Name, err)
+	return s.checkErrorRetryable(experiment.Kind, experiment.Name, err), err
 }
 
-func (s *SchedulerClient) StopExperiment(ctx context.Context, experiment *v1alpha1.Experiment) (error, bool) {
+func (s *SchedulerClient) StopExperiment(ctx context.Context, experiment *v1alpha1.Experiment, conn *grpc.ClientConn) (bool, error) {
```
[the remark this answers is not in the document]
sakoush: fixed in https://github.com/SeldonIO/seldon-core/pull/5726/commits/76b0f1f6c5253e8c956731cda91244b532a17595

operator/scheduler/pipeline.go
```diff
@@ -77,9 +79,8 @@ func (s *SchedulerClient) UnloadPipeline(ctx context.Context, pipeline *v1alpha1
 }
 
 // namespace is not used in this function
-func (s *SchedulerClient) SubscribePipelineEvents(ctx context.Context, conn *grpc.ClientConn, namespace string) error {
+func (s *SchedulerClient) SubscribePipelineEvents(ctx context.Context, grcpClient scheduler.SchedulerClient, namespace string) error {
```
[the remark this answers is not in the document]
sakoush: fixed in https://github.com/SeldonIO/seldon-core/pull/5726/commits/adb2ee84f9bffb8bc5b9e1891bf30ca3261304cc

operator/scheduler/experiment_test.go
sakoush: fixed in https://github.com/SeldonIO/seldon-core/pull/5726/commits/5946b88c2f1ebf3530a2be56a7682878e16036e4

PR #5747, opened 2024-07-09
fix(golang-lint): update lint config file to remove deprecated settings

lc525:
This removes a number of warnings when running recent golangci-lint versions


scheduler/Makefile
lc525, line 1: whitespace-only changes (ignore)

sakoush (approved):

PR #5749, opened 2024-07-11
build(grpc): update Makefile responsible for code generation

lc525:
Allow protoc and the language-specific plugins to be installed at specific versions locally and then use those versions to generate code.

This means we no longer depend on arbitrary versions installed on the system doing the code generation, making this reproducible.

The installs happen in the `apis/tools` directory, if the required tools are not found. Python packages are installed in a virtual environment under `apis/tools/python`

In this first phase, we retain the exact same tool versions that were used to generate the existing protobuf/grpc code. We also apply the licensing headers to those generated files automatically.

Because of a previous mismatch in how those licensing headers were generated, most generated files end up with a minor whitespace difference.

[IMPORTANT]: The kotlin/java codegen for `apis/mlops/v2_dataplane/v2_dataplane.proto` had not been previously generated using the makefile (it seems it was generated as a one-off). This means the version of the codegen tools used was already mismatched compared to the rest. Consequently, this PR adds this code generation to the Makefile, and re-generates the code files in order to bring everything to a consistent version.

**Issue(s) related to this PR**:

- INFRA-1076 - facilitate protobuf/gRPC codegen version updates

**Comments for reviewer**
- Most .pb.go and .kt files have one empty line removed after the license header. This makes them consistent with the other files where automatic licensing headers are being applied.
- Files in `apis/mlops/v2_dataplane/kotlin/*` are newly re-generated and contain non-whitespace differences


apis/Makefile
```diff
@@ -1,12 +1,115 @@
...
+
+################################################################################
+
+MAKEFILE := $(lastword $(MAKEFILE_LIST))
+BASE_DIR := $(realpath $(dir $(MAKEFILE)))
+TOOLS_DIR := $(BASE_DIR)/tools
+
+CMD_WGET   	?= wget
+CMD_UNZIP 	?= unzip
+CMD_GO			?= go
```
sakoush, line 56: nit: extra necessary tab?

sakoush (approved):
LGTM! Thanks for pinning these versions of protos generation, very useful moving forward.

apis/Makefile
```diff
@@ -1,12 +1,115 @@
...
+# https://mvnrepository.com/artifact/io.grpc/protoc-gen-grpc-kotlin
+PROTOC_GEN_KOTLIN_GRPC_VERSION ?= 1.2.1
+PROTOC_GEN_KOTLIN_GRPC_JDK ?= jdk7
+
+
+################################################################################
+
+MAKEFILE := $(lastword $(MAKEFILE_LIST))
+BASE_DIR := $(realpath $(dir $(MAKEFILE)))
+TOOLS_DIR := $(BASE_DIR)/tools
```
sakoush, line 52: nit: should this be `.tools` to make it explicit that these are not files to check in (we are guarded anyway with .gitignore)

lc525: Indeed, I thought of that briefly as well. Will update

PR #5761, opened 2024-07-16
fix(scheduler): Do not report back draining servers for status

sakoush:
There is an issue with "double" counting for models in the case of draining server replicas when then are reported with server statues. While the double counting is initially correct as we are loading models onto a different server replica while having the draining server server traffic. Once the drain process has finished and the draining server removed there is no event that triggers an update to the server statuses to reflect this change. 

As eventually the draining server replica is going to get removed, we decided to just not report draining servers back on server statuses. 

Also added a test to cover this edge case.

Fixes: Infra-1080 (internal)


lc525 (approved):
lgtm

PR #5763, opened 2024-07-17
fix(ci): Update changelog.yml

sakoush:
This PR updates workflow to commit changelog to v2 branch after releases.

Note that we use now [rhysd/changelog-from-release/action@v3](https://github.com/rhysd/changelog-from-release) to create the changelog which is slightly different from `auto-changelog` that is used when creating the releases initially.

Given that we manually update the release notes further, this is fine I think.

adapted from: https://github.com/SeldonIO/MLServer/blob/ce93346302fb3c521e1e92373af8962664370a1a/.github/workflows/publish.yml#L45-L72

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -3,40 +3,37 @@ name: Publish changelog to master on new release
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


lc525 (approved):
lgtm

PR #5770, opened 2024-07-18
feat(ansible): allow local (host) mounts in kind

lc525:
This allows mounting local directories in a one-node kind cluster. It can be
useful (for example) for being able to load local models rather than requiring
them to be first copied to MinIO or a cloud bucket.

The changes here only add the basic support at the kind-cluster level. A
separate PR will update the helm-charts to allow for Server podSpec changes,
including adding a volume that points to the local folder now mounted in kind.

You can now set the following additional variables (overwriting with -e when
calling ansible-playbook preferred):

- kind_local_mount (bool, default false): enable kind extraMounts option when creating the cluster
- kind_host_path(string, default /tmp/kind-cluster): the local path to mount
- kind_container_path(string, default /host-mount): the "containerPath"
extraMounts setting - this (somewhat confusingly) defines the value that should be set as
the path in `volumes[i].hostPath.path` when adding a volume to the pod spec. Example:

```
kind: Pod
...
spec:
  volumes:
    - name: host-models
      hostPath:
        path: {{ kind_container_path }}
  ...
  containers
    - image: ...
      volumeMounts:
        - name: host-models
          # the actual path inside the container
          mountPath: /mnt/host-models
```

- No ansible docs added (yet) because using this functionality requires helm-chart updates to be added in a separate PR. The ansible docs/examples will be added together with that PR.

### DIFF
--- a/ansible/playbooks/kind-cluster.yaml
+++ b/ansible/playbooks/kind-cluster.yaml
@@ -11,3 +11,8 @@
     kind_image_version: kindest/node:v1.29.2@sha256:51a1434a5397193442f0be2a297b488b6c919ce8a3931be0ce822606ea5ca245
 
     kind_kubectl_default_namespace: seldon-mesh
+    vars_file: vars/default.yaml
+
+  pre_tasks:
+  - name: "Load vars from {{ vars_file }}"
+    include_vars: "{{ vars_file }}"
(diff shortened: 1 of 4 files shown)


sakoush (approved):
lgtm

PR #5810, opened 2024-07-29
fix: Add experiments pipeline tests for envoy processor

sakoush:
This PR increases tests coverage to understand the behaviour of envoy when different resources change state.


scheduler/pkg/envoy/processor/incremental_test.go
```diff
@@ -356,6 +358,21 @@ func createTestModel(modelName string,
 	return f
 }
 
+func remoteTestModel(
```
lc525, line 361: should this be "removeTestModel" ?

lc525 (approved):
lgtm; with one question. Sorry for the delay in reviewing this

PR #5819, opened 2024-08-04
feat(scheduler): Add max elapsed duration for model load/unload

sakoush:
Previously we had a fixed elapsed duration defaulting to 15 minutes for model load retries. For loading large llms (e.g. llama3 70b), this fixed max elapsed time didnt allow for any retries consequently as the size of the model will take the duration of the load towards this limit. 

This change then introduces the following:
- Allowing to define a max elapsed duration for model load (defaults to 2 hours) and unload (defaults to 15 minutes)
- Increase an individual load timeout call to 1 hour.
- Reduced the number of retries to 5 from 10.

In the case of retries, hf can resume downloads in the case one download has not finished completely with a single call.

Note that this change doesnt make these values configurable yet for the user to specify.

Fixes: INFRA-1114 (internal).

### DIFF
--- a/scheduler/cmd/agent/main.go
+++ b/scheduler/cmd/agent/main.go
@@ -47,6 +47,8 @@ const (
 	maxElapsedTimeReadySubServiceBeforeStart = 15 * time.Minute // 15 mins is the default MaxElapsedTime
 	// period for subservice ready "cron"
 	periodReadySubService = 60 * time.Second
+	// max time to wait for a model server to load a model, including retrues
+	maxLoadElapsedTime = 120 * time.Minute
 	// number of retries for loading a model onto a server
 	maxLoadRetryCount = 10
 	// number of retries for unloading a model onto a server
@@ -275,6 +277,7 @@ func main() {
 			periodReadySubService,
 			maxElapsedTimeReadySubServiceBeforeStart,
 			maxElapsedTimeReadySubServiceAfterStart,
+			maxLoadElapsedTime,
 			maxLoadRetryCount,
 			maxUnloadRetryCount,
 		),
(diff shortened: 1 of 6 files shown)


scheduler/cmd/agent/main.go
```diff
@@ -47,6 +47,10 @@ const (
 	maxElapsedTimeReadySubServiceBeforeStart = 15 * time.Minute // 15 mins is the default MaxElapsedTime
 	// period for subservice ready "cron"
 	periodReadySubService = 60 * time.Second
+	// max time to wait for a model server to load a model, including retrues
```
lc525, line 50: nit s/retrues//retries

scheduler/pkg/agent/client_utils.go
```diff
@@ -112,7 +111,7 @@ func (b *backOffWithMaxCount) Reset() {
 }
 
 func (b *backOffWithMaxCount) NextBackOff() time.Duration {
-	if b.currentCount >= b.maxCount {
+	if b.currentCount >= b.maxCount-1 {
```
lc525, line 114: Just to confirm: does this now retry once if b.maxCount is one? or is it that we now consider the initial function call as part of the "maxCount"?

lc525 (approved):
lgtm, agreed that we should probably make this configurable

scheduler/pkg/agent/client_utils.go
lc525: I guess the question is whether for a config parameter named something like (`max*RetryCount`) one would expect a maximum number of `max*RetryCount` retries, or they would expect that the function runs a total of `max*RetryCount` times.

sakoush: I added a ticket to create docs to clarify these semantic differences

PR #5830, opened 2024-08-13
fix(agent): cleanup rclone directory on errors

lc525:
Previously, any error occurring after the downloading of the model artifact but
before the model was properly registered in the agent's model repository would
result in the rclone path not being cleaned-up. This leads to the PVC filling up
and requiring end-user manual intervention.

This change fixes this, by cleaning the rclone path even when errors occur.

- Fixes #INFRA-1126 (internal): rclone directory is not cleaned-up


sakoush (approved):
lgtm. Is there a way to test this change?

lc525:
I did add at least some minimal tests by improving on how we use httpmock.

PR #5857, opened 2024-08-30
fix(agent): cleaning up models that fail to load

driev:
If an error occurred whenever a model was loading, the artefacts belonging to the model would remain on disk in the `model` directory, potentially causing the disk to fill up unnecessarily. 

This PR adds a `cleanup` function, which is executed whenever an error that leads to a `LOAD_FAILED` event occurs.

### DIFF
--- a/scheduler/pkg/agent/client.go
+++ b/scheduler/pkg/agent/client.go
@@ -147,7 +147,6 @@ func NewClient(
 	drainerService interfaces.DependencyServiceInterface,
 	metrics metrics.AgentMetricsHandler,
 ) *Client {
-
 	opts := []grpc.CallOption{
 		grpc.MaxCallSendMsgSize(math.MaxInt32),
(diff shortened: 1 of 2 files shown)


lc525 (approved):
The changes lgtm.

If in the future we end up with more places where we return errors on `LoadModel` perhaps we'll move to a defer function with a named return value, where we test if the error is not null and do the cleanup (I've seen that in the initial commit you also had a named returned value).

driev:
> The changes lgtm.
> 
> If in the future we end up with more places where we return errors on `LoadModel` perhaps we'll move to a defer function with a named return value, where we test if the error is not null and do the cleanup (I've seen that in the initial commit you also had a named returned value).

Agreeed - a defer function was the inital approach, but this seems tidier for now due to: https://github.com/SeldonIO/seldon-core/pull/5857/files#diff-9b56b9f193a85a84314608136ea3c0f5eb0f64346b70c81ebef3481fccf7ba21R618.

PR #5865, opened 2024-09-02
fix(model-gateway): Create consumer if model can serve traffic

sakoush:

Currently we only  create then model gateway consumer when the model is explicitly set as Available. This is ok but it doesnt cover edge cases whether the model is partially available (e.g. during scale up when there is no extra server replica).

Fixes INFRA-1149 (internal)

### DIFF
--- a/scheduler/pkg/kafka/gateway/client.go
+++ b/scheduler/pkg/kafka/gateway/client.go
@@ -154,7 +154,8 @@ func (kc *KafkaSchedulerClient) SubscribeModelEvents() error {
 		logger.Infof("Received event name %s version %d state %s", event.ModelName, latestVersionStatus.Version, latestVersionStatus.State.State.String())
 
 		// if there are available replicas then we add the consumer for the model
-		// note that this will also get triggered if the model is already added and in this case it is a no-op
+		// note that this will also get triggered if the model is already added but there is a status change (e.g. due to scale up)
+		// and in the case then it is a no-op
 		if latestVersionStatus.State.GetAvailableReplicas() > 0 {
 			if kc.consumerManager.Exists(event.ModelName) {
 				logger.Debugf("Model consumer %s already exists", event.ModelName)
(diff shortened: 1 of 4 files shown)


scheduler/pkg/kafka/gateway/client.go
```diff
@@ -153,14 +153,19 @@ func (kc *KafkaSchedulerClient) SubscribeModelEvents() error {
 
 		logger.Infof("Received event name %s version %d state %s", event.ModelName, latestVersionStatus.Version, latestVersionStatus.State.State.String())
 
-		switch latestVersionStatus.State.State {
-		case scheduler.ModelStatus_ModelAvailable:
+		// if there are available replicas then we add the consumer for the model
+		// note that this will also get triggered if the model is already added and in this case it is a no-op
+		if latestVersionStatus.State.GetAvailableReplicas() > 0 {
```
lc525, line 158: I think this is entirely reasonable. The only question in my mind is whether we should deal differently with the case where the `availableReplicas < minReplicas`. Under the principle that if we can serve traffic we should, the > 0 condition is fine. The risk here is that we're going to overload the available replicas with traffic, but I guess autoscaling should kick in at that point...

scheduler/pkg/kafka/gateway/client.go
```diff
@@ -153,14 +153,19 @@ func (kc *KafkaSchedulerClient) SubscribeModelEvents() error {
 
 		logger.Infof("Received event name %s version %d state %s", event.ModelName, latestVersionStatus.Version, latestVersionStatus.State.State.String())
 
-		switch latestVersionStatus.State.State {
-		case scheduler.ModelStatus_ModelAvailable:
+		// if there are available replicas then we add the consumer for the model
+		// note that this will also get triggered if the model is already added and in this case it is a no-op
```
lc525, line 157: Just checking my understanding: this part also documents previous behaviour, right? As in, we weren't differentiating between already added models previously either.

sakoush: yes

scheduler/pkg/kafka/gateway/client.go
sakoush: While I can change it check for this condition that we are above min replicas, I think min replicas is specifically use for now for scaling down decisions and not necessarily representing requiring min load. I suggest to keep it as it is and will add a note.

PR #5872, opened 2024-09-04
adding pipeline and model name validation

driev:
Adding some validation to model and pipeline names. Only alphanumeric names with underscores and dashes are valid.

TODO : 
 - [ ] add model name validation check

### DIFF
--- a/scheduler/pkg/store/pipeline/errors.go
+++ b/scheduler/pkg/store/pipeline/errors.go
@@ -195,3 +195,11 @@ type PipelineInputErr struct {
 func (pie *PipelineInputErr) Error() string {
 	return fmt.Sprintf("pipeline %s input %s is invalid. %s", pie.pipeline, pie.input, pie.reason)
 }
+
+type PipelineNameValidationErr struct {
+	pipeline string
+}
+
+func (pnve *PipelineNameValidationErr) Error() string {
+	return fmt.Sprintf("pipeline %s does not have a valid name - it must be alphanmumeric and can contain underscores and dashes", pnve.pipeline)
+}
--- a/scheduler/pkg/store/pipeline/validate.go
+++ b/scheduler/pkg/store/pipeline/validate.go
@@ -10,6 +10,7 @@ the Change License after the Change Date as each is defined in accordance with t
 package pipeline
 
 import (
+	"regexp"
 	"strings"
 )
 
@@ -27,6 +28,9 @@ const (
 )
 
 func validate(pv *PipelineVersion) error {
+	if err := checkName(pv); err != nil {
+		return err
+	}
 	if err := checkStepsExist(pv); err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


scheduler/pkg/store/utils/validate.go
```diff
@@ -0,0 +1,11 @@
+package utils
+
+import "regexp"
+
+func CheckName(name string) bool {
+	ok, err := regexp.Match("^[a-zA-Z0-9][a-zA-Z0-9-_]*[a-zA-Z0-9]$", []byte(name))
```
lc525, line 6: The spec on what names are allowed in k8s is this one: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

k8s will already reject all names that don't obey the regex: `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*` (this is the regex that kubectl reports when trying to create a resource with an invalid name). In particular, upper-case letters and underscores are also excluded.

We'll still want to do this validation explicitly so that we deal in an uniform way with k8s and non-k8s deployments.

Beyond k8s's existing validation, we want to reserve the `.` for Seldon -- this is because internally we do header-based routing and a pipeline's name (for example) is referred to as '[name].pipeline`. 

I'm not sure why we don't just trim the element after the _last_ dot when post-processing such names (and allow dots in user-provided names). I don't have the context on whether there are any other reasons why we might reserve the `.`, but perhaps we take this as a separate issue to investigate.

scheduler/pkg/store/utils/validate_test.go
```diff
@@ -0,0 +1,58 @@
...
+			"an invalid name that doesn't begin with something alphanumeric",
+			"-----______1",
+			false,
+		},
+		{
+			"an invalid name with a dot",
+			"this-Is-some-NA.ME_that_is_valid1234-5",
+			false,
+		},
+	}
```
lc525, line 52: Suggest to add a test-case for uppercase letters failing the validation

lc525 (commented):
I think some minor changes are needed around the regex given what k8s supports (and corresponding test cases), but this lgtm as a structure otherwise.

scheduler/pkg/store/utils/validate.go
driev: Yes, some more investigation is necessary. It would be great to have this validation aligned with k8s.

scheduler/pkg/store/utils/validate_test.go
driev: Added it.

scheduler/pkg/store/utils/validate_test.go
```diff
@@ -0,0 +1,58 @@
...
+			"this-NAME_is_valid1234-5",
+			true,
+		},
+		{
+			"a valid numerical name",
+			"111111111111111111",
+			true,
+		},
+		{
+			"an invalid name that begins and ends with something alphanumeric",
```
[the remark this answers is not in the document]
driev: Fixed.

lc525 (approved):
lgtm

PR #5874, opened 2024-09-05
fix(envoy): upgrade models in experiment

sakoush:

This change addresses 2 issues related to `Experiments`:
- When upgrading to a new version of a model, an experiment route in envoy containing this model would still point to the old (and the new version). This is problematic as we unload the old versions after a new version in made available. 
- When upgrading to a new version of model and in an experiment that has a `Default` route pointing to this model, the old versions of the model  are not unloaded. This is incorrect behaviour.

The changes are respectively:

- When we process events to update an experiment due to a model change, we first remove an existing state (which had the old route) before creating the routes for the experiment which updated models etc.
- We always clean old versions of models regardless of how many routes existed for it. This was introduced initially to guard against 404s which is not true anymore as we have a first remove the routes from envoy (`UnloadEnvoyRequested`) before unloading a model (`UnloadRequested`).

Fixes # INFRA-1001 (internal)

### DIFF
--- a/scheduler/pkg/envoy/processor/incremental.go
+++ b/scheduler/pkg/envoy/processor/incremental.go
@@ -435,6 +435,13 @@ func (p *IncrementalProcessor) addExperiment(exp *experiment.Experiment) error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	routeName := fmt.Sprintf("%s.experiment", exp.Name)
+
+	// first clear any existing routes
+	if err := p.removeRouteForServerInEnvoyCache(routeName); err != nil {
+		logger.WithError(err).Errorf("Failed to remove traffic for experiment %s", routeName)
+		return err
+	}
+
 	if err := p.addTrafficForExperiment(routeName, exp); err != nil {
 		logger.WithError(err).Errorf("Failed to add traffic for experiment %s", routeName)
 		return err
@@ -640,6 +647,7 @@ func (p *IncrementalProcessor) modelUpdate(modelName string) error {
 func (p *IncrementalProcessor) callVersionCleanupIfNeeded(modelName string) {
 	logger := p.logger.WithField("func", "callVersionCleanupIfNeeded")
 	if routes, ok := p.xdsCache.Routes[modelName]; ok {
+		logger.Debugf("routes for model %s %v", modelName, routes)
 		activeRoutes := len(routes.Clusters)
 		if activeRoutes == 1 && p.versionCleaner != nil {
 			logger.Debugf("Calling cleanup for model %s", modelName)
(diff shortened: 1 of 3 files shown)


scheduler/pkg/envoy/processor/incremental.go
```diff
@@ -640,8 +647,8 @@ func (p *IncrementalProcessor) modelUpdate(modelName string) error {
 func (p *IncrementalProcessor) callVersionCleanupIfNeeded(modelName string) {
 	logger := p.logger.WithField("func", "callVersionCleanupIfNeeded")
 	if routes, ok := p.xdsCache.Routes[modelName]; ok {
-		activeRoutes := len(routes.Clusters)
-		if activeRoutes == 1 && p.versionCleaner != nil {
```
sakoush, line 644: We do not anymore need to do this check. It was also buggy because this didnt capture edge cases where the routes also point to a candidate model in the case of an experiment with a default model route.

scheduler/pkg/envoy/processor/incremental.go
```diff
@@ -435,6 +435,13 @@ func (p *IncrementalProcessor) addExperiment(exp *experiment.Experiment) error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	routeName := fmt.Sprintf("%s.experiment", exp.Name)
+
```
lc525, line 438: Does this code (addExperiment) ever get called if there's no direct change to the experiment itself? I see it's called from `handleExperimentUpdates`, but I can't find model version updates triggering something to publish on that hub topic.

sakoush: Yes check [here](https://github.com/SeldonIO/seldon-core/blob/5383368fc6a62ab9bc27aad28b6617ab41005143/scheduler/pkg/store/experiment/store.go#L173) which will trigger changes to models to update experiments as well. This is also covered by tests.

lc525: Yep, actually found the place and then just saw your update. Thanks! I did see the test as well, but that is just manually calling `experimentUpdate`

lc525 (commented):
lgtm;
I haven't done a smoke test in kind but everything looks ok code-wise. For some reason I thought this would require a larger change.

lc525 (approved):
lgtm;
I haven't done a smoke test in kind but everything looks ok code-wise. For some reason I thought this would require a larger change.

scheduler/pkg/envoy/processor/incremental.go
sakoush: oh I see, this was redundant in the test and I remove this explicit call to `experimentUpdate` now.

sakoush:
> lgtm; I haven't done a smoke test in kind but everything looks ok code-wise. For some reason I thought this would require a larger change.

I have done a smoke test in kind and added the steps for release testing.

PR #5875, opened 2024-09-06
feat(env): parameters exposed as env variables

abhimanyu003:
**What this PR does / why we need it**:

These parameters (defaults) are currently hardcoded [here](https://github.com/SeldonIO/seldon-core/blob/4b233f9fe0980779e26013abd0adb3e4e32bee63/scheduler/cmd/agent/main.go#L43-L57).

These parameters could be exposed as env variables and also allow the user to get them configured via helm.

**Which issue(s) this PR fixes**:

Fixes # INFRA-1115

PR #5889, opened 2024-09-10
feat(chart): exposed retry constants as helm chart

abhimanyu003:

This is a follow up PR of https://github.com/SeldonIO/seldon-core/pull/5875

These parameters (defaults) are currently hardcoded [here](https://github.com/SeldonIO/seldon-core/blob/4b233f9fe0980779e26013abd0adb3e4e32bee63/scheduler/cmd/agent/main.go#L43-L57).

These parameters could be exposed as env variables and also allow the user to get them configured via helm.

Fixes  # INFRA-1115


lc525:
Changes look ok to me at a first pass. Could you confirm whether you've already tested this PR by spinning up a Core 2 cluster (kind or klaw), with one of the values from the helm-chart updated and verifying that the agent was started with updated parameters?

abhimanyu003:
Hello @lc525  i'm able to test helm and its working fine.

```
helm upgrade seldon-core-v2 ./seldon-core-v2-setup -n seldon-mesh --set serverConfig.agent.maxElapsedTimeReadySubServiceAfterStartSeconds=10
```

image

lc525 (approved):
lgtm

PR #5892, opened 2024-09-12
fix(bug): time unit fix

abhimanyu003:

In a recent change #5875 we have created envars to configure parameters for different timeout logic. The init of these variables didnt respect their unit. 

Fixes # INFRA-1168


sakoush (approved):
LGTM. assuming it works for @driev in his tests.

PR #5893, opened 2024-09-12
fix(scheduler): Controller to update the list of servers to scheduler on reconnect

sakoush:

This change introduces the ability to send a list of servers from k8s `etcd` to the scheduler on connection. This is required to update the scheduler with the state of the work with regards to servers, which include also k8s metadata that is required for various logic in the scheduler.

In this change we also allow `ServerNotify` to send the list of servers in one go, which will be required in subsequent changes that will follow up from this PR. In any case it is more efficient.

Fixes: 
- INFRA-747  (internal, partial)
- INFRA-1161 (internal)

### DIFF
--- a/operator/scheduler/server.go
+++ b/operator/scheduler/server.go
@@ -25,7 +25,7 @@ import (
 
 func (s *SchedulerClient) ServerNotify(ctx context.Context, grpcClient scheduler.SchedulerClient, servers []*v1alpha1.Server) error {
 	logger := s.logger.WithName("NotifyServer")
-	if servers == nil || len(servers) == 0 {
+	if len(servers) == 0 {
 		return nil
 	}
 
--- a/operator/scheduler/server.go
+++ b/operator/scheduler/server.go
@@ -41,11 +41,15 @@ func (s *SchedulerClient) ServerNotify(ctx context.Context, server *v1alpha1.Ser
 	}
 
 	request := &scheduler.ServerNotifyRequest{
-		Name:             server.GetName(),
-		ExpectedReplicas: replicas,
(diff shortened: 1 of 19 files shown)


operator/scheduler/server.go
```diff
@@ -23,40 +23,54 @@ import (
...
-	var replicas int32
-	if !server.ObjectMeta.DeletionTimestamp.IsZero() {
-		replicas = 0
-	} else if server.Spec.Replicas != nil {
-		replicas = *server.Spec.Replicas
-	} else {
-		replicas = 1
+
+	if grpcClient == nil {
+		// we assume that all servers are in the same namespace
```
lc525, line 33: All servers associated with a given scheduler (connection) will be in the same namespace, right? As in, this is always a safe assumption to make under the current design, as far as I know. Useful to have the comment though in case we change things in the future.

operator/scheduler/server_test.go
```diff
@@ -0,0 +1,290 @@
...
+	g := NewGomegaWithT(t)
+
+	type test struct {
+		name               string
+		existing_resources []client.Object
+		results            []*scheduler.ServerStatusResponse
+		noSchedulerState   bool
+	}
+
+	// note expected state is derived in the test, maybe we should be explictl about it in the future
```
lc525, line 188: I would say this is ok. Having the expected state derived in the test here shows the **constraint** explicitly (correspondence between number of model replicas). Having the expected state declared would show the state explicitly but would "hide" the nature of the constraint -- which will be implemented manually and could be error prone.

I thus argue that in this case it might be better to have the expected state derived within the test.

lc525 (approved):
lgtm

PR #5918, opened 2024-09-18
feat(charts): adding metadata object to a seldonconfig component

### LINKED ISSUE #5185: SeldonV2 - Ability to set component level labels/annotations in SeldonConfig/SeldonRuntime
I noticed that both SeldonConfig and SeldonRuntime Custom Resources use [ v1.PodSpec](https://github.com/kubernetes/api/blob/v0.24.4/core/v1/types.go#L3050) to define podSpec for the components, and re-use the Annotations/Metadata set in the CustomResource in the Pod spec within Deployment/Statefulset. 

References: 
a) https://github.com/SeldonIO/seldon-core/blob/v2.6.0/operator/controllers/reconcilers/seldon/deployment_reconciler.go#L54..L55 
b) https://github.com/SeldonIO/seldon-core/blob/60b7401ae5052ccf183c1453575611085bca4bb3/operator/controllers/reconcilers/seldon/statefulset_reconciler.go#L89..L90

**Looking at this I understand there is no provision to set component specific labels - For example, if I wanted to set some custom labels/annotations for the seldon-scheduler, and some different labels/annotations for Envoy, I could not find a way to do so.**

My goal is to be able to run all the Pods (except Envoy) defined in SeldonV2 Runtime definition within Istio Service Mesh (i.e. inject istio sidecar). One option I considered was to add the Istio labels/annotations to the CustomResource definition but then I would not be able to selectively prevent istio-proxy injection to the Envoy pod. 

Could you please provide guidance for this?

driev:

This allows for granular customisations, which are described in https://github.com/SeldonIO/seldon-core/issues/5185 

Fixes #5185 

INFRA-642


k8s/helm-charts/seldon-core-v2-crds/templates/seldon-v2-crds.yaml
lc525, line 1: This CRD file is generated, via the Makefile in `[seldon-core-root]/k8s` (running `make create`). That make rule looks at files in `[seldon-core-root]/operator/config/crd`. 

Some of the files in that path are also generated by the operator (via kubebuilder), with the go sources in `[seldon-core-root]/operator/apis/mlops/v1alpha1`. 

So the typical workflow is to edit these source files first, build the operator via `make build` in the operator dir, then run `make create` in the `[seldon-core-root]/k8s` directory.

Sorry this is more cumbersome than perhaps is needed. There are also elements that don't rely on the operator source files so would be edited directly in `[seldon-core-root]/operator/config/crd/...`

lc525: For merging elements at the operator level (we do that merging for PodSpecs but not metadata), there is an example with a starting point in `[seldon-core-root]/operator/controller/reconcilers/common/patch_utils.go` (uses of `MergePodSpecs`, `MergeContainers`).

k8s/kustomize/helm-servers/patch_mlserver.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: mlops.seldon.io/v1alpha1
 kind: Server
 metadata:
   name: mlserver
+  labels:
```
lc525, line 5: is `labels: HACK_REMOVE_ME{{ toJson .Values.mlserver.labels }}` not possible?

driev: I don't think so - a valid label would be something like `HACK_REMOVE_ME: HACK_REMOVE_ME{{ toJson .Values.mlserver.labels }}`.

driev: This might work for a `serverconfig.component` though...

lc525: Got it now, kustomize is a bit smarter regarding labels than we would have liked in this particular case.

For the ugliest hack immaginable,
`labelz: HACK_REMOVE_ME{{....}}` would work under matadata, with us processing `labelz` via sed.....

k8s/Makefile
```diff
@@ -18,8 +18,8 @@ create-helm-charts:
 	kustomize build kustomize/helm-servers/ > ${HELM_SERVERS_BASE}/seldon-v2-servers.yaml
 	sed -i 's/\(.*\)\(imagePullSecrets:\).*/{{- with .Values.imagePullSecrets }}\n\1\2\n\1{{- toYaml . | nindent 8 }}\n{{- end }}/g' ${HELM_COMPONENTS_BASE}/seldon-v2-components.yaml
 	sed -i 's/HACK_REMOVE_ME//' ${HELM_COMPONENTS_BASE}/seldon-v2-components.yaml
-	sed 's/HACK_REMOVE_ME//' ${HELM_SERVERS_BASE}/seldon-v2-servers.yaml \
-		> ${HELM_SERVERS_BASE}/.seldon-v2-servers.yaml
+	sed 's/labelz:/labels:\n\ \ \ /' ${HELM_SERVERS_BASE}/seldon-v2-servers.yaml > ${HELM_SERVERS_BASE}/.seldon-v2-servers.yaml
```
driev, line 21: will do the same for annotationz...

k8s/Makefile
```diff
@@ -18,8 +18,9 @@ create-helm-charts:
...
```
lc525, line 21: Could we add a brief comment on why this ugly hack (which you can certainly blame on me) is needed?

operator/controllers/reconcilers/seldon/runtime_reconciler_test.go
```diff
@@ -94,7 +96,9 @@ func TestRuntimeReconcile(t *testing.T) {
 				Spec: mlopsv1alpha1.SeldonConfigSpec{
 					Components: []*mlopsv1alpha1.ComponentDefn{
 						{
-							Name: mlopsv1alpha1.PipelineGatewayName,
+							Name:        mlopsv1alpha1.PipelineGatewayName,
```
lc525, line 99: could we do anything to test that those labels are applied?

lc525 (approved):
lgtm with some minor comments

operator/controllers/reconcilers/seldon/runtime_reconciler_test.go
driev: added a test

k8s/Makefile
driev: done

PR #5930, opened 2024-09-23
fix(scheduler): Scheduler wait  on server connect

sakoush:

This change introduces a `Synchroniser` on the scheduler side to coordinate the startup process and especially to attempt to wait for expected model servers to connect before proceeding with any control plane operations. This is important in the case that the scheduler restarts so that we do no interrupt ongoing dataplane operations that should be valid to be served.

There are 2 implementations for this `Synchroniser`:
- `SimpleSynchroniser`: a timeout based sync for non k8s deployments
- `ServerBasedSynchroniser`: this is for k8s deployments that tries to match expected server states (from `etcd` k8s controller) with actual agents connecting to the scheduler. More description provided [here](https://github.com/sakoush/seldon-core/blob/f33d980eec666780204e8ba12b49d31fae115201/scheduler/pkg/synchroniser/servers_sync.go#L10-L33).

### Other changes in this PR:

- Explicitly mark the first `ServerNotify` event from the controller to indicate the list of expected servers collected from k8s etcd.
- Add ServerEvent in event hub so that the system propagates events related to servers, which are then consumed by the `Synchronise`. Thanks to @lc525 for this change.
- More test coverage for related part of the code base.
- Added param for helm to specify this value.

This PR builds on top of https://github.com/SeldonIO/seldon-core/pull/5893.

Fixes #
INFRA-747 (internal)


sakoush:
on `2.8.3`
image

with this change
image

scheduler/pkg/coordinator/model_test.go
sakoush, line 1: renamed to `events_test.go` and added more test cases.

scheduler/cmd/scheduler/main.go
```diff
@@ -223,17 +219,53 @@ func main() {
 		log.Warn("Not running with scheduler local DB")
 	}
 
-	s := schedulerServer.NewSchedulerServer(logger, ss, es, ps, sched, eventHub)
+	// Setup synchroniser
+	var sync synchroniser.Synchroniser
+
+	if namespace == "" {
```
lc525, line 225: Strictly speaking the namespace can be set via a command-line flag, but I think testing it with the empty string is still the right thing to do here?

scheduler/cmd/scheduler/main.go
```diff
@@ -223,17 +219,53 @@ func main() {
...
+	as := agent.NewAgentServer(logger, ss, sched, eventHub, !autoscalingDisabled)
 	err = as.StartGrpcServer(allowPlaintxt, agentPort, agentMtlsPort)
 	if err != nil {
 		log.WithError(err).Fatalf("Failed to start agent gRPC server")
 	}
 
+	// wait for model servers to be ready
+	sync.WaitReady()
+
+	// Start envoy xDS server
```
lc525, line 258: Moved at the end so that until we have a consistent state, the data-plane routing remains unchanged.

lc525 (approved):
lgtm, with some really minor comments. I think this addition is quite valuable in making the scheduler behave a lot better on restarts!

scheduler/pkg/synchroniser/servers_sync.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (s *ServerBasedSynchroniser) handleServerEvents(event coordinator.ServerEventMsg) {
+	logger := s.logger.WithField("func", "handleServerEvents")
+	logger.Debugf("Received sync for server %s", event.String())
+
+	// we do not want to block the event handler while waiting for the signal to be fired as it may cause a deadlock with
+	// other events handlers.
+	// we also do not care about order of events, so we can safely spawn a go routine to handle the signal
+	go func() {
```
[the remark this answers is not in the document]
sakoush: fixed in [2f66806](https://github.com/SeldonIO/seldon-core/pull/5930/commits/2f66806436e4ee82b81149824a20f660ad3464bc)

PR #5931, opened 2024-09-23
feat(docs): adding a mention of per component labels and annotations to the docs

### LINKED ISSUE #5185: SeldonV2 - Ability to set component level labels/annotations in SeldonConfig/SeldonRuntime

driev:

Ih the per component labels and annotations changes in the docs.

Fixes #5185 

Depends on https://github.com/SeldonIO/seldon-core/pull/5918

### DIFF
--- a/operator/config/seldonconfigs/default.yaml
+++ b/operator/config/seldonconfigs/default.yaml
@@ -44,10 +44,6 @@ spec:
       serviceAccountName: seldon-scheduler
       terminationGracePeriodSeconds: 5
   - name: seldon-envoy
-    annotations:
-        prometheus.io/path: "/stats/prometheus"
-        prometheus.io/port: "9003"
-        prometheus.io/scrape: "true"
     replicas: 1
     podSpec:
       containers:
--- a/operator/config/seldonconfigs/default.yaml
+++ b/operator/config/seldonconfigs/default.yaml
@@ -45,7 +45,7 @@ spec:
       terminationGracePeriodSeconds: 5
   - name: seldon-envoy
     annotations:
-        prometheus.io/path: /stats/prometheus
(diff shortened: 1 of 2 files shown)


operator/config/seldonconfigs/default.yaml
lc525, line 1: this should be moved to the other PR

driev: all done...

lc525 (approved):
lgtm

PR #5932, opened 2024-09-23
feat(operator): add Model selector for scale subresource to enable HPA-based scaling

lc525:
- updates the Model CRD to contain a pod selector in the scale subresource
- sets the selector to a label `server=[inference-server-name]` matching no actual pods
- docs [to be moved to gitbook before merging]

**Which issue(s) this PR fixes**:
Fixes #1190 (internal): allow HPA-based Model autoscaling

**Special notes for your reviewer**:
- docs added to the PR to show how HPA would work
- tested in kind (small-scale)

**TODO**:
- [ ] test availability during HPA scale-up/scale-down via k6 load test (can be done after merging)

PR #5935, opened 2024-09-25
feat(docs): add documentation for HPA-based autoscaling

lc525:
This describes a solution for scaling both Models and Servers based on HPA for
the case of single-model serving. In the example described in the docs, the
scaling is done based on Model RPS metrics fetched from Prometheus.

Fixes #INFRA-1190 (docs)


lc525:
Reviewed and approved by Sherif here: https://github.com/SeldonIO/seldon-core/pull/5932

PR #5936, opened 2024-09-25
feat(envoy): closing off the admin interface

driev:

Only expose `/stats` and `/ready` on port `9003`.

Fixes #

Tested this locally and it starts up fine.


lc525 (commented):
Do we need to also update the k8s configs in `scheduler/k8s/envoy/`? Or do we keep that unchanged for `/stats` and `/ready`?

driev:
> Do we need to also update the k8s configs in `scheduler/k8s/envoy/`? Or do we keep that unchanged for `/stats` and `/ready`?

They can remain unchanged.... although, I've updated some of the named ports and other small things like that, so they're a bit more specific.

lc525 (approved):
lgtm

PR #5937, opened 2024-09-25
Use latest released version in security_tests.yml

sakoush:
For core 1 security scans use v1.8.2 (latest)


cherrymu (approved):
LGTM, we have just bumped the version of the image tags for new scan

PR #5940, opened 2024-09-27
feat: adding -ldflag and -trimpath to go build

abhimanyu003:

It will reduce the binary size of output and will also remove prefixes from recorded source file paths.

This can be particularly useful for creating reproducible builds and for privacy reasons, as it strips out paths that might contain sensitive information like usernames or directory structures.

### DIFF
--- a/components/tls/Makefile
+++ b/components/tls/Makefile
@@ -18,7 +18,7 @@ lint: ${.GOLANGCILINT_PATH}/golangci-lint
 
 .PHONY: build
 build: test
-	go build -trimpath -ldflags="-s -w" ./pkg/...
+	go build -trimpath -ldflags="-w" ./pkg/...
 
 .PHONY: licenses/dep.txt
 licenses/dep.txt:
--- a/components/tls/Makefile
+++ b/components/tls/Makefile
@@ -18,7 +18,7 @@ lint: ${.GOLANGCILINT_PATH}/golangci-lint
 
 .PHONY: build
 build: test
-	go build ./pkg/...
+	go build -trimpath -ldflags="-s -w" ./pkg/...
 
 .PHONY: licenses/dep.txt
 licenses/dep.txt:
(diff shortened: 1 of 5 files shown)


components/tls/Makefile
```diff
@@ -18,7 +18,7 @@ lint: ${.GOLANGCILINT_PATH}/golangci-lint
 
 .PHONY: build
 build: test
-	go build ./pkg/...
+	go build -trimpath -ldflags="-s -w" ./pkg/...
```
lc525, line 21: Using -w (strip DWARF info) is justified, but I would argue against -s (stripping debug symbols). The debug symbols can offer us good information if things fail in production (i.e the customer gets a stack trace) and we need to figure what went wrong from looking at the logs. I know `-w -s` is common for production golang binaries, but I don't see the value of -s compared to the disadvantages.

Furthermore, the debug symbols alone do not take much space, and shouldn't really reveal sensitive information either.

Ideally we can also create detached debuginfo files (i.e based on no -w) per release for ourselves to go along the binaries in case we ever need to debug challenging issues (let's discuss this outside this PR)

abhimanyu003: I've updated the PR to remove the -s flag, as per the requirement. My initial thought was that since most things run in containers, we might not often need to check the assembly output. Even in the case of a code panic, we would still get the stack trace.

lc525: It wasn't exactly a requirement, more of a good thing to have. It's not about looking at the assembly output of the binaries, but about stack traces that may be printed as part of errors or during panics.

Not having the symbols would of course not stop the stack traces from appearing, but it will make them less useful: instead of them showing the function names & location across all the stack frames, we would just get memory addresses and byte offsets.

hodometer/Makefile
```diff
@@ -52,7 +52,7 @@ build-hodometer-local: copy-components build-hodometer
 
 .PHONY: build-receiver
 build-receiver: test-receiver
-	$(GO_ENV) go build -o bin/receiver ./cmd/receiver
+	$(GO_ENV) go build -trimpath -ldflags="-s -w" -o bin/receiver ./cmd/receiver
```
lc525, line 55: there's still a `-s` here

lc525 (approved):
lgtm once the last `-s` that remained is also removed. 

Thank you for this PR, it's really good to make incremental improvements like these when we notice that things can be easily improved!

hodometer/Makefile
abhimanyu003: Thanks for noticing.

PR #5943, opened 2024-09-30
Merge GitBook Docs with V2 master

ramonpzg:
This PR contains two changes:
1. The first is to merge the GitBook docs in `v2-docs-master` with those in the `v2` using a different directory name, `docs-gb`, to separate old and new ones while the old ones get phased out..
2. The second part is to add to the `hack` directory a python script that searches for all of the GitBook URLs that point to source code and it changes the branch name to that one in the tag release.

For example, the following function will check for the current git tag
```python
import os
import re
import subprocess
from typing import List, Tuple

def get_current_git_tag() -> str:
    return subprocess.check_output(['git', 'describe', '--tags', '--abbrev=0']).decode('utf-8').strip()
```
and return
image

The following functions will search find al of the files, search through all of them for URLs, update them, and return new files.

```python
def get_markdown_files(directory: str) -> List[str]:
    """
    Recursively find all markdown files in the given directory and all sub-directories.
    """
    markdown_files = []
    for root, _, files in os.walk(directory):
        for file in files:
            if file.endswith('.md'):
                markdown_files.append(os.path.join(root, file))
    return markdown_files

def update_urls(content: str, new_version: str) -> Tuple[str, int]:
    pattern = r'(https://github\.com/SeldonIO/seldon-core/blob/v)(\d+(?:\.\d+)*)(/.+)'
    
    def replace_version(match):
        prefix = match.group(1)
        old_version = match.group(2)
        suffix = match.group(3)
        return f"{prefix}{new_version[1:]}{suffix}"
    
    updated_content, count = re.subn(pattern, replace_version, content)
    return updated_content, count

def process_file(file_path: str, new_version: str) -> int:
    with open(file_path, 'r') as file:
        content = file.read()

    updated_content, count = update_urls(content, new_version)

    if count > 0:
        with open(file_path, 'w') as file:
            file.write(updated_content)

    return count
```
The implementation looks as follows.

```python
def main():
    codebase_dir = './docs-gb'  # GitBook directory, change if needed
    new_version = get_current_git_tag()

    markdown_files = get_markdown_files(codebase_dir)
    total_replacements = 0

    for file_path in markdown_files:
        replacements = process_file(file_path, new_version)
        if replacements > 0:
            print(f"Updated {replacements} URL(s) in {file_path}")
            total_replacements += replacements

    print(f"\nTotal replacements made: {total_replacements}")
    print(f"New version: {new_version}")

if __name__ == "__main__":
    main()
```

image


lc525:
I see some files are deleted under `/docs`. Is this intentional?

ramonpzg:
> I see some files are deleted under `/docs`. Is this intentional?

Let me double-check why that k8s file is missing. It shouldn't be.

ramonpzg:
@lc525 -- I have added the latest docs from the `v2` branch. Please let me know if there are other changes I should look into.

lc525:
I'll make some further minor changes to `v2-docs-master` directly ;-) Thank you very much for putting this PR together

lc525 (approved):
lgtm

PR #5944, opened 2024-09-30
fix(controller): Adjust controller -> scheduler state recreation upton scheduler disconnect.

sakoush:

This PR changes the way we are resetting scheduler / controller state upon connection drops. Previously the state reconstruction was tightly couples with stream handling and this had a few issues:
- blutring of concerns as establishing streams didnt overlap really with state reconstruction
- Not all streams disconnected (because of the way we handled errors) and therefore it is not always that we recreate the entire state upon disconnection.

The solution is to separate out the logic and trigger state reconstruction if _any_ of the streams disconnects. These oeprations are idempotent and therefore it is fine to send them anyway. 

Fixes #

INFRA-1198 (internal).

### DIFF
--- a/operator/scheduler/client.go
+++ b/operator/scheduler/client.go
@@ -148,27 +148,27 @@ func (s *SchedulerClient) startEventHanders(namespace string, conn *grpc.ClientC
 			// absorb signals from the same disconnected go routines
 			time.Sleep(1 * time.Second)
 			// on new reconnects we send a list of servers to the schedule
-			err := retryFnWithSchedulerState(handleRegisteredServers, conn, namespace, s.logger.WithName("handleRegisteredServers"), s)
+			err := retryFn(s.handleRegisteredServers, conn, namespace, s.logger.WithName("handleRegisteredServers"))
 			if err != nil {
 				s.logger.Error(err, "Failed to send registered server to scheduler")
 			}
 
 			if err == nil {
-				err = retryFnWithSchedulerState(handleExperiments, conn, namespace, s.logger.WithName("handleExperiments"), s)
+				err = retryFn(s.handleExperiments, conn, namespace, s.logger.WithName("handleExperiments"))
 				if err != nil {
 					s.logger.Error(err, "Failed to send experiments to scheduler")
 				}
 			}
 
 			if err == nil {
-				err = retryFnWithSchedulerState(handlePipelines, conn, namespace, s.logger.WithName("handlePipelines"), s)
+				err = retryFn(s.handlePipelines, conn, namespace, s.logger.WithName("handlePipelines"))
 				if err != nil {
 					s.logger.Error(err, "Failed to send pipelines to scheduler")
 				}
 			}
 
 			if err == nil {
-				err = retryFnWithSchedulerState(handleModels, conn, namespace, s.logger.WithName("handleModels"), s)
+				err = retryFn(s.handleModels, conn, namespace, s.logger.WithName("handleModels"))
 				if err != nil {
 					s.logger.Error(err, "Failed to send models to scheduler")
 				}
@@ -324,27 +324,6 @@ func retryFn(
 	fnWithArgs := func() error {
 		grpcClient := scheduler.NewSchedulerClient(conn)
 		return fn(context.Background(), grpcClient, namespace)
-	}
-	err := backoff.RetryNotify(fnWithArgs, backOffExp, logFailure)
-	if err != nil {
-		logger.Error(err, "Failed to connect to scheduler", "namespace", namespace)
-		return err
-	}
-	return nil
-}
-
-func retryFnWithSchedulerState(
-	fn func(ctx context.Context, namespace string, s *SchedulerClient, grpcClient scheduler.SchedulerClient) error,
-	conn *grpc.ClientConn, namespace string, logger logr.Logger, state *SchedulerClient,
-) error {
-	logger.Info("Retry sending initital state to scheduler", "namespace", namespace)
-	logFailure := func(err error, delay time.Duration) {
-		logger.Error(err, "Scheduler not ready")
-	}
-	backOffExp := backoff.NewExponentialBackOff()
-	fnWithArgs := func() error {
-		grpcClient := scheduler.NewSchedulerClient(conn)
-		return fn(context.Background(), namespace, state, grpcClient)
 	}
 	err := backoff.RetryNotify(fnWithArgs, backOffExp, logFailure)
 	if err != nil {
--- a/operator/scheduler/client.go
+++ b/operator/scheduler/client.go
@@ -94,7 +94,7 @@ func (s *SchedulerClient) startEventHanders(namespace string, conn *grpc.ClientC
 	// Subscribe the event streams from scheduler
 	go func() {
 		for {
-			err := retryFn(s.SubscribeModelEvents, conn, namespace, s.logger)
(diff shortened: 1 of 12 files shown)


operator/scheduler/client.go
```diff
@@ -108,14 +113,54 @@ func (s *SchedulerClient) startEventHanders(namespace string, conn *grpc.ClientC
...
+			} else {
+				s.logger.Info("Subscribe ended for control plane events", "namespace", namespace)
+			}
+		}
+	}()
+}
+
+func (s *SchedulerClient) handleStateOnReconnect(context context.Context, grpcClient scheduler.SchedulerClient, namespace string) error {
+	// on new reconnects we send a list of servers to the schedule
+	err := s.handleRegisteredServers(context, grpcClient, namespace)
```
lc525, line 138: For a future PR: all those contexts end up as context.Background(), so not cancellable.

operator/scheduler/control_plane_test.go
```diff
@@ -0,0 +1,288 @@
...
+		existing_resources              []client.Object
+		expected_requests_pipelines     []*scheduler.LoadPipelineRequest
+		expected_requests_models        []*scheduler.LoadModelRequest
+		expected_requests_models_unload []*scheduler.UnloadModelRequest
+		expected_requests_servers       []*scheduler.ServerNotify
+		expected_requests_experiments   []*scheduler.StartExperimentRequest
+	}
+	now := metav1.Now()
+
+	// note expected state is derived in the test, maybe we should be explictl about it in the future
```
lc525, line 97:
```suggestion
	// note expected state is derived in the test, maybe we should be explicit about it in the future
```

lc525: Most of the expected state is still declared, right? I assume this is about the deleted pipeline and experiment?

operator/scheduler/experiment.go
```diff
@@ -76,25 +76,15 @@ func (s *SchedulerClient) StopExperiment(ctx context.Context, experiment *v1alph
 func (s *SchedulerClient) SubscribeExperimentEvents(ctx context.Context, grpcClient scheduler.SchedulerClient, namespace string) error {
```
lc525, line 76: I quite like that these functions now end up less entangled with code reconciling the state of the scheduler

operator/scheduler/model_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	type test struct {
+		name               string
+		existing_resources []client.Object
+		results            []*scheduler.ModelStatusResponse
+		noSchedulerState   bool
+	}
+	now := metav1.Now()
+
+	// note expected state is derived in the test, maybe we should be explictl about it in the future
```
lc525, line 37:
```suggestion
	// note expected state is derived in the test, maybe we should be explicit about it in the future
```

operator/scheduler/model_test.go
```diff
@@ -0,0 +1,256 @@
...
+					if r.Versions[0].State.GetState() == scheduler.ModelStatus_ModelAvailable {
+						g.Expect(model.Status.IsReady()).To(BeTrueBecause("Model state is ModelAvailable"))
+					} else {
+						g.Expect(model.Status.IsReady()).To(BeFalseBecause("Model state is not ModelAvailable"))
+					}
+
+					g.Expect(uint32(model.Status.Replicas)).To(Equal(r.Versions[0].State.GetAvailableReplicas() + r.Versions[0].State.GetUnavailableReplicas()))
+					g.Expect(model.Status.Selector).To(Equal("server=" + r.Versions[0].ServerName))
+				} else {
+					model := &mlopsv1alpha1.Pipeline{}
```
lc525, line 240: Shouldn't this be `&mlopsv1alpha1.Model{}`? I think this shows that the simple expect on `err != nil` is perhaps less strong than needed for some cases.

operator/scheduler/utils.go
```diff
@@ -35,34 +35,42 @@ func handleLoadedExperiments(
...
 		// experiments that are not in the process of being deleted has DeletionTimestamp as zero
 		if experiment.ObjectMeta.DeletionTimestamp.IsZero() {
 			s.logger.V(1).Info("Calling start experiment (on reconnect)", "experiment", experiment.Name)
-			if _, err := s.StartExperiment(ctx, &experiment, grpcClient); err != nil {
-				// if this is a retryable error, we will retry on the next connection reconnect
+			if retryable, err := s.StartExperiment(ctx, &experiment, grpcClient); err != nil {
 				s.logger.Error(err, "Failed to call start experiment", "experiment", experiment.Name)
+				if retryable {
+					// if this is a retryable error, we will retry on the next connection reconnect
+					return err
```
lc525, line 49: If I understand things correctly, perhaps the comment here is slightly misleading: this will not only retry on the next connection reconnect, but will actually stop all the other object states from being sync-ed to the scheduler and trigger the controller to resubscribe to the scheduler.

operator/scheduler/control_plane_test.go
sakoush: thats a stale comment, removed now.

operator/scheduler/model_test.go
sakoush: Good catch, fixed.

operator/scheduler/utils.go
sakoush: yes will reflect.

scheduler/pkg/server/control_plane.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	pb "github.com/seldonio/seldon-core/apis/go/v2/mlops/scheduler"
+)
+
+func (s *SchedulerServer) SubscribeControlPlane(req *pb.ControlPlaneSubscriptionRequest, stream pb.Scheduler_SubscribeControlPlaneServer) error {
+	logger := s.logger.WithField("func", "SubscribeControlPlane")
+	logger.Infof("Received subscribe request from %s", req.GetSubscriberName())
+
+	// on reconnect we send the current state of the servers to the subscriber (controller) as we may have missed events
```
lc525, line 20: Does this comment refer to the general process happening outside SubscribeControlPlane? (given that we just reply with an empty `ControlPlaneResponse`). Doesn't the scheduler wait for the status of the controller before starting to send back server status updates back?

sakoush: good catch I dont think this is a relevant comment, will remove.

scheduler/pkg/synchroniser/servers_sync.go
lc525, line 1: Might be worth to add a comment making explicit that the main assumption behind the ServerBasedSynchronizer is that if the controller says that it has x Servers, the scheduler will wait for x agent connections.

I'm unsure myself if things remain consistent if Servers are deleted or scaled up/down from k8s after the Controller has sent a number over to the scheduler. Is it possible we get fewer agent connection events to the scheduler and block forever?

lc525 (approved):
I think this improves things considerably from our current state, and hope all the testing doesn't show any issues.

I do have a question regarding the ServerBasedSynchronizer for a particular edge case, and a couple of other nits, but otherwise all looks good!

scheduler/pkg/synchroniser/servers_sync.go
sakoush: While the state can change as the scheduler is restarting, this should not block the sync process indefinitely as there is a timeout for the sync process. I am not sure how to also factor in changes in kube state while the scheduler is restarting but I assume that this might be a bit rare.

PR #5945, opened 2024-09-30
fix(operator): Use StatefulSet.Spec.Replicas as reference for Server status

lc525:
Before this change, we decide to mark the Server StatefulSet as ready by
comparing the `StatefulSet.Status.ReadyReplicas` with
`StatefulSet.Status.Replicas`. When the two are equal, we mark the StatefulSet
as "Ready".

However, this doesn't account for the possibility of `StatefulSet.Spec.Replicas`
to differ from `StatefulSet.Status.Replicas` (desired state vs actual state).

The change corrects this so that we only mark the StatefulSet as ready when
the number of ready replicas matches the desired state.

Fixes #CSM929; #INFRA-1197: Sync Server status to actual StatefulSet state


sakoush (approved):
LGTM.

operator/controllers/reconcilers/server/statefulset_reconciler.go
```diff
@@ -206,8 +206,8 @@ const (
 )
 
 func (s *ServerStatefulSetReconciler) GetConditions() []*apis.Condition {
-	ready := s.StatefulSet.Status.ReadyReplicas >= s.StatefulSet.Status.Replicas
-	s.Logger.Info("Checking conditions for stateful set", "ready", ready, "replicas", s.StatefulSet.Status.Replicas, "availableReplicas", s.StatefulSet.Status.AvailableReplicas)
+	ready := s.StatefulSet.Status.ReadyReplicas >= *s.StatefulSet.Spec.Replicas
+	s.Logger.Info("Checking conditions for stateful set", "ready", ready, "replicas", *s.StatefulSet.Spec.Replicas, "availableReplicas", s.StatefulSet.Status.AvailableReplicas)
```
sakoush, line 210: perhaps add `s.StatefulSet.Status.Replicas` in logging?

abhimanyu003: I feel like we should add a nil pointer check here to prevent from unwanted panic?

lc525: @abhimanyu003  The * here derefs `Replicas`, which is a pointer. In our code that is always set (either explicitly or as a default), so I didn't check it. However, I'll add the check so that it doesn't cause confusion and makes that constraint explicit.

s is also a pointer, but it is the function pointer receiver. In the codebase, we typically don't check this for nil because we follow the pattern

```
s, err := NewStatefulSetReconciler(....)
if (err != nil) {
    // handle error and return
}
....
// use s, which is not nil
```

PR #5946, opened 2024-10-01
feat: stateful set persistent volume claim retention policy

abhimanyu003:

This PR is to enhance this default behaviour from Core v2, such that a PVC should be released/deleted when the associated STS pod is scaled down or deleted. 

```
  persistentVolumeClaimRetentionPolicy:
    whenDeleted: Retain
    whenScaled: Retain
```

**Screenshot from unit test**

Pasted Graphic

**Quick info**

From k8s v1.27 we have a beta feature called PersistentVolumeClaim retention [StatefulSets](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention) which can be used to manage the pvc’s which are created by STS. In Core v2 we are creating the PVC for a model as a part of serverconfig's volumeClaimTemplates spec which will be present in the statefulset which gets created as soon as a server is created.

Fixes #INFRA-851

### DIFF
--- a/operator/controllers/reconcilers/server/statefulset_reconciler.go
+++ b/operator/controllers/reconcilers/server/statefulset_reconciler.go
@@ -41,6 +41,7 @@ func NewServerStatefulSetReconciler(
 	podSpec *v1.PodSpec,
 	volumeClaimTemplates []mlopsv1alpha1.PersistentVolumeClaim,
 	scaling *mlopsv1alpha1.ScalingSpec,
+	policy *appsv1.StatefulSetPersistentVolumeClaimRetentionPolicy,
 	serverConfigMeta metav1.ObjectMeta,
 	annotator *patch.Annotator,
 ) *ServerStatefulSetReconciler {
@@ -48,7 +49,7 @@ func NewServerStatefulSetReconciler(
 	annotations := utils.MergeMaps(meta.Annotations, serverConfigMeta.Annotations)
 	return &ServerStatefulSetReconciler{
 		ReconcilerConfig: common,
-		StatefulSet:      toStatefulSet(meta, podSpec, volumeClaimTemplates, scaling, labels, annotations),
+		StatefulSet:      toStatefulSet(meta, podSpec, volumeClaimTemplates, scaling, policy, labels, annotations),
 		Annotator:        annotator,
 	}
 }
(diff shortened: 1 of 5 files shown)


operator/controllers/reconcilers/server/statefulset_reconciler.go
```diff
@@ -41,14 +41,15 @@ func NewServerStatefulSetReconciler(
 	podSpec *v1.PodSpec,
 	volumeClaimTemplates []mlopsv1alpha1.PersistentVolumeClaim,
 	scaling *mlopsv1alpha1.ScalingSpec,
+	policy *appsv1.StatefulSetPersistentVolumeClaimRetentionPolicy,
```
lc525, line 44: A suggestion to rename this to volumeClaimRetentionPolicy. Mostly because in the future we might end up with more policies for different things, and to eliminate any confusion.

operator/controllers/reconcilers/server/statefulset_reconciler_test.go
```diff
@@ -463,7 +466,13 @@ func TestToStatefulSet(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			statefulSet := toStatefulSet(test.meta, test.podSpec, test.volumeClaimTemplates, test.scaling, test.labels, test.annotations)
+			statefulSet := toStatefulSet(test.meta,
+				test.podSpec,
+				test.volumeClaimTemplates,
+				test.scaling,
+				test.statefulSetPersistentVolumeClaimRetentionPolicy,
```
lc525, line 473: While you're referencing `test.statefulSetPersistentVolumeClaimRetentionPolicy` here, this remains null in all the test cases created above. I would add a test case for this.

abhimanyu003: I have updated a current test, to have PersistentVolumeClaimRetentionPolicy in it :)

operator/controllers/reconcilers/server/statefulset_reconciler.go
abhimanyu003: Sure, this is update and renamed to `volumeClaimRetentionPolicy`

lc525 (approved):
lgtm! Thanks for working on this, it will allow us to apply a PVC delete policy by default in some of our test clusters as well

PR #5948, opened 2024-10-02
feat(scheduler):  set a ttl on deleted pipelines and experiments

driev:

This PR adds a TTL to deleted pipeline and experiment DB items. The store will continue to grow unless deleted resources are cleaned up. Also, on restarts, the event hub needlessly processes the entire history of pipelines and experiments.

The state of the in-memory store (which is a map) will not be in sync with the DB, until the cleanup goroutine run, but this shouldn't cause any issues.

Disk space should be reclaimed whenever Badger runs GC and removes expired items (this is acutally the client's resposibility). We don't run GC explicitly, but it seems to occur whenever the DB is restored.

Fixes #

INFRA-901

Please see - https://github.com/SeldonIO/seldon-core/compare/v2...driev:seldon-core:INFRA-901/scheduler-store-terminated-resource-cleanup?expand=1#diff-dd478884b4e1e3881b7a1e581ba1fe8e1db1e840b47f616a6eeac2674c342892R384

### DIFF
--- a/scheduler/pkg/store/experiment/db.go
+++ b/scheduler/pkg/store/experiment/db.go
@@ -62,22 +62,23 @@ func (edb *ExperimentDBManager) Stop() error {
 	return utils.Stop(edb.db)
 }
 
-// a nil ttl will save the experiment indefinitely
-func (edb *ExperimentDBManager) save(experiment *Experiment, ttl *time.Duration) error {
+func (edb *ExperimentDBManager) save(experiment *Experiment) error {
 	experimentProto := CreateExperimentSnapshotProto(experiment)
 	experimentBytes, err := proto.Marshal(experimentProto)
 	if err != nil {
 		return err
 	}
 
-	if ttl == nil {
+	if !experiment.Deleted {
 		return edb.db.Update(func(txn *badger.Txn) error {
 			err = txn.Set([]byte(experiment.Name), experimentBytes)
 			return err
 		})
 	} else {
+		// useful for testing
+		ttl := time.Until(experiment.DeletedAt.Add(utils.DeletedResourceTTL))
 		return edb.db.Update(func(txn *badger.Txn) error {
-			e := badger.NewEntry([]byte(experiment.Name), experimentBytes).WithTTL(*ttl)
+			e := badger.NewEntry([]byte(experiment.Name), experimentBytes).WithTTL(ttl)
 			err = txn.SetEntry(e)
 			return err
 		})
@@ -120,14 +121,8 @@ func (edb *ExperimentDBManager) restore(
 					return err
 				}
 				experiment := CreateExperimentFromSnapshot(&snapshot)
-				if experiment.Deleted && item.ExpiresAt() == 0 {
-					ttl := deletedExperimentTTL
-					err = edb.save(experiment, &ttl)
-					if err != nil {
-						edb.logger.WithError(err).Warnf("failed to set ttl for experiment %s", experiment.Name)
-					}
-				}
 				if experiment.Deleted {
+					experiment.DeletedAt = utils.GetDeletedAt(item)
 					err = stopExperimentCb(experiment)
 				} else {
 					// otherwise attempt to start the experiment
--- a/scheduler/pkg/store/experiment/db.go
+++ b/scheduler/pkg/store/experiment/db.go
@@ -10,6 +10,8 @@ the Change License after the Change Date as each is defined in accordance with t
 package experiment
 
 import (
+	"time"
+
 	"github.com/dgraph-io/badger/v3"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/proto"
@@ -60,16 +62,26 @@ func (edb *ExperimentDBManager) Stop() error {
 	return utils.Stop(edb.db)
 }
 
-func (edb *ExperimentDBManager) save(experiment *Experiment) error {
+// a nil ttl will save the experiment indefinitely
+func (edb *ExperimentDBManager) save(experiment *Experiment, ttl *time.Duration) error {
 	experimentProto := CreateExperimentSnapshotProto(experiment)
 	experimentBytes, err := proto.Marshal(experimentProto)
 	if err != nil {
 		return err
 	}
-	return edb.db.Update(func(txn *badger.Txn) error {
-		err = txn.Set([]byte(experiment.Name), experimentBytes)
-		return err
-	})
+
+	if ttl == nil {
+		return edb.db.Update(func(txn *badger.Txn) error {
+			err = txn.Set([]byte(experiment.Name), experimentBytes)
+			return err
+		})
+	} else {
+		return edb.db.Update(func(txn *badger.Txn) error {
+			e := badger.NewEntry([]byte(experiment.Name), experimentBytes).WithTTL(*ttl)
+			err = txn.SetEntry(e)
+			return err
+		})
+	}
 }
 
 func (edb *ExperimentDBManager) saveVersion() error {
@@ -88,7 +100,8 @@ func (edb *ExperimentDBManager) delete(name string) error {
 }
 
 func (edb *ExperimentDBManager) restore(
-	startExperimentCb func(*Experiment) error, stopExperimentCb func(*Experiment) error) error {
+	startExperimentCb func(*Experiment) error, stopExperimentCb func(*Experiment) error,
+) error {
 	return edb.db.View(func(txn *badger.Txn) error {
 		opts := badger.DefaultIteratorOptions
 		it := txn.NewIterator(opts)
@@ -107,6 +120,13 @@ func (edb *ExperimentDBManager) restore(
 					return err
 				}
 				experiment := CreateExperimentFromSnapshot(&snapshot)
+				if experiment.Deleted && item.ExpiresAt() == 0 {
+					ttl := deletedExperimentTTL
+					err = edb.save(experiment, &ttl)
+					if err != nil {
+						edb.logger.WithError(err).Warnf("failed to set ttl for experiment %s", experiment.Name)
+					}
+				}
 				if experiment.Deleted {
 					err = stopExperimentCb(experiment)
 				} else {
--- a/scheduler/pkg/store/experiment/db.go
+++ b/scheduler/pkg/store/experiment/db.go
@@ -10,8 +10,6 @@ the Change License after the Change Date as each is defined in accordance with t
 package experiment
 
 import (
-	"time"
-
 	"github.com/dgraph-io/badger/v3"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/proto"
@@ -75,8 +73,7 @@ func (edb *ExperimentDBManager) save(experiment *Experiment) error {
 			return err
 		})
 	} else {
-		// useful for testing
-		ttl := time.Until(experiment.DeletedAt.Add(utils.DeletedResourceTTL))
+		ttl := utils.DeletedResourceTTL
 		return edb.db.Update(func(txn *badger.Txn) error {
 			e := badger.NewEntry([]byte(experiment.Name), experimentBytes).WithTTL(ttl)
 			err = txn.SetEntry(e)
--- a/scheduler/pkg/store/experiment/db_test.go
+++ b/scheduler/pkg/store/experiment/db_test.go
@@ -324,84 +324,64 @@ func TestSaveAndRestore(t *testing.T) {
 func TestSaveAndRestoreDeletedExperiments(t *testing.T) {
 	g := NewGomegaWithT(t)
 	type test struct {
-		name        string
-		experiments []*Experiment
-		withTTL     bool
+		name       string
+		experiment Experiment
+		withTTL    bool
 	}
 
 	getStrPtr := func(val string) *string { return &val }
 
+	createDeletedExperiment := func(name string) Experiment {
+		return Experiment{
+			Name:         name,
+			ResourceType: ModelResourceType,
+			Default:      getStrPtr("model1"),
+			Candidates: []*Candidate{
+				{
+					Name:   "model1",
+					Weight: 50,
+				},
+				{
+					Name:   "model2",
+					Weight: 50,
+				},
+			},
+			KubernetesMeta: &KubernetesMeta{
+				Namespace:  "default",
+				Generation: 2,
+			},
+			Deleted: true,
+		}
+	}
+
 	tests := []test{
 		{
-			name: "deleted experiment with ttl",
-			experiments: []*Experiment{
-				{
-					Name:         "test-with",
-					ResourceType: ModelResourceType,
-					Default:      getStrPtr("model1"),
-					Candidates: []*Candidate{
-						{
-							Name:   "model1",
-							Weight: 50,
-						},
-						{
-							Name:   "model2",
-							Weight: 50,
-						},
-					},
-					KubernetesMeta: &KubernetesMeta{
-						Namespace:  "default",
-						Generation: 2,
-					},
-					Deleted: true,
-				},
-			},
-			withTTL: true,
+			name:       "deleted experiment with ttl",
+			experiment: createDeletedExperiment("with-ttl"),
+			withTTL:    true,
 		},
 		{
-			name: "deleted experiment without ttl",
-			experiments: []*Experiment{
-				{
-					Name:         "test-without",
-					ResourceType: ModelResourceType,
-					Default:      getStrPtr("model1"),
-					Candidates: []*Candidate{
-						{
-							Name:   "model1",
-							Weight: 50,
-						},
-						{
-							Name:   "model2",
-							Weight: 50,
-						},
-					},
-					KubernetesMeta: &KubernetesMeta{
-						Namespace:  "default",
-						Generation: 2,
-					},
-					Deleted: true,
-				},
-			},
-			withTTL: false,
+			name:       "deleted experiment without ttl",
+			experiment: createDeletedExperiment("without-ttl"),
+			withTTL:    false,
 		},
 	}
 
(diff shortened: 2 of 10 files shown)


scheduler/pkg/store/experiment/store.go
```diff
@@ -381,6 +381,7 @@ func (es *ExperimentStore) StopExperiment(experimentName string) error {
 			es.eventHub.PublishModelEvent(experimentStopEventSource, *modelEvt)
 		}
 		if pipelineEvt != nil {
+			// TODO: is this a bug?
```
lc525, line 384: It is a bug... waiting to happen. At the moment we don't actually propagate the event source to the event handlers, because we don't take different actions depending on where the event is coming from. The only exception is for Servers, with the changes recently introduced [here](https://github.com/sakoush/seldon-core/blob/2f66806436e4ee82b81149824a20f660ad3464bc/scheduler/pkg/coordinator/server.go#L67)

As a consequence, setting the wrong source here has no effect right now, but we should update it nonetheless.

driev: Thanks for the context... updated it.

scheduler/pkg/store/experiment/db_test.go
```diff
@@ -268,6 +321,112 @@ func TestSaveAndRestore(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			path := fmt.Sprintf("%s/db", t.TempDir())
+			logger := log.New()
+			edb, err := newExperimentDbManager(getExperimentDbFolder(path), logger)
+			g.Expect(err).To(BeNil())
+			for _, exp := range test.experiments {
```
lc525, line 395: Not a strong opinion either way, but a thought: Can we skip any experiments that have `Deleted: false` here? The assertions wouldn't hold for them. I know the test name in itself should prevent someone from adding such a test case but....

lc525 (approved):
I think this is good to merge. Ideally, we would also have a periodic cleanup for the in-memory store for deleted things, but I think it could go as a separate PR. This is because we should generally assume the scheduler to be long-running. Thoughts?

scheduler/pkg/store/experiment/db_test.go
driev: I've added an expectation that they're deleted instead.

driev:
> I think this is good to merge. Ideally, we would also have a periodic cleanup for the in-memory store for deleted things, but I think it could go as a separate PR. This is because we should generally assume the scheduler to be long-running. Thoughts?

There should be some cleanup of those stores, too. And I agree that it should go in a separate PR - a new field that stores the time of deletion, would make implementing this easy. The main purpose of this one was to alleviate the load on the event hub on restarts (I'll add this to the description).

scheduler/pkg/store/pipeline/db.go
```diff
@@ -109,6 +122,14 @@ func (pdb *PipelineDBManager) restore(createPipelineCb func(pipeline *Pipeline))
 				if err != nil {
 					return err
 				}
+
+				if pipeline.Deleted && item.ExpiresAt() == 0 {
+					ttl := time.Duration(time.Hour * 24)
```
sakoush, line 127: Should this be set as a constant similar to experiments?

scheduler/pkg/store/experiment/db.go
```diff
@@ -107,6 +120,13 @@ func (edb *ExperimentDBManager) restore(
 					return err
 				}
 				experiment := CreateExperimentFromSnapshot(&snapshot)
+				if experiment.Deleted && item.ExpiresAt() == 0 {
+					ttl := deletedExperimentTTL
+					err = edb.save(experiment, &ttl)
```
sakoush, line 125: Ideally this logic of adding ttl for deleted resources could have been done as part of a migration process to a new version of the db or at least as part of the construction of our db object as a migration is not strictly required here.

Having it here (while logically correct) is not making the code legible as it is mixing the concern of restore (the purpose of this function) with the extra logic of adding ttl in case it doesnt exist.

Happy to discuss views here.

sakoush: Also we can perhaps utilise the timestamps on these records and for deleted resources more than 24 hours ago we can remove them, which will allow users to get the benefit from this change upon upgrade to the new version (as opposed to the next restart of the scheduler).

lc525: To clarify some elements in case Niall goes looking for migration logic. There is no logic for migration, just the code enabling that (a previous PR from Sherif has introduced versions that are stored in the DB and could be checked to implement custom logic) -- atm the migration from "v1" to "v2" mostly consists of setting that version in the DB if it doesn't exist.

I briefly pondered along the same lines as Sherif while reviewing the PR, and I think he makes a good point here. My take was that we're not really changing the db format, just the way in which we're using it (with TTLs) so it's not a "migration" per se, just a new feature. However, I now agree it would be cleaner to separate concerns and have this as part of migration (actual migration code would exist in `new{Pipeline,Experiment}DBManager`).

lc525: With respect to record timestamps, elements in the db have a `item.Version` key which is supposed to hold the timestamp of the transaction where the item gets committed. However, I don't know if there are guarantees about that timestamp being related to wall-clock time -- it might be now but that may also change in the future (something to check).

If it exists and it can be fetched reliably (i.e is already being set automatically by the transactions), it would be quite nice to have this cleanup as we load old databases in the new version.

sakoush: > I think we can do this even without changing current{Pipeline,Experiment}snapshotVersion, by calling adding a separate function that adds TTLs and calling that from new{Pipeline,Experiment}DBManager. However, it would cost us an additional iteration of the db (rather than piggy-backing on the one we already do in restore)

I think my worry is that this logic of dealing with ttl if not done as a part of a proper migration is going to stick with us indefinitely in `restore` or in `new{Pipeline,Experiment}DBManager`, and this might be problematic in the future:
- if done in `restore` we will always have to keep this piggy-backing logic, which can restrict the way we might want to do restore in the future.
- if done in `new{Pipeline,Experiment}DBManager` we will always have to read the db once on a new scheduler even if it is not required.

However if the db has an new explicit version 3 (meaning that we have dealt with ttl). Then we dont have to do anything and it is cleaner.

driev: I've seen the DB migration logic and my thinking was similar to @lc525, in that it's not a schema migration, but rather some backfilling. I was thinking that this could be the first step in implementing proper logic for the deletion of resources, which would also take into account the in-memory store. I think the benefit of having the logic outside of the migration step, is that it will run on every restart, as opposed to once. If resources are not updated again after they have been deleted, this logic will be redundant, so a once only migration should suffice. If this logic does become redundant, as expected, it can be removed. Also the current restore logic is convulated anyway, in that it's restoring the DB and filling the in-memory store.

Regarding the `item.Version` timestamps, I think it would be much better to explicitly set some field on the resource itself, for example `DeletedAt`. The time a resource was deleted at will be useful for managing the in-memory store too.

sakoush: >  If this logic does become redundant, as expected, it can be removed

My point is in fact that we dont know when this logic has became redundant as it depends on the state of the db, which is not well defined unless we update the version record in the db to a new one.

As I appreciate it is not a schema change then perhaps a full migration is not strictly required. I still think that we should increase the version in the db to 3 (to know that we dont have any long lived deleted resources anymore) and if we decide to keep the logic in `restore` then we can first check whether the version is < 3 (to make it explicit).

I suggest also to have a quick think about how this will interact with the in-memory as it might affect the implementation here. However as we have progressed as far, this can happen in a follow-up work as suggested in one of the earlier comments.

lc525: @sakoush Indeed, increasing the version would give us the benefit of only running this when needed. I also agree now that we should do that. 

At the moment, `migrateToDBCurrentVersion` is only called when there's no version, and a comment suggests that further migration logic would still be triggered from `new{Pipeline,Experiment}DBManager` though. So perhaps some small refactoring is needed for the migration logic.

@driev I agree that longer term we want that `DeletedAt` field. However, what Sherif was referring to was that on this particular migration (from "v2" to "v3"), what will happen is that the old database will get loaded, and the deleted experiments will be re-added to the in-memory store (via the callbacks in `restore`), irrespective of whether they were already deleted for a long time or not. So the point here is whether, even for this migration, we could already detect that a deleted experiment loaded from the old database is quite old, and no longer add it to the in-memory store.

The question is whether we could do that based on `item.Version`.

driev: > The question is whether we could do that based on `item.Version`. Otherwise, we give the customers v2.8.4, their in-memory stores remain just as big as they are now, and only on the first scheduler restart they will get the cleanup (if TTL expired)

Yeah, that would definitely work.

lc525: @driev Sorry to pile this up in this PR's comments, but if we started discussing...

There is also the issue of rollback, which should be tested when we make changes to the persistent store. 

Say, a customer has deployed our next release, new scheduler comes up and updates the database (perhaps in this imaginary world we did make some changes to the format, like adding the "DeletedAt" field or perhaps other types of changes). 

Now, something goes terribly wrong and a significant fault is discovered in that Core 2 release. 

Customer wants to roll back. It should be possible to do so (use the new database with the old scheduler code). This should "just work" because we're serialising protobufs in the db, but I'm just adding this here so that we're on the lookout for updates to our code which might break this.

driev: As mentioned offline - `item.Version` isn't useful. Proper migration logic is necessary at some point, but I don't think it's required here, as the schema isn't changing. I've added some "cleanup" go routines, to keep the in-memory stores in sync with the DB (roughly).

scheduler/pkg/store/experiment/store.go
```diff
@@ -461,3 +471,22 @@ func (es *ExperimentStore) GetExperiments() ([]*Experiment, error) {
 	}
 	return foundExperiments, nil
 }
+
+func (es *ExperimentStore) cleanupDeletedExperiments() {
+	es.logger.Info("cleaning up deleted experiments")
+	for _, experiment := range es.experiments {
+		if experiment.Deleted {
+			es.mu.Lock()
+			defer es.mu.Unlock()
```
sakoush, line 480: I think the lock should be done once before the for loop? otherwise using `defer` here is potentially going to cause a deadlock.

scheduler/pkg/store/pipeline/store.go
```diff
@@ -426,3 +445,22 @@ func (ps *PipelineStore) handleModelEvents(event coordinator.ModelEventMsg) {
 		}
 	}()
 }
+
+func (ps *PipelineStore) cleanupDeletedPipelines() {
+	ps.logger.Info("cleaning up deleted pipelines")
+	for _, pipeline := range ps.pipelines {
+		if pipeline.Deleted {
+			ps.mu.Lock()
+			defer ps.mu.Unlock()
```
sakoush, line 454: ditto

scheduler/pkg/store/pipeline/store.go
```diff
@@ -204,6 +208,20 @@ func (ps *PipelineStore) addPipelineImpl(req *scheduler.Pipeline) (*coordinator.
 	}, nil
 }
 
+func (ps *PipelineStore) generationMatches(req *scheduler.Pipeline, lastPipeline *PipelineVersion) bool {
```
sakoush, line 211: thanks for creating this helper.

scheduler/pkg/store/pipeline/store.go
```diff
@@ -97,6 +99,15 @@ func (ps *PipelineStore) InitialiseOrRestoreDB(path string) error {
 	if err != nil {
 		return err
 	}
+
+	go func() {
+		ticker := time.NewTicker(utils.DeletedResourceCleanupFrequency)
+		defer ticker.Stop()
```
sakoush, line 105: the for loop will run indefinitely and defer will not be called in practice.  Could consider a `done` channel pattern as well, that can be linked to the process exit.  Otherwise just remove?

scheduler/pkg/store/experiment/store.go
```diff
@@ -120,10 +122,18 @@ func (es *ExperimentStore) InitialiseOrRestoreDB(path string) error {
...
 	es.db = db
 	// If database already existed we can restore else this is a noop
-	err = es.db.restore(es.StartExperiment, es.AddExperimentInMap)
+	err = es.db.restore(es.StartExperiment, es.addExperimentInMap)
 	if err != nil {
 		return err
 	}
+	go func() {
+		ticker := time.NewTicker(utils.DeletedResourceCleanupFrequency)
+		defer ticker.Stop()
```
sakoush, line 131: ditto

scheduler/pkg/store/experiment/store.go
```diff
@@ -321,7 +331,7 @@ func (es *ExperimentStore) StartExperiment(experiment *Experiment) error {
 	}
 	if es.eventHub != nil {
 		if modelEvt != nil {
-			es.eventHub.PublishModelEvent(experimentStateEventSource, *modelEvt)
+			es.eventHub.PublishModelEvent(experimentStartEventSource, *modelEvt)
```
sakoush, line 334: nice catch!

scheduler/pkg/store/experiment/db.go
```diff
@@ -66,10 +66,20 @@ func (edb *ExperimentDBManager) save(experiment *Experiment) error {
...
-	})
+
+	if !experiment.Deleted {
+		return edb.db.Update(func(txn *badger.Txn) error {
+			err = txn.Set([]byte(experiment.Name), experimentBytes)
+			return err
+		})
+	} else {
+		ttl := utils.DeletedResourceTTL
+		return edb.db.Update(func(txn *badger.Txn) error {
```
sakoush, line 77: nit: given that we have a cron now to recycle delete resources from the store we could just do the same thing by just deleting the record using the `delete` helper? therefore it is less changes and feels the logic of recycling resources is contained in one place / logic.

scheduler/pkg/store/experiment/store.go
sakoush: as mentioned above I think that this logic can also delete from badger db and not rely on ttl as well. just to make it simpler?

sakoush (commented):
I think the proposed change is neat, I left some comments though and especially I am not sure if strictly speaking we need to rely on badger ttl anymore given that we have cleanup crons now that can also delete from the db (we already have `delete` helpers). In this case it is less code/logic that we need to maintain.

scheduler/pkg/store/experiment/db.go
driev: It could be done like that, but the date and time of deletiong needs to be persisted somehow. Some sort of longer term refactor of this portion of the code would be ideal though, for example, the in-memory store could be merged with the DB layer and become an implementation detail there. I think this would reduce the coupling between the functions and keep everything contained.

sakoush: ok got it, so currently we rely on the `ExpiresAt()` api that already exists as opposed to persisting a new field `DeletedAt`.

scheduler/pkg/store/pipeline/store.go
driev: good spot

scheduler/pkg/store/pipeline/store.go
driev: I removed it - probably no need to shut it down.

scheduler/pkg/store/experiment/db.go
driev: Yep... maybe as part of the longer term refactoring, a new field could be persisted. I've got a few different versions of this change on different branches (one with a DB migration, another with the in-memory store and DB merged). I think this would benefit from a slight redesign, but some alignment on the requirements is a prerequisite for that.

scheduler/pkg/store/experiment/store.go
```diff
@@ -461,3 +470,24 @@ func (es *ExperimentStore) GetExperiments() ([]*Experiment, error) {
...
+			if experiment.DeletedAt.IsZero() {
+				experiment.DeletedAt = time.Now()
+				if es.db != nil {
+					err := es.db.save(experiment)
+					if err != nil {
+						es.logger.Warnf("could not update DB TTL for experiment: %s", experiment.Name)
+					}
+				}
+			} else if experiment.DeletedAt.Add(utils.DeletedResourceTTL).Before(time.Now()) {
+				es.experiments[experiment.Name] = nil
```
sakoush, line 489: should the style be `delete(es.experiments, experiment.Name)`?

scheduler/pkg/store/pipeline/store.go
```diff
@@ -426,3 +444,24 @@ func (ps *PipelineStore) handleModelEvents(event coordinator.ModelEventMsg) {
...
+			if pipeline.DeletedAt.IsZero() {
+				pipeline.DeletedAt = time.Now()
+				if ps.db != nil {
+					err := ps.db.save(pipeline)
+					if err != nil {
+						ps.logger.Warnf("could not update DB TTL for pipeline: %s", pipeline.Name)
+					}
+				}
+			} else if pipeline.DeletedAt.Add(utils.DeletedResourceTTL).Before(time.Now()) {
+				ps.pipelines[pipeline.Name] = nil
```
sakoush, line 463: ditto (use `delete`)?

sakoush (approved):
LGTM. left a minor style question.

scheduler/pkg/store/experiment/store.go
driev: yep... updated!

PR #5949, opened 2024-10-02
fix(k8s/yaml): add timing environment variables to triton agent

lc525:
In #5889, we've missed updating the agent for triton in the published
components.yaml

Running `make create` in `[seldon-core]/k8s/` has surfaced the diff.


k8s/yaml/components.yaml
lc525, line 1: generated file

driev (approved):

PR #5964, opened 2024-10-08
refactor: error checks and nil pointer check

abhimanyu003:

* Removes not-required error checks.
* Nil pointer update for switch case.
* Test cases update

### DIFF
--- a/operator/apis/mlops/v1alpha1/experiment_types.go
+++ b/operator/apis/mlops/v1alpha1/experiment_types.go
@@ -154,7 +154,7 @@ func (es *ExperimentStatus) IsConditionReady(t apis.ConditionType) bool {
 func (es *ExperimentStatus) SetCondition(conditionType apis.ConditionType, condition *apis.Condition) {
 	switch {
 	case condition == nil:
-		experimentConditionSet.Manage(es).MarkUnknown(conditionType, "", "")
+		return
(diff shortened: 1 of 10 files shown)


operator/apis/mlops/v1alpha1/experiment_types.go
```diff
@@ -154,7 +154,7 @@ func (es *ExperimentStatus) IsConditionReady(t apis.ConditionType) bool {
 func (es *ExperimentStatus) SetCondition(conditionType apis.ConditionType, condition *apis.Condition) {
 	switch {
 	case condition == nil:
-		experimentConditionSet.Manage(es).MarkUnknown(conditionType, "", "")
+		return
```
sakoush, line 157: could you look for all/similar occurrences for this `SetCondition` helper and follow the same pattern you have suggested here?

sakoush (commented):
LGTM. I think that there are still other occurrences where this  might happen and need fixing, so perhaps search for `SetCondition` to get all cases.

operator/apis/mlops/v1alpha1/experiment_types.go
abhimanyu003: I have updated all the cases to have similar pattern  for `SetCondition`

sakoush (approved):
LGTM

PR #5967, opened 2024-10-09
fix(scheduler): Increase drain grace period wait

sakoush:
This PR increases the drain grace period to 2s inline with other changes for grace periods. This is to allow envoy to settle down after changes.


lc525 (approved):
lgtm

PR #5969, opened 2024-10-10
fix(agent): Skip agent out of order control messages

sakoush:

Currently the agent, due to the way it processes control messages from the scheduler, can see out of order messages. For example consider the following flow
```
- Model A unload request
- Model A unload request - sleep
- Model A load request proceeds and lets scheduler model is loaded
- Model A unload request (out of order) proceeds and scheduler is confused
- Model A is unloaded while route to it exists in envoy
```

In this PR we introduce a way to detect out-of-order messages and we skip them. We record the milli ticks for successful operations per model and if an incoming operation has a tick that is earlier than whats recorded, we skip it.

Fixes #
INFRA-1204 (internal)

This PR also bumps up server drain grace period to 3s, in a future PR this will be configuratble.

### DIFF
--- a/scheduler/pkg/agent/client.go
+++ b/scheduler/pkg/agent/client.go
@@ -56,6 +56,7 @@ type Client struct {
 	modelScalingClientStream agent.AgentService_ModelScalingTriggerClient
 	settings                 *ClientSettings
 	modelTimestamps          sync.Map
+	startTime                time.Time
 	ClientServices
 	SchedulerGrpcClientOptions
 	KubernetesOptions
@@ -195,6 +196,7 @@ func NewClient(
 		stop:            atomic.Bool{},
 		settings:        settings,
 		modelTimestamps: sync.Map{},
+		startTime:       time.Now(),
 	}
 }
 
@@ -477,13 +479,13 @@ func (c *Client) StartService() error {
 
 		c.logger.Infof("Received operation")
 
-		timestamp := time.Now().UnixMilli()
+		ticksSinceStart := time.Since(c.startTime).Milliseconds()
 		switch operation.Operation {
 		case agent.ModelOperationMessage_LOAD_MODEL:
 			c.logger.Infof("calling load model")
 
 			go func() {
-				err := c.LoadModel(operation, timestamp)
+				err := c.LoadModel(operation, ticksSinceStart)
 				if err != nil {
 					c.logger.WithError(err).Errorf(
 						"Failed to handle load model %s:%d",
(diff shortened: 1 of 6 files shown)


scheduler/pkg/agent/client.go
```diff
@@ -474,12 +477,13 @@ func (c *Client) StartService() error {
 
 		c.logger.Infof("Received operation")
 
+		timestamp := time.Now().UnixMilli()
```
lc525, line 480: Something to discuss: UnixMilli will return a wall-clock time, which is not monotonic. In particular, it goes back when NTP adjusts the machine's time. So we might see some reordering/out-of-order messages because of that, when the actual events were not out-of-order.

The typical NTP adjustements can get into the 100 - 200 ms range (in principle NTP still works/can make adjustements for large drifts in the tens of second range, but I think those are unlikely to occur frequently)

scheduler/pkg/agent/client_test.go
```diff
@@ -393,7 +393,7 @@ func TestLoadModel(t *testing.T) {
 			time.Sleep(50 * time.Millisecond)
 
 			// Do the actual function call that is being tested
-			err := client.LoadModel(test.op)
+			err := client.LoadModel(test.op, time.Now().Unix())
```
lc525, line 396: Probably doesn't matter for the test, but for the test to match the actual code, could we also use `time.Now().UnixMilli()`? Or was there another reason why the LoadModel is passed seconds in the test (here and below) rather than milliseconds?

lc525: If it was just because in the test we don't need that fine of a time resolution, perhaps that's just fine, but wanted to check

sakoush: yes in the tests we can even have handcrafted ts and not necessarily unix ticks. I will actually do that to make it clear that it doesnt have to be unix ticks.

sakoush: done in https://github.com/SeldonIO/seldon-core/pull/5969/commits/c90bc4a855ba1058d32fbb2298a133e00e3aaca2

scheduler/pkg/agent/client.go
sakoush: good point. I can just use a counter internally instead of `UnixMilli`

sakoush: In fact `time.Now` contains a monotonic clock, will use that instead.

sakoush: fixed in [9265c30](https://github.com/SeldonIO/seldon-core/pull/5969/commits/9265c30632b83b8eeabb8ea0d2b7f367027d66bb)

lc525 (approved):
lgtm

PR #5977, opened 2024-10-14
feat(k6): Add k6 controlplane with dataplane runs

sakoush:
This PR introduces model replicas changes as inference go through to understand the behaviour of the system under such scenarios.

The scenarios that are enabled are:
- `infer_constant_rate`
- `infer_constant_vu`

To trigger model replica changes, set `ENABLE_MODEL_REPLICA_CHANGE`. The number of replicas is chosen uniformly between 1 and `MAX_MODEL_REPLICAS`. 

We also support sleeping after each model replica change by setting `SLEEP_BETWEEN_REPLICA_CHANGE`.

### DIFF
--- a/tests/k6/scenarios/infer_constant_rate.js
+++ b/tests/k6/scenarios/infer_constant_rate.js
@@ -1,9 +1,6 @@
 import { getConfig } from '../components/settings.js'
-import { doInfer, setupBase, teardownBase, getVersionSuffix } from '../components/utils.js'
-import { generateModel } from '../components/model.js';
-import * as k8s from '../components/k8s.js';
+import { doInfer, setupBase, teardownBase, getVersionSuffix, applyModelReplicaChange } from '../components/utils.js'
 import { vu } from 'k6/execution';
-import { sleep } from 'k6';
 
 var kubeClient = null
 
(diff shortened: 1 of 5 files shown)


tests/k6/scenarios/infer_constant_rate.js
```diff
@@ -47,6 +53,16 @@ export default function (config) {
     } else {
         doInfer(modelName, modelNameWithVersion, config, false, idx)
     }
+
+    // for simplicity we only change model replicas in the first VU
```
lc525, line 57: Could we move this into a function in `utils.js` (receiving config as an argument) and call it from both places? I'll probably want to reuse this piece of code in other scenarios as well.

I know in general one might one to customise this on their own load scenario (so reusing the common code will not be feasible all the time), but it would be good to have something to call for a default replica change feature that reads the existing config environment variables.

lc525 (approved):
lgtm; just a small comment on extracting the common functionality if you also think it's reusable.

tests/k6/scenarios/infer_constant_rate.js
sakoush: fixed in [c3d1c77](https://github.com/SeldonIO/seldon-core/pull/5977/commits/c3d1c7732f4e11890f26368cbceaae8a8192e976)

PR #5978, opened 2024-10-18
fix: Use keep alive grpc settings on client side

sakoush:
This change introduces keepalive pings from the client side on grpc streams, otherwise in cases of short network partitions the stream might be broken and not re-established.

with keepalive settings we get this error
```
level=error msg="event recv failed" Name=Client error="rpc error: code = Unavailable desc = keepalive ping failed to receive ACK within timeout" func=StartService
```
which will break our loop and tries to reestablish the connection with the other end.

Check here: https://pkg.go.dev/google.golang.org/grpc/keepalive#ClientParameters

### DIFF
--- a/operator/scheduler/client.go
+++ b/operator/scheduler/client.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/status"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


operator/scheduler/client.go
```diff
@@ -34,6 +35,9 @@ const (
 	// these 2 constants in combination with the backoff exponential function will give us a max backoff of 13.5 minutes
 	SchedulerConnectMaxRetries    = 12
 	SchedulerConnectBackoffScalar = 200 * time.Millisecond
+	ClientKeapAliveTime           = 10 * time.Second
```
lc525, line 38: The [docs](https://grpc.io/docs/guides/keepalive/) say: "To avoid DDoSing, it’s important to take caution when setting the keepalive configurations. Thus, it is recommended to avoid enabling keepalive without calls and for clients to avoid configuring their keepalive much below one minute."

I suppose we'd not be ddos-ing ourselves with 10s and just a couple of services -- although it would depend on the number of agents connecting, but we'd probably need to test this value thouroughly before the next release.

lc525 (approved):
Besides the tuning &testing of the actual timeout value, lgtm. Do we want to increase the default just a bit perhaps to be on the safe side?

PR #5987, opened 2024-10-24
fix: Deal with zero servers control plane synchronisation

sakoush:

This change allows fast synchronisation of the scheduler state in the cases we dont have any model server deployed. While this is an edge case and not likely to happen in practice as core 2 cant operate without model servers in the first place, we consider it for sanity reasons.

The change in the code deals with zero servers conditions, covered also by testing.

Fixes INFRA-1217 (internal)

### DIFF
--- a/operator/scheduler/server.go
+++ b/operator/scheduler/server.go
@@ -25,9 +25,6 @@ import (
 
 func (s *SchedulerClient) ServerNotify(ctx context.Context, grpcClient scheduler.SchedulerClient, servers []v1alpha1.Server, isFirstSync bool) error {
 	logger := s.logger.WithName("NotifyServer")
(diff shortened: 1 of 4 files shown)


operator/scheduler/server.go
```diff
@@ -25,9 +25,6 @@ import (
 
 func (s *SchedulerClient) ServerNotify(ctx context.Context, grpcClient scheduler.SchedulerClient, servers []v1alpha1.Server, isFirstSync bool) error {
 	logger := s.logger.WithName("NotifyServer")
-	if len(servers) == 0 {
```
sakoush, line 28: allows control messages to be created for empty server list

driev (approved):

PR #5990, opened 2024-10-24
fix: Upgrade go 1.22

abhimanyu003:

This PR upgrades to 
* go1.22
* Upgrade github action setup-go to v5.0.1 using sha `0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32` because tag can be changed but she can't
* upgrade controller-gen to v0.16.2 from 0.9.2 ( version 0.9.2 causing panic on go1.22 )

Please note that 

Fixes #INFRA-1151


.github/workflows/draft-release.yml
```diff
@@ -69,9 +69,9 @@ jobs:
           node-version: 14
 
       - name: Setup Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.1
```
lc525, line 72: Could we pin to @v5 instead of the actual hash? The setup-go docs seem to reference the major version, so I would assume they would want to keep things mostly compatible within the same major version (so we wouldn't get issues with the action being updated under us, but instead we could benefit from things like security updates)

tests/integration/go.mod
```diff
@@ -1,77 +1,77 @@
 module github.com/seldonio/seldon-core/tests/integration/v2
 
-go 1.21
+go 1.22.0
```
lc525, line 3:
```suggestion
go 1.22
```

tests/integration/go.mod
```diff
@@ -1,77 +1,77 @@
...
 
 require (
-	github.com/docker/docker v24.0.7+incompatible
-	github.com/onsi/gomega v1.32.0
+	github.com/docker/docker v26.1.0+incompatible
```
lc525, line 6: are those package updates needed for go 1.22 support?

tests/integration/go.mod
abhimanyu003: Thanks for noticing this is updated

lc525 (approved):
lgtm, really good to be on golang 1.22

abhimanyu003:
Tested is using smoke test with docker `./smoke-tests.sh 5 seldon` thing are looking good.

PR #5992, opened 2024-10-25
always send model events for deleted models

driev:

Some deleted models can end up in a state where model status events are not sent when the model is deleted. This leads to the deleted model getting stuck and it's never removed by the operator. 

The change is to always send model update events for deleted models.

Internal : INFRA-1118

Fixes #

The test case `DeletedModelNoReplicas` covers this change and replicates the state of the model referenced in the logs attached to the ticket.

### DIFF
--- a/scheduler/pkg/store/memory.go
+++ b/scheduler/pkg/store/memory.go
@@ -311,10 +311,10 @@ func (m *MemoryStore) updateLoadedModelsImpl(
 		return nil, fmt.Errorf("failed to find server %s", serverKey)
 	}
 	for _, replica := range replicas {
-		_, ok := server.replicas[replica.GetReplicaIdx()]
+		_, ok := server.replicas[replica.replicaIdx]
 		if !ok {
 			return nil, fmt.Errorf(
-				"Failed to reserve replica %d as it does not exist on server %s",
+				"failed to reserve replica %d as it does not exist on server %s",
 				replica.GetReplicaIdx(), serverKey,
 			)
 		}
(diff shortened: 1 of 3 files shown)


scheduler/pkg/store/memory.go
```diff
@@ -311,10 +311,10 @@ func (m *MemoryStore) updateLoadedModelsImpl(
 		return nil, fmt.Errorf("failed to find server %s", serverKey)
 	}
 	for _, replica := range replicas {
-		_, ok := server.replicas[replica.GetReplicaIdx()]
+		_, ok := server.replicas[replica.replicaIdx]
```
sakoush, line 314: is there a reason why not to use the public method?

scheduler/pkg/store/memory.go
```diff
@@ -326,67 +326,47 @@ func (m *MemoryStore) updateLoadedModelsImpl(
...
-			logger.Debugf(
-				"model %s on server %s replica %d already loaded",
-				modelKey, serverKey, replica.GetReplicaIdx(),
+				modelKey, modelVersion.version, serverKey, replica.replicaIdx,
 			)
+			modelVersion.SetReplicaState(replica.replicaIdx, LoadRequested, "")
+			m.updateReservedMemory(LoadRequested, serverKey, replica.replicaIdx, modelVersion.GetRequiredMemory())
 		}
-		updatedReplicas[replica.GetReplicaIdx()] = true
+		updatedReplicas[replica.replicaIdx] = true
```
sakoush, line 341: nit: 
I think that we use `updatedReplicas` in a subtle way:
step 1: essentially we do mark the server replicas that we are going to load the model replica  onto
step 2: for any existing replicas `ReplicaState()`  that are not part of `updatedReplicas` we unload the model replica from them.

It is then less legible that also when unloading the model replica we also set `updatedReplicas` (instead of marking a flag `updated`), I think we should document this behaviour in the code so that it is clearer.

scheduler/pkg/store/memory_status.go
```diff
@@ -105,6 +107,8 @@ func updateModelState(isLatest bool, modelVersion *ModelVersion, prevModelVersio
 }
 
 func (m *MemoryStore) FailedScheduling(modelVersion *ModelVersion, reason string, reset bool) {
+	modelVersion.mu.Lock()
+	defer modelVersion.mu.Unlock()
```
sakoush, line 111: release lock before sending to the event hub as otherwise it can cause deadlock

sakoush: caller also `scheduleToServer` is grabbing a lock on the model so it feels safe in this particular instance not to have a lock?

sakoush (commented):
left some comments

scheduler/pkg/store/memory_status.go
driev: Yeah, I'll remove this... it was an early change that was unrelated to the issue.

scheduler/pkg/store/memory.go
driev: Yeah, I can make it more explicit.

scheduler/pkg/store/memory.go
driev: There's mixed usage of  `replicaIdx` as a param, var and via the public method, and I was making that uniform to read.

scheduler/pkg/store/memory.go
```diff
@@ -310,83 +311,65 @@ func (m *MemoryStore) updateLoadedModelsImpl(
...
 			)
 		}
+		assignedReplicaIds[replica.replicaIdx] = struct{}{}
 	}
 
 	if modelVersion.HasServer() && modelVersion.Server() != serverKey {
 		logger.Debugf("Adding new version as server changed to %s from %s", modelVersion.Server(), serverKey)
 		m.addNextModelVersion(model, model.Latest().modelDefn)
-		return m.updateLoadedModelsImpl(modelKey, model.Latest().GetVersion(), serverKey, replicas)
+		modelVersion = model.Latest()
```
driev, line 329: If you follow the recursive call here, all that changes is the `modelVersion`. This makes the method hard to follow.

driev: Oh...  this will also need to be a loop, to ensure the that if a `modelVersion` has a server, it matches the server key... seems unnecessary though.

scheduler/pkg/store/memory.go
driev: `updatedReplicas` was essentially a map of the ids of all the `replicas` and it's populated as part of the loop to update replicas that are not already loading, so I've moved it to the first pass of the replicas to validate them, which I think makes the whole method read much more clearly.

scheduler/pkg/store/memory.go
```diff
@@ -310,83 +311,65 @@ func (m *MemoryStore) updateLoadedModelsImpl(
...
 			logger.Debugf(
 				"Setting model %s version %d on server %s replica %d to LoadRequested",
-				modelKey, modelVersion.version, serverKey, replica.GetReplicaIdx(),
-			)
-			modelVersion.SetReplicaState(replica.GetReplicaIdx(), LoadRequested, "")
-			m.updateReservedMemory(LoadRequested, serverKey, replica.GetReplicaIdx(), modelVersion.GetRequiredMemory())
-			updated = true
-		} else {
-			logger.Debugf(
-				"model %s on server %s replica %d already loaded",
```
sakoush, line 344: I actually think that these debug log messages about the state of the model replica has been useful to understand and inspect the system in case of issues, now we are coalescing everything in  "Model status update not required for model %s server %s as no replicas were updated". I suggest to still log the specific detail why an update is not required.

sakoush (approved):
LGTM. I left some comments, the main one is to try and keep log details for debugging purposes.

scheduler/pkg/store/memory.go
```diff
@@ -310,83 +311,65 @@ func (m *MemoryStore) updateLoadedModelsImpl(
...
+				replicaStateUpdated = true
 			}
 		}
 	}
 
 	// in cases where we did have a previous ScheduleFailed, we need to reflect the change here
 	// this could be in the cases where we are scaling down a model and the new replica count can be all deployed
-	if updated || modelVersion.state.State == ScheduleFailed {
+	// and always send an update for deleted models, so the operator will remove them from k8s
+	if replicaStateUpdated || modelVersion.state.State == ScheduleFailed || model.IsDeleted() {
```
sakoush, line 364: Having another thought about the conditions here, I feel that, while the aim is to be specific about which cases are legit to trigger an update, these conditions are a bit arbitrary. In fact there is an issue that I am currently looking at that is going to add yet another condition to this filter. 

This made me realise that if we are already in this function, there is a high probability that there is a status change that triggered it and therefore it might feel that we should try to update the status of the model and send an event downstream. Have you spotted any cases where we should **not** be sending an event?

scheduler/pkg/store/memory.go
driev: Yeah, fair enough, I'll log the state for each replica.

scheduler/pkg/store/memory.go
```diff
@@ -310,83 +311,65 @@ func (m *MemoryStore) updateLoadedModelsImpl(
...
 
-	// Update model that need to be placed on a replica to request loading
-	updatedReplicas := make(map[int]bool)
-	updated := false
-	for _, replica := range replicas {
-		existingState := modelVersion.replicas[replica.GetReplicaIdx()]
+	//  resevere memory for existing replicas that are not already loading or loaded
+	replicaStateUpdated := false
+	for replicaIdx := range assignedReplicaIds {
+		existingState := modelVersion.replicas[replicaIdx]
```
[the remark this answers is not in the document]
driev: Yeah, good shout... updated it.

PR #6014, opened 2024-10-29
fix: Model gateway silently ignores errors

abhimanyu003:
This PR is adding errors to kafka errorTopic.

To replicate this issue follow: #5125

Fixes #INFRA-1165

### DIFF
--- a/scheduler/pkg/kafka/gateway/worker.go
+++ b/scheduler/pkg/kafka/gateway/worker.go
@@ -287,8 +287,13 @@ func (iw *InferWorker) produce(
 		return err
 	}
 	go func() {
-		<-deliveryChan
+		e := <-deliveryChan
 		span.End()
+		m := e.(*kafka.Message)
+		if m.TopicPartition.Error != nil {
+			iw.logger.WithError(m.TopicPartition.Error).Errorf("Failed to produce event for model %s", topic)
+		}
+		close(deliveryChan)
 	}()
 
 	return nil
--- a/scheduler/pkg/kafka/gateway/worker.go
+++ b/scheduler/pkg/kafka/gateway/worker.go
@@ -312,7 +312,7 @@ func (iw *InferWorker) restRequest(ctx context.Context, job *InferWork, maybeCon
 	}
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, restUrl.String(), bytes.NewBuffer(data))
 	if err != nil {
-		return err
+		return iw.produce(ctx, job, iw.topicNamer.GetModelErrorTopic(), []byte(err.Error()), true, nil)
 	}
 
 	req.Header.Set("Content-Type", "application/json")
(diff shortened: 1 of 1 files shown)


scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -343,14 +343,19 @@ func (iw *InferWorker) restRequest(ctx context.Context, job *InferWork, maybeCon
 		return iw.produce(ctx, job, iw.topicNamer.GetModelErrorTopic(), b, true, nil)
 	}
 
-	return iw.produce(
+	err = iw.produce(
```
lc525, line 346: Just a note that if this produce fails, it's likely that the produce in the error handling will also fail. But let's keep it like this, because there are edge cases (like if someone deleted the output topic) where it's worth exposing that on the error topic.

scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -312,7 +312,7 @@ func (iw *InferWorker) restRequest(ctx context.Context, job *InferWork, maybeCon
 	}
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, restUrl.String(), bytes.NewBuffer(data))
 	if err != nil {
-		return err
+		return iw.produce(ctx, job, iw.topicNamer.GetModelErrorTopic(), []byte(err.Error()), true, nil)
```
lc525, line 315: `iw.produce(...)` returns an error only when `iw.consumer.Produce(...)` returns an error.  The latter calls the Kafka api ([here](https://github.com/SeldonIO/seldon-core/blob/294b5f823fc5784f49157639a3e6e97b2617a460/scheduler/pkg/kafka/gateway/infer.go#L160)). That `kc.producer.Produce(msg, deliveryChan)` is an asynchronous call - it merely enqueues the message in internal kafka client buffers, then returns.

We pass a deliveryChan to it, and we would receive any actual errors on that channel. However, at the moment we're [just ending the trace span when we get an event on the channel](https://github.com/SeldonIO/seldon-core/blob/294b5f823fc5784f49157639a3e6e97b2617a460/scheduler/pkg/kafka/gateway/worker.go#L290). 

I think we should also check there if the returned Kafka event on that channel shows an error or not, and log it if it does.

Something along the lines of:
```
 e := <-deliveryChan
 span.End()
 m := e.(*kafka.Message)

 if m.TopicPartition.Error != nil {
     // log error
 }

 close(deliveryChan)
 ```

lc525 (changes requested):
Added some comments, suggesting to log a wider class of errors here. Ideally we would also catch those errors and write to the error topic.

lc525 (approved):
lgtm

lc525:
Please test with latest changes, and let's aim to merge this by the end of this week

abhimanyu003:
I have tested the new changes they working well. I'm merging this. :)

PR #6016, opened 2024-10-30
fix: Add server keep alive enforcement policy

sakoush:

This PR adds the relevant parameters to make client and server side keep alive work nice together. We should not allow these parameters to be set independently but in some cases they exist in different packages, so manual care is required.

This PR also tidy up some exponential backoff settings.

### DIFF
--- a/scheduler/pkg/agent/modelserver_controlplane/oip/v2.go
+++ b/scheduler/pkg/agent/modelserver_controlplane/oip/v2.go
@@ -52,9 +52,9 @@ func CreateV2GrpcConnection(v2Config V2Config) (*grpc.ClientConn, error) {
 	}
 
 	kacp := keepalive.ClientParameters{
-		Time:                util.ClientKeapAliveTime,
-		Timeout:             util.ClientKeapAliveTimeout,
-		PermitWithoutStream: util.ClientKeapAlivePermit,
+		Time:                util.ClientKeepAliveTime,
+		Timeout:             util.ClientKeepAliveTimeout,
+		PermitWithoutStream: util.ClientKeepAlivePermit,
 	}
 
 	opts := []grpc.DialOption{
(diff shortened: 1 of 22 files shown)


scheduler/pkg/agent/modelserver_controlplane/oip/v2.go
```diff
@@ -51,19 +50,12 @@ func CreateV2GrpcConnection(v2Config V2Config) (*grpc.ClientConn, error) {
 		grpc_retry.WithMax(v2Config.GRPRetryMaxCount),
 	}
 
-	kacp := keepalive.ClientParameters{
```
lc525, line 54: could we add a note on why the keepalive was removed here?

scheduler/pkg/util/grpc.go
```diff
@@ -0,0 +1,40 @@
...
+func GetServerKeepAliveEnforcementPolicy() keepalive.EnforcementPolicy {
+	return keepalive.EnforcementPolicy{
+		MinTime:             gRPCKeepAliveTime,
+		PermitWithoutStream: gRPCKeepAlivePermit,
+	}
+}
+
+func GetClientExponentialBackoff() *backoff.ExponentialBackOff {
+	backOffExp := backoff.NewExponentialBackOff()
+	backOffExp.MaxElapsedTime = 0 // Never stop due to large time between calls
```
lc525, line 36: do we want to move these to constants as well?

scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -124,12 +123,6 @@ func (iw *InferWorker) getGrpcClient(host string, port int) (v2.GRPCInferenceSer
 		creds = insecure.NewCredentials()
 	}
 
-	kacp := keepalive.ClientParameters{
```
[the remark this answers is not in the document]
sakoush: this is a grpc unary call to envoy for to perform inferences, not relevant yet for keepalive settings.

scheduler/pkg/agent/modelserver_controlplane/oip/v2.go
sakoush: agent and model server are on the same pod/network and therefore no need to set keepalive

lc525 (approved):
lgtm, just some minor nits.

scheduler/pkg/util/grpc.go
sakoush: done in [bb639c1](https://github.com/SeldonIO/seldon-core/pull/6016/commits/bb639c1dc845bae35279249628fe8817978e80df)

PR #6017, opened 2024-10-30
fix(modelgateway): Allow modelgateway consumers for transient error

sakoush:

This changes allows modelgateway not to remove the consumer of particular model if it has transitioned to `ScheduleFailed` or `ModelProgressing`. There could be transient failures on the control plane which induce these issues but then data plane should be working just fine.

The consumer will be removed when the user eventually deletes the model.

Fixes #

### DIFF
--- a/scheduler/pkg/kafka/gateway/client.go
+++ b/scheduler/pkg/kafka/gateway/client.go
@@ -159,6 +159,16 @@ func (kc *KafkaSchedulerClient) SubscribeModelEvents() error {
 
 		logger.Infof("Received event name %s version %d state %s", event.ModelName, latestVersionStatus.Version, latestVersionStatus.State.State.String())
 
+		// if the model is in a failed state and the consumer exists then we skip the removal
+		// this is to prevent the consumer from being removed during transient failures of the control plane
+		// in this way data plane can potentially continue to serve requests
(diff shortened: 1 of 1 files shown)


sakoush:
note that this is not yet fixing envoy side of things and we could get some 4xx still for a short period

```
ent-Type: application/json" -d '{"inputs": [{"name": "predict", "shape": [1, 4], "datatype": "FP32", "data": [[1, 2, 3, 4]]}]'
*   Trying 0.0.0.0:9000...
* Connected to 0.0.0.0 (127.0.0.1) port 9000 (#0)
> POST /v2/pipelines/iris-pipeline/infer HTTP/1.1
> Host: 0.0.0.0:9000
> User-Agent: curl/8.1.2
> Accept: */*
> Content-Type: application/json
> Content-Length: 94
> 
< HTTP/1.1 400 Bad Request
< x-forwarded-proto: http
< x-pipeline-version: 1
< x-request-id: iris-pipeline.csh6i9edjf2c73fnp4u0
< date: Wed, 30 Oct 2024 17:13:09 GMT
< content-length: 46
< content-type: text/plain; charset=utf-8
< x-envoy-upstream-service-time: 97
< x-seldon-route: :iris-pipeline.pipeline:
< server: envoy

```

lc525 (approved):
lgtm, let's test this as discussed. I think it doesn't place us in a worse state compared to before

PR #6020, opened 2024-10-31
fix(agent): allow serial order for servers connection per instance

sakoush:

In some cases during rolling upgrades of servers we can get a situation where the scheduler detects the disconnection of an old version of a server _after_ the new server had connected and in this case we dont want to actually act upon this disconnection.

We do that by enforcing a serial order per instance, meaning that the current server has to disconnect before the new one can connect.

There are 2 components affected: agent server and chainer server.

Fixes # INFRA-1226 (internal)

Tested by unit tests

### DIFF
--- a/scheduler/pkg/agent/server.go
+++ b/scheduler/pkg/agent/server.go
@@ -394,7 +394,7 @@ func (s *Server) Subscribe(request *pb.AgentSubscribeRequest, stream pb.AgentSer
 	defer mu.(*sync.Mutex).Unlock()
 
 	logger.Infof("Received subscribe request from %s:%d", request.ServerName, request.ReplicaIdx)
-	
+
 	fin := make(chan bool)
 
 	s.mutex.Lock()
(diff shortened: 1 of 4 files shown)


scheduler/pkg/agent/server.go
```diff
@@ -383,12 +385,20 @@ func (s *Server) ModelScalingTrigger(stream pb.AgentService_ModelScalingTriggerS
 
 func (s *Server) Subscribe(request *pb.AgentSubscribeRequest, stream pb.AgentService_SubscribeServer) error {
 	logger := s.logger.WithField("func", "Subscribe")
+	key := ServerKey{serverName: request.ServerName, replicaIdx: request.ReplicaIdx}
+
+	// this is forcing a serial order per agent (serverName, replicaIdx)
+	// in general this will make sure that a given agent disconnects fully before another agent is allowed to connect
+	mu, _ := s.agentMutex.LoadOrStore(key, &sync.Mutex{})
+	mu.(*sync.Mutex).Lock()
```
lc525, line 393: I think this is probably the cleanest way to do it in the end.

lc525 (approved):
lgtm!

PR #6021, opened 2024-11-01
fix(dataflow): make each replica use unique subscription names

lc525:
Following #6020, it was no longer possible to have multiple replicas of dataflow-engine subscribing simultaneously to the scheduler, because all were connecting with the same subscriber name, and a lock was added per name, first waiting the disconnection of the old subscriber before allowing a new one to progress.

We update the dataflow-engine code so that each replica connects with its own hostname as the subscriber name. If the hostname can not be determined, we subscribe with the name `seldon-dataflow-engine-` followed by a hex string of 50 random characters.

The subscriber name can also be explicitly controlled by passing the `--dataflow-replica-id` argument or the `DATAFLOW_REPLICA_ID` environment variable, wich will take precedence, in that order, to setting the value as the hostname.

- tested in kind, hostname gets correctly picked up:
  - On the dataflow end:
    dataflow-id
  - On the scheduler end: 
    scheduler-dataflow-id

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
@@ -27,6 +27,7 @@ import io.klogging.noCoLogger
 import io.seldon.dataflow.kafka.security.KafkaSaslMechanisms
 import io.seldon.dataflow.kafka.security.KafkaSecurityProtocols
 import java.net.InetAddress
+import java.util.UUID
 
 object Cli {
     private const val ENV_VAR_PREFIX = "SELDON_"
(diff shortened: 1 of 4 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
```diff
@@ -105,10 +109,26 @@ object Cli {
...
+    private fun getSystemConfig(): Configuration {
+        val dataflowIdPair = this.dataflowReplicaId to getDataflowId()
+        return ConfigurationMap(dataflowIdPair)
+    }
+
+    private fun getDataflowId(): String {
+        return try {
+            InetAddress.getLocalHost().hostName
+        } catch (e: Exception) {
+            val hexCharPool: List<Char> = ('a'..'f') + ('0'..'9')
```
sakoush, line 128: nit: is there a package for uuid?

sakoush (approved):
lgtm

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
lc525: When first writing this, I thought that an UUID would be too long for what we need (and slicing out of the UUID provides fewer guarantees due to the constrained format (version/variant fields that are not random).

However, then I went and generated a 50 char hex random string, which is longer than the 32 we would get with a UUID :-). In short, I'll use a UUID v4 here as you propose

PR #6022, opened 2024-11-01
fix(ci): fix flaky test

sakoush:

This PR fixes a flaky test.

Fixes #

### DIFF
--- a/scheduler/pkg/agent/server_test.go
+++ b/scheduler/pkg/agent/server_test.go
@@ -12,6 +12,7 @@ package agent
 import (
 	"context"
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


scheduler/pkg/agent/server_test.go
```diff
@@ -1062,9 +1066,9 @@ func TestSubscribe(t *testing.T) {
 				}(idx, s)
 			}
 
-			time.Sleep(10 * time.Second)
-
-			g.Expect(len(server.agents)).To(Equal(test.expectedAgentsCountAfterClose))
+			for len(server.agents) != test.expectedAgentsCountAfterClose {
```
lc525, line 1069: this will just sleep forever if the test fails, right?

sakoush: yes thats true, I can also cap it to a max count?

lc525 (approved):
lgtm after adding max count in the for loops

scheduler/pkg/agent/server_test.go
sakoush: done in https://github.com/SeldonIO/seldon-core/pull/6022/commits/dfb347230c479062d310d00f1edce4e1df41b6e4

PR #6028, opened 2024-11-05
fix: Reconcile models after network partition

sakoush:

This PR introduces changes required to fix model versions when agent reconnects after 1: network partition and 2: scheduler restart.

The scheduler restart is going to cause a reset of model versions as we dont keep them persisted. This means that on reconnect and from agents that have different views on the model versions we need to reconcile these versions.

The main idea is that if we detect a mismatch of versions we have to induce a new version of the model so that we reconcile the state. This is done as extension to `addModelVersionIfNotExists` which gets triggered when a new agent connects with a set of loaded models.
Note that the workflow will trigger a reschedule after this state which will cause the system to reconcile.

Another change is to mark the agent version when a mismatch happens, this is only used in cases of unloading only and not affecting the rest of the system

A third change is to allow status updates for models that are stuck in `ModelProgressing` state, this is done in  `updateLoadedModelsImpl`

Fixes INFRA-1227 (internal)

This is a subtle change and care should be taken with any implications of it.


sakoush:
superseded by https://github.com/SeldonIO/seldon-core/pull/6029

lc525:
@sakoush Shall we close this PR?

PR #6029, opened 2024-11-05
fix: Use generation id to bootstrap version id

sakoush:

This PR adds the ability to use k8s generation ids to bootstrap version ids for models. This is required so that in the case of scheduler restarts (where the state of the models is lost) the versions carry on from the max versions that could be set. In this way during network partitions we guarantee that the new version is going to be bigger than any potentially existing version from the agents and therefore simplifies the reconciliation process.

We also intentionally waits for the synchronisation process to finish before the controller sending any resources state that it has in the cache and therefore we allow agents to connect first and update the scheduler cache with any existing loaded models before proceeding. This is done by doing a 2 step process on the control plane stream between the scheduler and the the controller. The first step is to signal the number of potential servers (agents) to connect then once they connect and advertise their state we proceed with the second stage. We introduce an event type on this stream to differentiate between the 2 operations.

Another change is to allow status updates for models that are stuck in ModelProgressing state, this is done in `updateLoadedModelsImpl`

Fixes INFRA-1227 (internal)


scheduler/pkg/store/memory.go
```diff
@@ -329,7 +333,7 @@ func (m *MemoryStore) updateLoadedModelsImpl(
 		modelVersion = model.Latest()
 	}
 
-	//  resevere memory for existing replicas that are not already loading or loaded
+	//  revere memory for existing replicas that are not already loading or loaded
```
lc525, line 336:
```suggestion
	//  reserve memory for existing replicas that are not already loading or loaded
```

lc525 (approved):
lgtm. I believe this new approach, being significantly simpler compared to the previous one, really helps us wrt testing. Thank you!

PR #6031, opened 2024-11-06
feat(k6): add scenario with multiple stages ramping up/down RPS

lc525:
The added scenario allows one to configure an arbitrary number of stages,
with each consisting of a linear ramp-up/down to the desired rate and a
hold/plateau time.

Within each stage, the duration for which the inference RPS is held constant
is configured via one element in the `CONSTANT_RATE_DURATIONS_SECONDS`
environment variable (a vector of comma separated values), with the ramp-up/
down duration preceding it being 1/3rd of the hold time.

- scenario already used in HPA autoscaling testing & demo

### DIFF
--- a/tests/k6/scenarios/infer_multiple_rates.js
+++ b/tests/k6/scenarios/infer_multiple_rates.js
@@ -1,0 +1,79 @@
+import { getConfig } from '../components/settings.js'
+import { doInfer, setupBase, teardownBase, getVersionSuffix, applyModelReplicaChange } from '../components/utils.js'
+import { vu } from 'k6/execution';
+
+export const options = {
+    thresholds: {
+        'http_req_duration{scenario:default}': [`max>=0`],
+        'http_reqs{scenario:default}': [],
+        'grpc_req_duration{scenario:default}': [`max>=0`],
+        'data_received{scenario:default}': [],
+        'data_sent{scenario:default}': [],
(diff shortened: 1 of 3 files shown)


tests/k6/scenarios/infer_multiple_rates.js
```diff
@@ -0,0 +1,73 @@
...
+        } else {
+            doInfer(modelName, modelNameWithVersion, config, false, idx) // grpc
+        }
+    } else if (rest_enabled) {
+        doInfer(modelName, modelNameWithVersion, config, true, idx)
+    } else if (grpc_enabled) {
+        doInfer(modelName, modelNameWithVersion, config, false, idx)
+    } else {
+        throw new Error('Both REST and GRPC protocols are disabled!')
+    }
```
sakoush, line 68: should we add the option to trigger `applyModelReplicaChange` as well in these tests?

sakoush (approved):
LGTM. left a minor point.

tests/k6/scenarios/infer_multiple_rates.js
lc525: yes, I'll add that. Until now the data-plane tests and the control-plane tests were quite well separated, now I believe we should always have the option of some control-plane changes during the data-plane load test.

PR #6033, opened 2024-11-07
fix(envoy): add keepalive settings to all envoy configs

sakoush:
This is a change to add keepalive settings to other envoy config files that were missed in https://github.com/SeldonIO/seldon-core/pull/6016 (https://github.com/SeldonIO/seldon-core/pull/6016/commits/9f18ee222a6c6d2bd097c2ae8b4a678f7c92b365)


lc525 (approved):
Lgtm!

PR #6038, opened 2024-11-08
fix: Allow model versions to increase in k6 tests

sakoush:

This change adds some randomisation to rollout new versions of a model as opposed to new replicas.

### DIFF
--- a/tests/k6/components/utils.js
+++ b/tests/k6/components/utils.js
@@ -626,10 +626,12 @@ export function applyModelReplicaChange(config) {
         const modelId = Math.floor(Math.random() * config.maxNumModels[j])
         const modelName = config.modelNamePrefix[j] + modelId.toString()
 
-        let replicas =  Math.floor(Math.random() * config.maxModelReplicas[j]) + 1
-        const model = generateModel(config.modelType[j], modelName, 1, replicas, config.isSchedulerProxy, config.modelMemoryBytes[j], config.inferBatchSize[j])
+        const rand = Math.random()
+        const replicas =  Math.floor(rand * config.maxModelReplicas[j]) + 1
+        const memory = (rand > 0.5)? replicas + "k": "1k"   // this will induce a change in memory causing a new version of the model to be created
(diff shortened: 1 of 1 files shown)


tests/k6/components/utils.js
```diff
@@ -626,10 +626,12 @@ export function applyModelReplicaChange(config) {
         const modelId = Math.floor(Math.random() * config.maxNumModels[j])
         const modelName = config.modelNamePrefix[j] + modelId.toString()
 
-        let replicas =  Math.floor(Math.random() * config.maxModelReplicas[j]) + 1
-        const model = generateModel(config.modelType[j], modelName, 1, replicas, config.isSchedulerProxy, config.modelMemoryBytes[j], config.inferBatchSize[j])
+        const rand = Math.random()
+        const replicas =  Math.floor(rand * config.maxModelReplicas[j]) + 1
+        const memory = (rand > 0.5)? replicas + "k": "1k"   // this will induce a change in memory causing a new version of the model to be created
```
lc525, line 631: it's a bit awkward to also update the memory in an util function called "applyModelReplicaChange". But let's add it like this, and I can modify things afterwards.

This is because I'm already modifying both replicas and memory in the control-plane tests, and the replica (and now memory) changes in utils.js have been introduced independently. I'll unify those.

lc525 (approved):
lgtm, with the comment that I'll update this to be more consistent across scenarios.

tests/k6/components/utils.js
sakoush: The memory itself is not the target for the test here, I am just trying to induce a version change in the test, if you are changing it later then perhaps its fine to get it merged for now.

lc525, line 631: more concretely on the change itself: this will set the memory to "replicas" K. I assume we'll always use this with small models, without any actual issue in terms of scheduling onto servers where the memory hint actuall matters.

sakoush: thats a good point, perhaps ideally I should induce a small change to the provided memory by the user.

sakoush: As discussed will merge as it stands and improve in a follow up PR

PR #6054, opened 2024-11-12
feat(scheduler): account for multiple instances of a model per server when scheduling

driev:

A model may be loaded into memory multiple times depending on server, or model's configuration. The scheduler and agent do not take this into consideration when scheduling and loading a model respectively, which can lead to models being scheduled onto servers that do not have sufficient memory. This aims to acquire the information that's necessary to make more accurate decisions when scheduling.

Fixes # INFRA-1146


scheduler/pkg/agent/repository/mlserver/mlserver.go
```diff
@@ -321,3 +323,12 @@ func (m *MLServerRepositoryHandler) findHighestVersionInPath(modelPath string) (
 	}
 	return "", nil
 }
+
+func (m *MLServerRepositoryHandler) GetModelConfig(path string) (*agent.ModelConfig, error) {
```
lc525, line 327:
```suggestion
func (m *MLServerRepositoryHandler) GetModelConfig(_ string) (*agent.ModelConfig, error) {
```

Given that we have the argument name ("path") in the interface (for documentation purposes), and MLServer doesn't use the path at all, perhaps it's cleaner to mark that explicitly as unused? Just a suggestion, I don't have a strong view on this.

scheduler/pkg/agent/repository/triton/triton_test.go
```diff
@@ -336,3 +338,85 @@ func TestCopyNonConfigFilesToModelRepo(t *testing.T) {
...
+			expectedInstanceCount: 2,
+		}, {
+			name: "other backend without a count defaults to 2",
+			modelConfig: &pb.ModelConfig{
+				Backend:       "other",
+				InstanceGroup: []*pb.ModelInstanceGroup{{Kind: pb.ModelInstanceGroup_KIND_CPU}},
+			},
+			expectedInstanceCount: 1,
+		}, {
+			name: "not KIND_CPU defaults to 1",
```
lc525, line 374: Question: This is 1 specifically because atm we only expose system RAM as something a model can request as required memory, right? I'm asking because by default Triton creates one model copy per available GPU.

apis/mlops/agent/agent.proto
```diff
@@ -27,6 +27,17 @@ message ModelEventMessage {
   Event event = 5;
   string message = 6;
   uint64 availableMemoryBytes = 7;
+  ModelConfig modelConfig = 8;
+}
+
+message ModelConfig {
+  uint32 instanceCount = 1; // The number of instances of the model on a replica
+  enum Resource {
```
lc525, line 35: I'm a bit unsure of the current structure here, mostly because in principle a Model should be able to request _both_ Memory and GPU memory, **and** different number of instances running on CPU, GPUs, etc.  For a LLM, the weights will be first be loaded into RAM and then copied into GPU memory from there (there are fancier options for DMA directly from disks or network cards into GPU memory but let's ignore those for now). This means that I need approximately the same Memory as the GPU memory requirement.

Triton is limited in its expressiveness for this to something like "give me 2 copies on the CPU, 1 copy on GPU 1 and 2 copies each on GPUs 2 and 3". MLServer is limited to "load this model n times" (where we assume the result needs n times the memory)

Given that at the moment our scheduler only considers system memory, I propose we just expose instanceCount, remove the Resource field and iterate on the exact shape that ModelConfig would take in the future in the design doc & future PRs.

scheduler/pkg/agent/model_state.go
```diff
@@ -153,7 +152,7 @@ type modelVersion struct {
 }
 
 func (version *modelVersion) getVersionMemory() uint64 {
-	return version.versionInfo.GetModel().GetModelSpec().GetMemoryBytes()
+	return version.versionInfo.GetModel().GetModelSpec().GetMemoryBytes() * uint64(version.versionInfo.ModelConfig.GetInstanceCount())
```
lc525, line 155: The fact that atm we'll be using the instance count like this (simply ignoring the resource) is another reason why I've proposed we remove the resource field for now.

lc525 (commented):
The general layout of this change is solid, and I think merging this would reduce some surprises in terms of scheduling.

I've suggested some changes to the ModelConfig given that this is a preliminary change, before we start taking into account things like GPU memory, GPU+CPU memory, etc and make our CRDs more expressive. Please let me know what you think.

apis/mlops/agent/agent.proto
driev: Yeah, I think it's better to remove it for now. I need to read through the docs again, but you can either run instances on CPU, or GPU for a model, I don't think you can choose both. 

I'm thinking that maybe a completely separate message for each model config using `oneof` will make it much more flexible, instead of trying to shoehorn it all into one succint message.

scheduler/pkg/agent/repository/triton/triton_test.go
driev: Defaulting to one, as the number of GPUs available is unknown for now.

apis/mlops/agent/agent.proto
driev: You can choose both: https://github.com/triton-inference-server/python_backend/blob/f3068c03ed82e099cef5e2b40e9d1d79b1eab7ac/examples/instance_kind/README.md

apis/mlops/agent/agent.proto
```diff
@@ -27,8 +27,35 @@ message ModelEventMessage {
...
+    }
+}
+
+message MLServerModelConfig {
+  uint32 instanceCount = 1;
+}
+
+message TritonModelConfig {
+  TritonCPU cpu = 1;
+}
```
driev, line 53: Went with `cpu` instead of having just  `instanceCount`, so it would fit in a bit better with something like ` repeated TritonGPU gpu = 2;` in future.

scheduler/pkg/agent/repository/triton/triton.go
```diff
@@ -217,3 +219,26 @@ func (t *TritonRepositoryHandler) SetExplainer(modelRepoPath string, explainerSp
 func (t *TritonRepositoryHandler) SetExtraParameters(modelRepoPath string, parameters []*scheduler.ParameterSpec) error {
 	return nil
 }
+
+func (t *TritonRepositoryHandler) GetModelConfig(path string) (*agent.ModelConfig, error) {
+	configPath := filepath.Join(path, TritonConfigFile)
+	tritonConfig, err := t.loadConfigFromFile(configPath)
+	if err == nil {
+		instanceGroup := tritonConfig.InstanceGroup
+		if len(instanceGroup) > 0 {
```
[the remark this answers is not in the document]
driev: I'll check the triton implementation.

apis/mlops/agent/agent.proto
```diff
@@ -27,8 +27,35 @@ message ModelEventMessage {
...
 }
 
+message ModelConfig {
+  enum Type {
+        UNKNOWN_TYPE = 0;
+        MLSERVER = 1;
+        TRITON = 2;
+        }
+
+  Type type = 1;
```
lc525, line 40: Do we still need the type if you're using oneof? We could just test which of the oneof configs is present via doing a switch on the config type. This would also avoid invalid messages with one Type and the incompatible config.

lc525 (approved):
lgtm, some (hopefully minor) suggestions

apis/mlops/agent/agent.proto
driev: Removed it.

apis/mlops/agent/agent.proto
```diff
@@ -27,8 +27,35 @@ message ModelEventMessage {
   Event event = 5;
   string message = 6;
   uint64 availableMemoryBytes = 7;
+  ModelConfig modelConfig = 8;
 }
 
+message ModelConfig {
```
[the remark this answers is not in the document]
driev: Thanks for the suggestion - I do prefer `ModelRuntimeInfo` to `ModelConfig`.

PR #6056, opened 2024-11-13
feat: Expose pvc retention policy via helm

abhimanyu003:

Expose PVC retention policy via helm. **Retain** is the default value.
Info on retention policy can be found here: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention

Example YAML

```yaml
kind: Server
  name: triton
  namespace: seldon
spec:
  replicas: 1
  serverConfig: triton
    whenDeleted: Retain
    whenScaled: Retain
```

Fixes #INFRA-1235


sakoush (approved):
LGTM. Left a minor comment.

Has this been tested?

PR #6057, opened 2024-11-14
fix(deps): fix mismatched dependency versions

driev:

I was getting some errors and this resolves them.

### DIFF
--- a/tests/integration/go.mod
+++ b/tests/integration/go.mod
@@ -6,8 +6,8 @@ require (
 	github.com/docker/docker v26.1.0+incompatible
 	github.com/onsi/gomega v1.33.1
 	github.com/seldonio/seldon-core/operator/v2 v2.0.0-00010101000000-000000000000
-	k8s.io/api v0.31.2
-	k8s.io/apimachinery v0.31.2
+	k8s.io/api v0.29.2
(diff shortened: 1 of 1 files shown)


tests/integration/go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/docker/docker v26.1.0+incompatible
 	github.com/onsi/gomega v1.33.1
 	github.com/seldonio/seldon-core/operator/v2 v2.0.0-00010101000000-000000000000
-	k8s.io/api v0.31.2
-	k8s.io/apimachinery v0.31.2
+	k8s.io/api v0.29.2
+	k8s.io/apimachinery v0.29.2
 	k8s.io/client-go v0.29.2
```
lc525, line 11: Does upgrading client-go to `v0.31.2` instead lead to failures?

lc525: The versions here should actually be close to the k8s version running in the cluster, or we risk incompatibilities. Therefore perhaps pushing to 31.2 was a bit ambitious.

lc525 (approved):
lgtm

tests/integration/go.mod
driev: And `v0.29.2` matches all of the other modules.

PR #6082, opened 2024-11-19
feat(envoy): switch to ADS

driev:

ADS allows for guranteed order of updates, as the xDS resources are sent over the same stream, instead of independent streams. This will be important later, as we'd like to send cluster Y to a node, before updating a route to point to Y instead of another cluster.

Fixes # INFRA-1420

Moved `server.go` under `processor`, so it's easier to test.

### DIFF
--- a/scheduler/pkg/envoy/processor/server_test.go
+++ b/scheduler/pkg/envoy/processor/server_test.go
@@ -1,0 +1,115 @@
+package processor
+
+import (
+	"context"
+	"slices"
+	"sync"
+	"testing"
+	"time"
+
+	clusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	core "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	client "github.com/envoyproxy/go-control-plane/pkg/client/sotw/v3"
+	resource "github.com/envoyproxy/go-control-plane/pkg/resource/v3"
+	envoyServerControlPlaneV3 "github.com/envoyproxy/go-control-plane/pkg/server/v3"
+	. "github.com/onsi/gomega"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


scheduler/pkg/envoy/processor/server_test.go
```diff
@@ -0,0 +1,156 @@
...
+		go func() {
+			// watch for configs
+			resp, err := c.Fetch()
+			g.Expect(err).To(BeNil())
+			g.Expect(len(resp.Resources)).To(Equal(6))
+			for _, r := range resp.Resources {
+				cluster := &clusterv3.Cluster{}
+				err := anypb.UnmarshalTo(r, cluster, proto.UnmarshalOptions{})
+				g.Expect(err).To(BeNil())
+				g.Expect(slices.Contains(expectedClusterNames, cluster.Name)).To(BeTrue())
```
sakoush, line 131: should we append a list of clusters we get back and then check for equality (instead of `Contains`)?

scheduler/pkg/envoy/processor/server_test.go
```diff
@@ -0,0 +1,156 @@
...
```
sakoush, line 95: should we append a list of clusters we get back and then check for equality (instead of Contains)?

sakoush (approved):
LGTM.

scheduler/pkg/envoy/processor/server_test.go
```diff
@@ -0,0 +1,156 @@
...
+func TestFetch(t *testing.T) {
+	g := NewGomegaWithT(t)
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	logger := log.New()
+
+	snapCache := cache.NewSnapshotCache(true, cache.IDHash{}, logger)
+	go func() {
+		err := startAdsServer(ctx, snapCache, 18001)
```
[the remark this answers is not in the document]
driev: Yep, thanks.

PR #6120, opened 2024-11-29
fix(envoy): use ADS in dynamic config config

driev:

This is breaking envoy and should have been added in https://github.com/SeldonIO/seldon-core/pull/6082


sakoush (approved):
lgtm

PR #6124, opened 2024-12-04
fix(operator): regenerate CRDs

sakoush:

This PR regenerates the CRDs based on the new deps we have checked in.

- Pin kustomize to 5.2.1
- Add CI check for CRD changes

This has been tested on a kind cluster.


.github/workflows/tests.yml
```diff
@@ -22,3 +22,11 @@ jobs:
         run: make -C components/tls test
       - name: test-components-kafka
         run: make -C components/kafka test
+      - name: check CRDs changes
+        run: | 
+          make -C operator generate
```
lc525, line 27:
```suggestion
          make -C operator generate manifests
```
the manifests target is also required

lc525 (approved):

PR #6125, opened 2024-12-04
fix: Add timeout to contexts in client calls

sakoush:

This PR introduced timeouts to client grpc calls in the places that they make sense. This will allow the system in some cases not to wait indefinitely for a response. One example is when we are making a model infer call in `modelgateway.
There are other places where we have introduced a timeout logic on top of GRPC calls and in these cases we left them as they are.

Fixes # INFRA-1252 (internal)


operator/scheduler/experiment.go
```diff
@@ -102,79 +102,78 @@ func (s *SchedulerClient) SubscribeExperimentEvents(ctx context.Context, grpcCli
...
-					}
-					return nil
-				})
-				if retryErr != nil {
-					logger.Error(err, "Failed to remove finalizer after retries")
 				}
+				return nil
+			})
+			if retryErr != nil {
+				logger.Error(err, "Failed to remove finalizer after retries")
```
lc525, line 127: Can we have `!event.Active` but `latestExperiment.ObjectMeta.DeletionTimestamp.IsZero()` ? Because in that case we don't even attempt to remove the finalizer, and this error message might be confusing. For example, I don't know if RetryOnConflict doesn't return an error on context timeout. I suppose we still have the underlying error wrapped here.

lc525: Also should we `continue` after logging here (rather than continuing with the status update)?

operator/scheduler/model.go
```diff
@@ -161,10 +161,12 @@ func (s *SchedulerClient) SubscribeModelEvents(ctx context.Context, grpcClient s
 		// Handle terminated event to remove finalizer
 		if canRemoveFinalizer(latestVersionStatus.State.State) {
 			retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-				latestModel := &v1alpha1.Model{}
+				ctxWithTimeout, cancel := context.WithTimeout(ctx, constants.K8sAPICallTimeout)
```
lc525, line 164: One thing I've realised here is that, despite the name `K8sAPICallTimeout` covers not just one k8s API call but all calls done in one function. So in the example here, both the `s.Get(...)` and `s.Update(...)` need to complete within a total time < than the timeout.

sakoush: This is the intention to simplify things. I can come up with a better naming for the constant to make it clearer.

scheduler/pkg/agent/k8s/secrets_test.go
```diff
@@ -67,7 +67,7 @@ parameters:
 		t.Run(test.name, func(t *testing.T) {
 			fakeClientset := fake.NewSimpleClientset(test.secret)
 			s := NewSecretsHandler(fakeClientset, test.secret.Namespace)
-			data, err := s.GetSecretConfig(test.secretName)
+			data, err := s.GetSecretConfig(test.secretName, 1)
```
lc525, line 70: Isn't this converted into a "1 nanosecond" Duration by default? Even for a mock client, perhaps it would be better to set it to `1 * Millisecond` to avoid any flaky tests.

scheduler/pkg/util/constants.go
```diff
@@ -46,6 +46,21 @@ const (
...
+)
+
+// Servers control plane
+const (
+	ServerControlPlaneTimeout = time.Second * 5
+)
+
+// inference
+const (
+	InferTimeoutDefault = 10 * time.Minute // TODO: expose this as a config (map)?
```
lc525, line 61: In general, I would like us to expose the inference timeout more explicitly and preferably in one place. Atm it's not entirely clear (at least to me) where things timeout (envoy? mlserver? agent? one of the gateways? etc). Perhaps we should add a jira ticket to handle this.

lc525 (approved):
lgtm, I feel more confident with having timeouts everywhere, but I think this change will require intensive testing before the next release, to make sure things are solid (for example, under network partitions, slow connections, etc). Also, I think we might need to expose some of those timeouts as settings further down the line.

I've left a couple of comments/clarifying questions, but all are minor. The git diff did not help on this particular PR, as lots of code appeared changed but most was just moved slightly. Hopefully I haven't missed anything obvious.

Thank you very much for taking this on, it was long overdue.

operator/scheduler/experiment.go
sakoush: actually there is a subtle behaviour here as `event.Active==False` could mean that some models that are part of an experiments are not ready anymore and therefore the whole experiment is not active. This overlaps with also deleting an experiment where `latestExperiment.ObjectMeta.DeletionTimestamp.IsZero()!=0`.

Actually we could `continue` if we managed to successfully remove the finalizer as the CR is going to be removed from k8s.

sakoush: > Also should we `continue` after logging here (rather than continuing with the status update)?

we need to detect first whether we removed the finalizer. I will leave it as a follow up PR though.

PR #6126, opened 2024-12-04
fix(k6): change default k6 image in kustomize

sakoush:

This is for sanity reasons we change the default image name for seldonio. In practice this is not really required as users should deploy via `make` targets.

Fixes #


lc525 (approved):
lgtm!

PR #6139, opened 2024-12-10
fix(docs): add a table for core 2 dependencies in docs

sakoush:
This PR list the 3rd party dependencies of Core 2 that users should be aware of. It also includes min and max supported versions.

Fixes #


paulb-seldon:
We say "Ensure that the version of the Kubernetes cluster is v1.27 or later" but we do have a max version. Should we not explicitly say v1.27-1.31 or the versions below instead?

paulb-seldon (approved):

PR #6146, opened 2024-12-12
fix(ansible): Upgrade deps in ansible install

sakoush:

This PR upgrades the different components to recent supported versions. It also installs additionally prometheus adapter 0.12.0 which is required for HPA demos.

The main change is to move to kafka 3.8 which required some changes to strimzi configs. We completely remove zookeeper support as a result.

The full list of changes can be checked in the commit messages.

Fixes # (internal) INFRA-1267


kafka/strimzi/templates/pool.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: kafka.strimzi.io/v1beta2
```
lc525, line 1: This will work as a first cut, but I've seen issues with scaling-up/down. The settings that I've ended up with are here: https://github.com/SeldonIO/gcp-infra-automation-prototype/blob/main/kafka/strimzi/templates/cluster.yaml

lc525: We can leave those changes for a future PR

ansible/roles/opentelemetry/defaults/main.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 opentelemetry_namespace: opentelemetry-operator-system
 
-opentelemetry_version: v0.92.0
+opentelemetry_version: v0.114.1
```
lc525, line 4: does this actually work in the cluster? I found I had to change the config for otel >= v0.104 because some {receivers, processors, exporters} were deprecated (https://github.com/SeldonIO/gcp-infra-automation-prototype/blob/main/tracing/k8s/otel-collector-beta1.yaml). Furthermore, the OpenTelemetryCollector CR moved from `opentelemetry.io/v1alpha1` to `opentelemetry.io/v1beta1` and for me that looked like this: https://github.com/SeldonIO/gcp-infra-automation-prototype/blob/main/tracing/k8s/otel-collector-beta1.yaml

lc525: However, in that project I'm installing the operator via its helm chart. I've been using it as an experiment in updating dependencies, and I know we'll want to unify the ansible roles down the line.

kafka/strimzi/templates/pool.yaml
sakoush: I am copying across most of the changes.

ansible/roles/prometheus/tasks/main.yaml
```diff
@@ -14,3 +14,12 @@
     chart_ref: "kube-prometheus"
     chart_version: "{{ seldon_monitoring_prometheus_operator_chart_version }}"
     values: "{{ seldon_monitoring_prometheus_operator_values }}"
+
+
+- name: Install Prometheus Adapter
```
lc525, line 19: we'll also want to install the config that Seldon uses, similar to here: `https://docs.seldon.ai/seldon-core-2/kubernetes/hpa-rps-autoscaling#installing-and-configuring-the-prometheus-adapter`. Again, we can take this as a follow-up task (the config map would probably end up having the name `seldon-prometheus-adapter-prometheus-adapter` given the name chosen below for the helm release. In the docs I've called it `hpa-metrics` to avoid the duplication.

kafka/strimzi/templates/pool.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - broker
+    - controller
+
+  storage:
+    type: jbod
+    volumes:
+      - id: 0
+        type: persistent-claim
+        size: {{ .Values.broker.pvcSize }}
+        deleteClaim: false
```
lc525:
```suggestion
        deleteClaim: false

```
adding newline

ansible/roles/opentelemetry/defaults/main.yaml
sakoush: It does work but maybe we should move to the new api?

kafka/strimzi/templates/pool.yaml
lc525: To clarify one change in particular, the split in two volumes, with one of type `type: ephemeral`, was because with KRaft the cluster gets an ID that gets persisted on that volume. 

On scale-down followed by scale-up, the cluster comes up with a new ID but the old volume is loaded showing the past ID. KRaft gets confused and stops the cluster from actually loading.

With an ephemeral volume, the persisted ID data goes away on scale-down and is recreated fresh on scale-up.

lc525 (approved):
lgtm, left a couple of questions and improvement suggestions around otel/strimzi

ansible/roles/opentelemetry/defaults/main.yaml
sakoush: image

lc525: Perhaps what I saw was just some log spam as a result of some deprecated settings, I may have had a PR updating them for the main repo as well though

ansible/roles/prometheus/tasks/main.yaml
sakoush: ok will add the config as well.

kafka/strimzi/templates/pool.yaml
sakoush: added now in [6f9119c](https://github.com/SeldonIO/seldon-core/pull/6146/commits/6f9119c0065b84545287fea6e1d9c38db156a1c9)

PR #6164, opened 2024-12-17
fix(cli): fix port to 9004 in seldon cli deps

sakoush:

In some cases the envars required for the example cli pod deployment are not set up properly. Until we investigate, we pin it now to 9004 which has not changed recently.

Fixes #


cherrymu (approved):
LGTM!

PR #6131, opened 2024-12-09
[update] installation draft

Rajakavitha1:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #6166, opened 2024-12-19
Add Managed Kafka page to latest docs

paulb-seldon:
Porting over Managed Kafka installation page. This page will need to be put into the right section as https://github.com/SeldonIO/seldon-core/pull/6131 is to be merged.


lc525 (approved):
lgtm, I think this is ready to merge as-is. Thanks for porting the docs over to Core 2!

PR #6168, opened 2024-12-19
Remove faulty link

paulb-seldon:
Link in docs page incorrectly takes users home


lc525 (approved):
+1, thanks for catching that

PR #6183, opened 2025-01-02
feat(scheduler): account for number of model instances when scheduling

driev:

As a follow up from https://github.com/SeldonIO/seldon-core/pull/6054, the number of times a model will loaded into memory, will now be taken into account  when scheduling. 

When the control plane initially loads a model, this information will not be available until the first model event message is sent from the agent, which will contain the model specific settings. Model runtime info should only be set once per model version.

Fixes # INFRA-1146

Moved the model runtime info field to the scheduler API, as this is what's stored in the in-memory model store.


scheduler/pkg/agent/client_utils.go
```diff
@@ -54,11 +55,13 @@ func isReady(service interfaces.DependencyServiceInterface, logger *log.Entry, m
 	return backoff.RetryNotify(readyToError, backoffWithMax, logFailure)
 }
 
-func getModifiedModelVersion(modelId string, version uint32, originalModelVersion *agent.ModelVersion, modelRuntimeInfo *agent.ModelRuntimeInfo) *agent.ModelVersion {
+func getModifiedModelVersion(modelId string, version uint32, originalModelVersion *agent.ModelVersion, modelRuntimeInfo *scheduler.ModelRuntimeInfo) *agent.ModelVersion {
 	mv := proto.Clone(originalModelVersion)
 	mv.(*agent.ModelVersion).Model.Meta.Name = modelId
+	if mv.(*agent.ModelVersion).Model.ModelSpec != nil && modelRuntimeInfo != nil {
```
sakoush, line 61: can `ModelSpec` be actually nil? or it is just a safe guard?

Maybe add a note as readers could think that runtime info can be set while model spec is nil?

scheduler/pkg/agent/repository/mlserver/mlserver.go
```diff
@@ -324,12 +323,12 @@ func (m *MLServerRepositoryHandler) findHighestVersionInPath(modelPath string) (
 	return "", nil
 }
 
-func (m *MLServerRepositoryHandler) GetModelRuntimeInfo(_ string) (*agent.ModelRuntimeInfo, error) {
+func (m *MLServerRepositoryHandler) GetModelRuntimeInfo(_ string) (*scheduler.ModelRuntimeInfo, error) {
```
sakoush, line 326: this is perhaps a design question as opposed to something related to this PR:

If for mlserver we are getting the parallelWorkersEnvVar set at the server level, why the choice was to do it at the model level (as all models will have the same value for a given server and can be exposed at the point of the server connecting to the scheduler)?

Is this because Triton does not follow the same principle?

scheduler/pkg/store/memory_test.go
```diff
@@ -982,7 +987,7 @@ func TestUpdateModelState(t *testing.T) {
 					"server": {
 						name: "server",
 						replicas: map[int]*ServerReplica{
-							0: {loadedModels: map[ModelVersionID]bool{{Name: "foo", Version: 2}: true, {Name: "foo", Version: 1}: true}, reservedMemory: memBytes, uniqueLoadedModels: map[string]bool{"foo": true}},
+							0: {loadedModels: map[ModelVersionID]bool{{Name: "foo", Version: 2}: true, {Name: "foo", Version: 1}: true}, reservedMemory: memBytes * 2, uniqueLoadedModels: map[string]bool{"foo": true}},
```
sakoush, line 990: This is not related to this incremental change specifically but I am not sure if we are considering instance count for `reservedMemory` in the actual logic. 
Specifically `reservedMemory` in the cases of parallel loads of models on a given server before the actual memory is used and therefore we can guard against overloading beyong the memory limit. We should also consider (perhaps as a follow up PR) how we can expose the instance count "before" loading the model. For MLServer this seems simple as per my comment about the envar but I am not sure about triton yet.

sakoush (approved):
LGTM. I left a couple of comments for potential follow up PR.

scheduler/pkg/store/memory_test.go
driev: Yeah, that's a good point. I was thinking that getting the instance count before loading the model was merely an optimisation, but it would have a some purpose. Triton would take some effort to solve and might never work before loading the model, as GPUs will be considered when scheduling at some point.

PR #6193, opened 2025-01-15
fix: Delete all versions of a deleted model

abhimanyu003:

**How to replicate the issue**

* Apply model 

```yaml
kind: Model
metadata:
  name: add10
spec:
  storageUri: "gs://seldon-models/scv2/samples/triton_23-03/add10"
  requirements:
  - triton
  - python
 ```

* Update manifest by applying faulty update

```
kind: Model
metadata:
  name: add10
spec:
  storageUri: "gs://seldon-models/scv2/samples/triton_23-03/add10-faulty-url"
  requirements:
  - triton
  - python
```

* Do `seldon model unload add10` 
* Model will remain in memory and never get cleand-up

Fixes #INFRA-1230

### DIFF
--- a/scheduler/pkg/agent/client.go
+++ b/scheduler/pkg/agent/client.go
@@ -621,6 +621,21 @@ func (c *Client) LoadModel(request *agent.ModelOperationMessage, timestamp int64
 		logger.Errorf("there was a problem getting the config for model: %s", modelName)
 	}
 
+	// we are going to clean-up all the older stale version of model
+	for _, v := range c.stateManager.modelVersions.getVersionsForAllModels() {
+		// skip if version is same as current one
+		if v.GetVersion() == modelVersion {
+			continue
+		}
+		c.logger.WithField("func", c.stateManager.modelVersions.getVersionsForAllModels()).Infof("----")
+
+		err := c.ModelRepository.RemoveModelVersion(util.GetVersionedModelName(modelName, v.GetVersion()))
+		if err != nil {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scheduler/pkg/agent/client.go [hunk 1]
```diff
@@ -621,6 +621,21 @@ func (c *Client) LoadModel(request *agent.ModelOperationMessage, timestamp int64
...
 
+	// we are going to clean-up all the older stale version of model
+	for _, v := range c.stateManager.modelVersions.getVersionsForAllModels() {
+		// skip if version is same as current one
+		if v.GetVersion() == modelVersion {
+			continue
+		}
+		c.logger.WithField("func", c.stateManager.modelVersions.getVersionsForAllModels()).Infof("----")
+
+		err := c.ModelRepository.RemoveModelVersion(util.GetVersionedModelName(modelName, v.GetVersion()))
```
sakoush, line 632: This flow is probably not the correct flow as we are not yet sure that we should be removing the old version of this model.

In progressive rollout we have to make sure that the new version of the model is up before removing an old version otherwise we risk inference requests not being served in this transition.

I dont think this is the correct place also to do the clean up as one agent (i.e. a server replica loading one model replica) doesnt have visibility with regards to the actual state of the new version of the model (e.g. that the envoy routes have been updated). Therefore I think that the clean up process should be done somewhere at the scheduler side and not on the agent.

sakoush (commented):
This change is potentially going to cause downtime. This will need to be revisited.

sakoush:
We will revisit in a different PR

PR #6194, opened 2025-01-15
feat(envoy): use the healthcheck filter and a prestop hook to gracefully terminate Envoy

driev:

Some clients will receive resets when `seldon-envoy` is restarting. These can be reduced by slowing down the termination sequence and triggering readiness checks to fail, so the kubelet will remove the pod's endpoint from Service load balancing.   

Fixes # INFRA-1254

Perhaps this should be improved in future with something that at least logs some information about the `preStop` sequence.

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
+++ b/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
@@ -952,7 +952,7 @@ spec:
               - -c
               - |
                 echo -ne "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n" > /dev/tcp/localhost/9901
-                sleep 30
+                sleep '{{ .Values.envoy.preStopSleepPeriodSeconds }}'
         name: envoy
         ports:
         - containerPort: 9000
@@ -978,7 +978,8 @@ spec:
 {{- end }}
       securityContext: {{- toYaml .Values.envoy.securityContext | nindent
         8 }}
-      terminationGracePeriodSeconds: 5
+      terminationGracePeriodSeconds: {{ .Values.envoy.terminationGracePeriodSeconds
+        }}
     replicas: 1
   - annotations: {{- toYaml .Values.dataflow.annotations | nindent 8
       }}
--- a/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
+++ b/k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
@@ -951,7 +951,7 @@ spec:
               - /bin/sh
               - -c
               - |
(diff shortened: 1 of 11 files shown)


k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
```diff
@@ -944,6 +944,15 @@ spec:
...
         imagePullPolicy: '{{ .Values.envoy.image.pullPolicy }}'
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - |
+                echo -ne "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n" > /dev/tcp/localhost/9901
+                sleep 30
```
sakoush, line 955: can 30 be made configurable via helm values?

k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
```diff
@@ -944,6 +944,15 @@ spec:
...
           }}:{{ .Values.envoy.image.tag }}'
         imagePullPolicy: '{{ .Values.envoy.image.pullPolicy }}'
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - |
+                echo -ne "POST /healthcheck/fail HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n" > /dev/tcp/localhost/9901
```
sakoush, line 954: what is ip 9901 (admin?) can this change?

scheduler/k8s/envoy/envoy.yaml
```diff
@@ -47,4 +47,4 @@ spec:
           containerPort: 9000
         - name: envoy-stats
           containerPort: 9003
-      terminationGracePeriodSeconds: 5
+      terminationGracePeriodSeconds: 120
```
sakoush, line 50: should this be made configurable?

sakoush (approved):
LGTM. Let minor comments that can happen as follow ups.

k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
driev: yes, added it.

k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
driev: Yes, it's the admin interface. Made it configurable.

scheduler/k8s/envoy/envoy.yaml
driev: Updated.

PR #6218, opened 2025-01-21
feat: enable min/max replica for Server CR

sakoush:

Work done mostly by @lc252

Fixes #


scheduler/pkg/server/server_status.go
```diff
@@ -163,12 +163,11 @@ func (s *SchedulerServer) SubscribeServerStatus(req *pb.ServerSubscriptionReques
 	}
 }
 
-// TODO - Create a ServerStatusMsg type to disambiguate?
-func (s *SchedulerServer) handleServerEvent(event coordinator.ModelEventMsg) {
+func (s *SchedulerServer) handleServerModelEvent(event coordinator.ModelEventMsg) {
 	logger := s.logger.WithField("func", "handleServerEvent")
```
driev, line 167: I don't have a strong opinion on this, but it might be worth coming up with different name that would describe this event more clearly. For example, `handleModelInstanceEvent`, or something.

driev (approved):

scheduler/pkg/server/server_status.go
sakoush: changed to `handleModelEventForServerStatus`

PR #6220, opened 2025-01-23
fix(envoy): filters are in the wrong place

driev:
Envoy will not start.


sakoush (approved):
LGTM assuming this has been tested.

PR #6221, opened 2025-01-23
feat(scheduler): add partial scheduling based on min replicas

sakoush:
This PR introduces the ability to do partial scheduling for a given model if:
- The number of replicas for all candidate servers is less than the desired replica of the model
- The min replicas of the model is less than the available replicas of the server to load the model onto

In this case we load the model on as many replicas as possible assuming it meets min replicas requirement; we mark the model as Available. We also make sure that when there is more capacity available we retry scheduling to make sure so that the model can potentially make use of this extra capacity.

We also favour servers that have the models loaded to minimise ping-ping behaviour.

Fixes #

### DIFF
--- a/scheduler/pkg/store/memory.go
+++ b/scheduler/pkg/store/memory.go
@@ -377,6 +377,8 @@ func (m *MemoryStore) updateLoadedModelsImpl(
 	// also send an update for progressing models so the operator can update the status in the case of a network glitch where the model generation has been updated
 	// also send an update if the model is not yet at desired replicas, if we have partial scheduling
 
+	// note that we use len(modelVersion.GetAssignment()) to calculate the number of replicas as the status of the model at this point might not reflect the actual number of replicas
+	// in modelVersion.state.AvailableReplicas (we call updateModelStatus later)
 	if replicaStateUpdated || modelVersion.state.State == ScheduleFailed || model.IsDeleted() || modelVersion.state.State == ModelProgressing ||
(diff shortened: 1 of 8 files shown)


scheduler/pkg/store/memory.go
```diff
@@ -375,7 +375,10 @@ func (m *MemoryStore) updateLoadedModelsImpl(
 	// this could be in the cases where we are scaling down a model and the new replica count can be all deployed
 	// and always send an update for deleted models, so the operator will remove them from k8s
 	// also send an update for progressing models so the operator can update the status in the case of a network glitch where the model generation has been updated
-	if replicaStateUpdated || modelVersion.state.State == ScheduleFailed || model.IsDeleted() || modelVersion.state.State == ModelProgressing {
+	// also send an update if the model is not yet at desired replicas, if we have partial scheduling
+
+	if replicaStateUpdated || modelVersion.state.State == ScheduleFailed || model.IsDeleted() || modelVersion.state.State == ModelProgressing ||
+		(modelVersion.state.State == ModelAvailable && len(modelVersion.GetAssignment()) < modelVersion.DesiredReplicas()) {
```
driev, line 381: Oof.. this statement is becoming unwieldy (if it wasn't already). I wonder if it's worth having an explicit task to figure out how to refactor it.

sakoush: Yeah we should fix it, it keeps growing.

PR #6222, opened 2025-01-23
feat(operator): adding a patch for server/spec/replicas upon scaling request

driev:

This adds a method to patch a server's replica count, which is called when a scaling request is received from the scheduler.

Fixes # INFRA-1048

### DIFF
--- a/operator/scheduler/server_test.go
+++ b/operator/scheduler/server_test.go
@@ -252,6 +252,32 @@ func TestSubscribeServerEvents(t *testing.T) {
 			},
 		},
 		{
+			name: "server - with no scaling spec",
+			existingServer: mlopsv1alpha1.Server{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "foo",
+					Namespace:  "seldon",
+					Generation: 1,
+				},
+			},
+			response: &scheduler.ServerStatusResponse{
+				Type:       scheduler.ServerStatusResponse_ScalingRequest,
+				ServerName: "foo",
+				Resources: []*scheduler.ServerReplicaResources{
+					{
+						ReplicaIdx: 0,
+					},
+				},
+				AvailableReplicas:      3,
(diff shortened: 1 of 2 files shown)


operator/scheduler/server_test.go
```diff
@@ -178,103 +178,165 @@ func TestSubscribeServerEvents(t *testing.T) {
 	g := NewGomegaWithT(t)
 
 	type test struct {
-		name               string
-		existing_resources []client.Object
-		results            []*scheduler.ServerStatusResponse
+		name           string
+		existingServer mlopsv1alpha1.Server
```
sakoush, line 182: nit: I prefer to keep using the interface client.Object to follow the pattern used in similar tests. What is the reason to have it specific to `mlopsv1alpha1.Server`?

operator/scheduler/server_test.go
```diff
@@ -178,103 +178,165 @@ func TestSubscribeServerEvents(t *testing.T) {
...
 		{
-			// no scheduler state means lost servers metadata
-			name: "servers - no scheduler state",
-			existing_resources: []client.Object{
-				&mlopsv1alpha1.Server{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:       "foo",
-						Namespace:  "default",
-						Generation: 1,
+			name: "server - with a valid scaling request",
```
sakoush, line 189: could we add a test case for `ServerStatusResponse_ScalingRequest` for a server that has no min and max replicas set in the CR?

sakoush (approved):
LGTM. I left a minor comment around testing.

operator/scheduler/server_test.go
driev: Yes, I'll add that.

operator/scheduler/server_test.go
driev: No need to cast the server when checking if it has a valid scaling spec.

PR #6235, opened 2025-01-28
feat(scheduler): mms send scaling request when model shceduling fails

driev:

When the number of server replicas is less than number of desired model replicas, i.e. when a model is partially available, the scheduler needs to send a scaling request to the operator to increase the number of server replicas. 

Fixes # INFRA-1048 

 - [ ] kind test

### DIFF
--- a/scheduler/pkg/server/server_status.go
+++ b/scheduler/pkg/server/server_status.go
@@ -12,6 +12,7 @@ package server
 import (
 	"time"
 
+	"github.com/pkg/errors"
 	pb "github.com/seldonio/seldon-core/apis/go/v2/mlops/scheduler"
 
 	"github.com/seldonio/seldon-core/scheduler/v2/pkg/coordinator"
@@ -164,7 +165,7 @@ func (s *SchedulerServer) SubscribeServerStatus(req *pb.ServerSubscriptionReques
 }
 
 func (s *SchedulerServer) handleModelEventForServerStatus(event coordinator.ModelEventMsg) {
-	logger := s.logger.WithField("func", "handleServerEvent")
+	logger := s.logger.WithField("func", "handleModelEventForServerStatus")
 	logger.Debugf("Got server state change for %s", event.String())
 
 	err := s.updateServerModelsStatus(event)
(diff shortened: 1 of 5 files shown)


scheduler/pkg/scheduler/scheduler.go
```diff
@@ -194,19 +201,22 @@ func (s *SimpleScheduler) scheduleToServer(modelName string) error {
...
-		logger.Debug(msg)
-		// we do not want to reset the server if it has live replicas or loading replicas
-		// in the case of loading replicas, we need to make sure that we can unload them later.
-		// for example in the case that a model is just marked as loading on a particular server replica
-		// then it gets a delete request (before it is marked as loaded or available) we need to make sure
-		// that we can unload it from the server
-		s.store.FailedScheduling(latestModel, msg, !latestModel.HasLiveReplicas() && !latestModel.IsLoadingOrLoadedOnServer())
-		return errors.New(msg)
+	if !ok {
+		s.serverScaleUp(latestModel)
```
sakoush, line 205: we try not to publish to the eventhub while holding a lock (as it can deadlock). I suggest to return the msg and send the event in the caller instead. This also follows the same pattern elsewhere.

scheduler/pkg/server/server_status.go
```diff
@@ -228,21 +258,24 @@ func (s *SchedulerServer) sendServerStatus() {
...
-			hasExpired, err := sendWithTimeout(func() error { return stream.Send(ssr) }, s.timeout)
-			if hasExpired {
-				// this should trigger a reconnect from the client
-				close(subscription.fin)
-				delete(s.serverEventStream.streams, stream)
-			}
-			if err != nil {
-				logger.WithError(err).Errorf("Failed to send server status event to %s", subscription.name)
-			}
+func (s *SchedulerServer) sendServerStatusResponse(ssr *pb.ServerStatusResponse) {
```
sakoush, line 266: coupling server status and server event in one place is probably not legible. I will deal with it though in a my scale down pr.

scheduler/pkg/server/server_status.go
```diff
@@ -206,7 +225,18 @@ func (s *SchedulerServer) updateServerModelsStatus(evt coordinator.ModelEventMsg
...
+	return err
+}
+
+func (s *SchedulerServer) incrementExpectedReplicas(server *store.ServerSnapshot) {
+	// TODO: should there be some sort of velocity check ?
+	logger := s.logger.WithField("func", "incrementExpectedReplicas")
+	logger.Debugf("will attempt to scale servers to %d for %v", server.Stats.MaxNumReplicaHostedModels, server.Name)
+
+	ssr := createServerStatusUpdateResponse(server)
+	ssr.ExpectedReplicas = int32(server.Stats.MaxNumReplicaHostedModels)
```
sakoush, line 237: should we respect `maxReplicas` on the server if defined? i.e. it should be `min(maxReplicas, server.Stats.MaxNumReplicaHostedModels)`?

scheduler/pkg/store/mesh.go
```diff
@@ -407,6 +409,14 @@ func getInstanceCount(modelRuntimeInfo *pb.ModelRuntimeInfo) uint64 {
 	}
 }
 
+func (m *ModelVersion) ShouldScaleUp() bool {
+	if !util.AutoscalingEnabled(m.GetDeploymentSpec().GetMinReplicas(), m.GetDeploymentSpec().GetMaxReplicas()) {
```
sakoush, line 413: I am not sure if this condition is valid here as we want to decouple server scaling from model min/max replicas being defined.

scheduler/pkg/server/utils.go
```diff
@@ -33,3 +35,16 @@ func sendWithTimeout(f func() error, d time.Duration) (bool, error) {
 		return false, err
 	}
 }
+
+func shouldScaleUp(server *store.ServerSnapshot) bool {
+	if server.Stats != nil {
+		stats := server.Stats
+
+		if server.ExpectedReplicas < 0 { // it's not set
```
sakoush, line 43: is this possible in practice?

sakoush: I ended up in my PR returning the target replica count to set, which turned out to be easier to test as well. check https://github.com/SeldonIO/seldon-core/blob/99bda5e914c2bd3a70659c01a6ff55e7ed6f17c3/scheduler/pkg/server/utils.go#L48

scheduler/pkg/util/model.go
```diff
@@ -36,3 +36,15 @@ func GetOrignalModelNameAndVersion(versionedModel string) (string, uint32, error
 func GetPinnedModelVersion() uint32 {
 	return 1
 }
+
+// if min and max replicas are not set, we do not allow autoscaling
```
sakoush, line 40: just to note that this is just for our internal model autoscaling using lag. in the future we also do not want to make the decision whether autoscaling is enabled by check `minReplicas` and `maxReplicas` (as a future PR).

sakoush (commented):
LGTM. I left some comments.

scheduler/pkg/scheduler/scheduler.go
driev: Ah yes, forgot about that.

scheduler/pkg/server/server_status.go
driev: Yeah, it should - I just checked `isValidScalingEvent` in the controller.

scheduler/pkg/server/utils.go
driev: It's possible - there's even a test for this.... 

I'll add the exra return value.

scheduler/pkg/util/model.go
driev: yep, that makes sense.

scheduler/pkg/store/mesh.go
driev: I've removed it, since the check is made in the agent for the model and the controller will check the server's params.

sakoush (approved):
LGTM. nice work!

PR #6246, opened 2025-02-07
feat(scheduler): Scale down server logic

sakoush:

This PR implements policies for scaling down servers naively as part of the system. The policies are:
- we are trying to pack replicas: max number of replicas for any hosted model is less than the number of expected replicas. If this is met we only trigger it  25% of the time to reduce impact on infer latency.
- we have empty replicas and the server has more than one expected replicas.

Currently the trigger of this logic happens when we unload a model replica as it is a good signal that we might need to also reduce the number of server replicas.  However this doesnt capture all edge cases and in a follow up PR we add more triggers for scaling down.

Fixes # (internal INFRA-1234)

- main logic is defined in `shouldScaleDown`

### DIFF
--- a/scheduler/pkg/server/server_status.go
+++ b/scheduler/pkg/server/server_status.go
@@ -207,7 +207,7 @@ func (s *SchedulerServer) handleServerEvents(event coordinator.ServerEventMsg) {
 	if event.UpdateContext == coordinator.SERVER_SCALE {
 		if shouldScaleDown(server, AllowPackingPercentage) {
 			logger.Infof("Server %s is scaling down", event.ServerName)
-			// TODO send control message to scale down
+			s.sendServerScale(server, uint32(server.ExpectedReplicas)-1)
 		}
 	}
 }
--- a/scheduler/pkg/server/server_status.go
+++ b/scheduler/pkg/server/server_status.go
@@ -194,6 +194,16 @@ func (s *SchedulerServer) handleModelEventForServerStatus(event coordinator.Mode
 	}
 }
 
+func (s *SchedulerServer) handleServerEvents(event coordinator.ServerEventMsg) {
+	logger := s.logger.WithField("func", "handleServerEvent")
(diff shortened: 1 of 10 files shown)


scheduler/pkg/server/server_status.go
```diff
@@ -173,6 +173,24 @@ func (s *SchedulerServer) handleModelEventForServerStatus(event coordinator.Mode
...
+	logger := s.logger.WithField("func", "handleServerEvent")
+	logger.Debugf("Got server state %s change for %s", event.ServerName, event.String())
+
+	server, err := s.modelStore.GetServer(event.ServerName, true, true)
+	if err != nil {
+		logger.WithError(err).Errorf("Failed to get server %s", event.ServerName)
+		return
+	}
+
+	if event.UpdateContext == coordinator.SERVER_SCALE {
```
sakoush, line 186: @driev this is the current proposed location to consume server scale (up / down) events and decide on whether actually to change server replicas. Open to suggestions if there are better places for this logic to be defined.

sakoush: For example `SchedulerServer` could be extended with a `ServerScaler` utility where the logic of scaling up and down could be defined. This is perhaps something we can consider as a follow up PR after we get scale up / down done?

driev: Yeah, since it's likely to change again, it's probably better to wait a bit.

scheduler/pkg/server/utils.go
```diff
@@ -18,6 +19,15 @@ import (
 	"github.com/seldonio/seldon-core/scheduler/v2/pkg/store"
 )
 
+const (
+	// percentage of time we try to pack server replicas if models are not fully packed
```
driev, line 23:
```suggestion
	// percentage of time we try to pack server replicas, i.e. number of server replicas is more than `MaxNumReplicaHostedModels`
```

driev (approved):

PR #6283, opened 2025-02-17
feat(helm): adding min and max replicas to the server helm chart

driev:

Min and max replicas should be easily configurable for consumers of helm.

Fixes # INFRA-1335

### DIFF
--- a/k8s/helm-charts/seldon-core-v2-servers/templates/seldon-v2-servers.yaml
+++ b/k8s/helm-charts/seldon-core-v2-servers/templates/seldon-v2-servers.yaml
@@ -8,6 +8,8 @@ metadata:
   name: mlserver
   namespace: '{{ .Release.Namespace }}'
 spec:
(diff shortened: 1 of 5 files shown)


k8s/helm-charts/seldon-core-v2-servers/templates/seldon-v2-servers.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
   name: mlserver
   namespace: '{{ .Release.Namespace }}'
 spec:
+  maxReplicas: {{ .Values.mlserver.maxReplicas }}
```
lc525, line 11: I think those should be updated here: `https://github.com/SeldonIO/seldon-core/blob/v2/k8s/kustomize/helm-servers/patch_mlserver.yaml` (and similarly for triton) in order for the changes not to be overwritten when running `make create-helm-charts` in the k8s dir.

lc525 (approved):
lgtm!

PR #6284, opened 2025-02-18
fix(scheduler): Remove all versions when model is deleted

sakoush:

This PR makes sure that all versions of a deleted model are unloaded.

Fixes #
(INFRA-1230 internal)


lc525 (approved):
lgtm, thank you very much for sorting this edge case out!

PR #6285, opened 2025-02-19
fix(operator): allow scaling requests for older generation

sakoush:

When we are scaling/up down servers automatically from the scheduler and using the current flow, there is a period of time where generation of the CR is not yet reflected in the scheduler at the time of making the request to scale (because of concurrency).

This change allows all scaling requests to go ahead even if the generations dont match.

Fixes #

Is this safe in all cases?

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -69,7 +69,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .GOLANGCILINT_VERSION := v1.63.4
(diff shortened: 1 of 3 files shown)


lc525 (approved):
lgtm; summary of out-of-band communication for reference:

The counter-example would be other entities (HPA, manual CR changes) modifying the Server CR replicas concurrently. However, we want to avoid dealing with such dual-control cases by marking them as "not supported". 

In the case of manual CR changes, we would expect the client to retry applying the change, although I'll run some tests as well to see exactly how the behaviour here looks from an end-user perspective.

PR #6293, opened 2025-02-27
feat: implemented grpc model streaming

RobertSamoilescu:

This PR implements grpc model streaming introduced in MLServer in [this PR](https://github.com/SeldonIO/MLServer/pull/1750).

Example of python client for stream-stream:
```python

import subprocess
import grpc
from mlops.v2_dataplane.v2_dataplane_pb2_grpc import GRPCInferenceServiceStub
from mlops.v2_dataplane.v2_dataplane_pb2 import ModelInferRequest, InferTensorContents

def get_mesh_ip():
    cmd = f"kubectl get svc seldon-mesh -n seldon-mesh -o jsonpath='{{.status.loadBalancer.ingress[0].ip}}'"
    return subprocess.check_output(cmd, shell=True).decode("utf-8")

def generate_requests():
    prompts = ["Once upon a time"]
    for prompt in prompts:
        yield ModelInferRequest(
            model_name="text-model",
            inputs=[ModelInferRequest.InferInputTensor(
                name="prompt",
                datatype="BYTES",
                shape=[1, 1],
                contents=InferTensorContents(bytes_contents=[prompt.encode()])
            )]
        )


endpoint = get_mesh_ip() + ":80"
channel = grpc.insecure_channel(endpoint)
stub = GRPCInferenceServiceStub(channel)


# Define headers (metadata)
metadata = [
    ("seldon-model", "text-model"),
    ("seldon-internal-model", "text-model_1")
]

# Make the streaming request with metadata
response_stream = stub.ModelStreamInfer(generate_requests(), metadata=metadata)

for response in response_stream:
    print(response)
```

Fixes #


scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -206,16 +208,8 @@ func (rp *reverseGRPCProxy) setTrailer(ctx context.Context, trailer metadata.MD,
...
-	internalModelName, externalModelName, err := rp.extractModelNamesFromContext(ctx)
-	if err != nil {
-		return nil, err
-	}
-	r.ModelName = internalModelName
-	r.ModelVersion = ""
-
-	// to sync between scalingMetricsSetup and scalingMetricsTearDown calls running in go routines
+// to sync between scalingMetricsSetup and scalingMetricsTearDown calls running in go routines
+func syncScalingMetrics(rp *reverseGRPCProxy, internalModelName string, logger log.FieldLogger) {
```
sakoush, line 212: if we are passing `reverseGRPCProxy` here already does it make sense for `syncScalingMetrics` to be a receiver method instead? i.e `func (rp *reverseGRPCProxy) syncScalingMetrics(internalModelName string`?

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -231,6 +225,20 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
...
+	logger := rp.logger.WithField("func", "ModelInfer")
+	internalModelName, externalModelName, err := rp.extractModelNamesFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	r.ModelName = internalModelName
+	r.ModelVersion = ""
+
+	// handle scaling metrics
+	syncScalingMetrics(rp, internalModelName, logger)
```
sakoush, line 240: based on the comment above I suggest to change the interface to be rp.syncScalingMetrics(internalModelName)` (`rp` has access to logger already)?

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -260,6 +268,88 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
...
+func (rp *reverseGRPCProxy) ModelStreamInfer(stream v2.GRPCInferenceService_ModelStreamInferServer) error {
+	ctx := stream.Context()
+	logger := rp.logger.WithField("func", "ModelStreamInfer")
+	internalModelName, externalModelName, err := rp.extractModelNamesFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	// handle scaling metrics
+	syncScalingMetrics(rp, internalModelName, logger)
```
sakoush, line 280: ditto

sakoush (commented):
I left a minor comment until I review the rest of the PR.

scheduler/pkg/agent/rproxy_grpc_test.go
```diff
@@ -123,6 +126,44 @@ func TestReverseGRPCServiceSmoke(t *testing.T) {
...
+		ctx = metadata.AppendToOutgoingContext(ctx, util.SeldonInternalModelHeader, dummyModelNamePrefix+modelSuffixInternal, util.SeldonModelHeader, dummyModelNamePrefix+modelSuffix)
+
+		stream, err := client.ModelStreamInfer(ctx)
+		if err != nil {
+			logger.Errorf("Error creating stream: %v", err)
+			return nil, err
+		}
+
+		responses := []*v2.ModelInferResponse{}
+		for i := 0; i < numStreamMessages; i++ {
```
sakoush, line 146: I think this is only testing the happy path ie. when there are not errors. perhaps in grpc stream there are more subtle issues when errors occur perhaps worth adding to testing.

sakoush (commented):
I left some comments, the main ones are:
- We are not setting trailer(s) with grpc stream and this is going to affect tracing. I think this should be done.
- While the code is dealing well with errors, I am not sure if this captures all the edge cases. So either expand testing or do manual integration testing to confirm.
- Out of sanity could you test with triton as well to make sure that this change is universal.
- Metrics seems fine but in the future we might want to add more metrics related to grpc streaming (future PR).
- Check tracing (jaeger) that it is behaving ok. if running in `compose` tracing should be exposed automatically [here](https://docs.seldon.ai/seldon-core-2/installation/getting-started/docker-installation#tracing).

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -260,6 +268,88 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
...
+				logger.WithError(err).Error("Failed to forward request")
+				return
+			}
+		}
+	}()
+
+	// receive responses and forward them
+	var finalErr error
+	for {
+		client_stream_resp, err := client_stream.Recv()
```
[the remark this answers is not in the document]
RobertSamoilescu: I think it can be done that way too, but note that `finalErr` is gonna be overwritten by `:=` operator. Note that we also don't set `finalErr` when EOF error happens. If we just use one, I would need to do checks for EOF instead of simply comparing it to nil ... Happy to modify it if you really think it would make the code more clear

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -206,16 +208,8 @@ func (rp *reverseGRPCProxy) setTrailer(ctx context.Context, trailer metadata.MD,
...
-	internalModelName, externalModelName, err := rp.extractModelNamesFromContext(ctx)
-	if err != nil {
-		return nil, err
-	}
-	r.ModelName = internalModelName
-	r.ModelVersion = ""
-
-	// to sync between scalingMetricsSetup and scalingMetricsTearDown calls running in go routines
+// to sync between scalingMetricsSetup and scalingMetricsTearDown calls running in go routines
+func (rp *reverseGRPCProxy) syncScalingMetrics(internalModelName string, logger log.FieldLogger) {
```
sakoush, line 212: `reverseGRPCProxy` has already access to a logger, there is no need to pass it here?

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -260,6 +268,104 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
...
+		elapsedTime := time.Since(startTime).Seconds()
+		go rp.metrics.AddModelInferMetrics(externalModelName, internalModelName, metrics.MethodTypeGrpc, elapsedTime, codes.NotFound.String())
+		return status.Error(codes.NotFound, fmt.Sprintf("Model %s not found (err: %s)", internalModelName, err))
+	}
+
+	// Create an outgoing context for the proxy call to service from incoming context
+	outgoingCtx, requestId := rp.createOutgoingCtxWithRequestId(ctx)
+
+	var trailer metadata.MD
+	opts := append(rp.callOptions, grpc.Trailer(&trailer), grpc_retry.Disable())
```
sakoush, line 295: why do we disable `grpc_retry`?

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -260,6 +268,104 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
...
+
+	client_stream, err := rp.getV2GRPCClient().ModelStreamInfer(outgoingCtx, opts...)
+	if err != nil {
+		logger.WithError(err).Error("Failed to create stream")
+		return err
+	}
+
+	// receive incoming request from envoy and forward them to the model
+	var req_err error
+	done_req := make(chan bool)
```
sakoush: nit: alternatively this could be a wait group? or empty struct as I dont think we make use of the actual boolean value?

sakoush (commented):
left some minor comments

scheduler/pkg/agent/rproxy_grpc.go
RobertSamoilescu: streaming does not support retry

sakoush (approved):
Great work! just add some inline comments to describe the logic that you followed and it is then good to go.

scheduler/pkg/agent/rproxy_grpc.go
```diff
@@ -260,6 +268,154 @@ func (rp *reverseGRPCProxy) ModelInfer(ctx context.Context, r *v2.ModelInferRequ
 	return resp, err
 }
 
+type InferMessage interface {
+	*v2.ModelInferRequest | *v2.ModelInferResponse
+}
+
+// forwardStream forwards messages from recv to send, modifying them with modify
+// before sending. Two instances of forwardStream are launched, one for each
+// direction of the stream, from client to server and server to client. If one
```
sakoush, line 277:
```suggestion
// direction of the stream, from client <> reverse proxy and  reverse proxy <> (model) server. If one
```

sakoush (approved):
LGTM

PR #6310, opened 2025-03-11
fix: Allow kafka client (`librdkafka`)  to respect log level

sakoush:

Currently we expose on pipelinegateway and modelgateway `--log-level` to set the log level for these components. This PR sets also kafka log level (for librdkafka) from this arg.

Note that `librdkafka` uses sysinfo log levels (in `log_level` @ https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md)  which is slightly different from logrus levels. We introduce a mapping to allow it to be used accordingly. 

We also expose this arg via helm, which can be set for these components.

Fixes # internal INFRA-1343

- [x] Enable settings for scheduler as well?

### DIFF
--- a/components/kafka/pkg/config/config.go
+++ b/components/kafka/pkg/config/config.go
@@ -38,13 +38,13 @@ const (
 	KafkaLogLevel         = "log_level"
 )
 
-type sysinfoLogLevel int
+type sysLogLevel int
 
 // Based on syslog levels
 // https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1
 // which is used in librdkafka log_level
 const (
-	emergLevel sysinfoLogLevel = iota
+	emergLevel sysLogLevel = iota
 	alertLevel
 	critLevel
 	errLevel
@@ -76,7 +76,7 @@ func CloneKafkaConfigMap(m kafka.ConfigMap) kafka.ConfigMap {
 }
 
 // note that we also try to match logrus levels here as well
-func parseSysInfoLogLevel(lvl string) (sysinfoLogLevel, error) {
+func parseSysLogLevel(lvl string) (sysLogLevel, error) {
 	switch strings.ToLower(lvl) {
 	case "panic", "emerg":
 		return emergLevel, nil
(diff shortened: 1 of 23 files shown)


components/kafka/pkg/config/config.go
```diff
@@ -34,6 +35,23 @@ type stringSet = map[string]none
 const (
 	KafkaBootstrapServers = "bootstrap.servers"
 	KafkaDebug            = "debug"
+	KafkaLogLevel         = "log_level"
+)
+
+type sysinfoLogLevel int
+
+// Based on sysinfo log levels
```
lc525, line 43:
```suggestion
// Based on syslog levels
```

components/kafka/pkg/config/config.go
```diff
@@ -34,6 +35,23 @@ type stringSet = map[string]none
...
+// e.g. https://linuxconfig.org/introduction-to-the-linux-kernel-log-levels
```
lc525, line 44:
```suggestion
// https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1
```

components/kafka/pkg/config/config.go
```diff
@@ -34,6 +35,23 @@ type stringSet = map[string]none
 const (
 	KafkaBootstrapServers = "bootstrap.servers"
 	KafkaDebug            = "debug"
+	KafkaLogLevel         = "log_level"
+)
+
+type sysinfoLogLevel int
```
lc525, line 41: Propose we rename everywhere to sysLogLevel (including in function names)

lc525 (approved):
lgtm with a single proposed name change for clarity (and pointing to the Syslog RFC which defines the levels).

components/kafka/pkg/config/config.go
sakoush: done in [7edde1a](https://github.com/SeldonIO/seldon-core/pull/6310/commits/7edde1a5a3f8fcb55a990899378bafe7a7fc64de)

PR #6330, opened 2025-03-18
fix(compose): Allow triton rclone port to be set

sakoush:

This is to revert a partial change done in https://github.com/SeldonIO/seldon-core/pull/6312/files#diff-759bb154316174296ba09eb1d405f8adaa82fd21162a04001bb3592ac88427d5, which would allow the rclone container to be started with a custom port (for triton).

Fixes #


lc525 (approved):
lgtm

PR #6331, opened 2025-03-19
fix(scheduler): Do not try to unload versions that are not live

sakoush:

This PR adds a guard to skip unloading oldversions of the model that are not live as this is not required.

Otherwise this will cause issues downstream. 

Fixes #

INFRA-1367 (internal)

Expanded test to cover this case.


lc525 (approved):
lgtm

PR #6349, opened 2025-03-25
fix(operator): Apply scheduler runtime podSpec override

sakoush:

This PR actually applies the runtime override for scheduler.podSpec as perviously they were not including in the merging process for deployment.

Fixes #

Tested in kind


lc525 (approved):
lgtm, and thank you for also adding test coverage for this!

PR #6350, opened 2025-03-25
fix: adjust logging level for dataflow

sakoush:

This PR fixes log level for dataflow engine in the following ways:
- Apply app log level properly, previously it was always fixed at debug. 
- Separate out kafka log level from app log level helm config. To this extent, `dataflow.logLevelKafka` is introduced.

Fixes #

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
@@ -21,11 +21,12 @@ object Logging {
         appLevel: Level = Level.INFO,
         kafkaLevel: Level = Level.WARN,
     ) = loggingConfiguration {
-        kloggingMinLogLevel(appLevel)
         sink(STDOUT_SINK, RENDER_ISO8601, STDOUT)
         logging {
-            fromLoggerBase("io.seldon")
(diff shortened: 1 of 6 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
```diff
@@ -21,11 +21,12 @@ object Logging {
         appLevel: Level = Level.INFO,
         kafkaLevel: Level = Level.WARN,
     ) = loggingConfiguration {
-        kloggingMinLogLevel(appLevel)
```
lc525, line 24: I think we can also keep kloggingMinLogLevel, set to appLevel, which sets the logging level for the logger itself: https://klogging.io/docs/internals/internal-logger/

lc525 (approved):
lgtm, with a minor suggestion

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Logging.kt
sakoush: done in [5598a67](https://github.com/SeldonIO/seldon-core/pull/6350/commits/5598a67f4bf20d47b43b5ddcc90a1469a95faed2)

PR #6353, opened 2025-03-25
feat: allow removal of kafka model topics

RobertSamoilescu:

This PR allows the user to specify whether the input and output kafka topics associated with a model should be deleted when the model is unloaded. This can be achieved by setting the `cleanTopicsOnDelete` flag in the manifest file. Example of manifest file:
```yaml
kind: Model
metadata:
  name: iris
spec:
  storageUri: "gs://seldon-models/scv2/samples/mlserver_1.6.0/iris-sklearn"
  dataflow:
    cleanTopicsOnDelete: true
  requirements:
  - sklearn
  memory: 100Ki
```

If the flag is missing, it defaults to `cleanTopicsOnDelete: false`.

Fixes #


apis/mlops/scheduler/scheduler.proto
```diff
@@ -31,6 +31,7 @@ message DeploymentSpec {
   uint32 minReplicas = 2;
   uint32 maxReplicas = 3;
   bool logPayloads = 4;
+  bool CleanTopicsOnDelete = 5; // clean up the kafka topic on model delete
```
lc525, line 34: nit: `cleanTopicsOnDelete` to match the camelCase of other protobuf fields?

scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -296,9 +332,10 @@ func (kc *InferKafkaHandler) AddModel(modelName string) error {
 	return nil
 }
 
-func (kc *InferKafkaHandler) RemoveModel(modelName string) error {
+func (kc *InferKafkaHandler) RemoveModel(modelName string, cleanTopicsOnDeletion bool) error {
```
lc525, line 335: RemoveModel is called any time we receive a status update from the scheduler and that reports that the model [has zero available replicas](https://github.com/SeldonIO/seldon-core/blob/a6deb439330b97e486d91d89609690cc2cab470f/scheduler/pkg/kafka/gateway/client.go#L176). This might be a bit eager: I suspect that if the scheduler gets disconnected from inference server agents (network partition), or (easier to test) if a model gets scaled down to 0 replicas, then the topics get deleted. We would want to avoid that.  We should also test what happens if a model gets updated (new version of the CR applied): are the topics deleted & recreated? We would also want to avoid that scenario (loosing topic data on model version update)

Would you be able to test those scenarios?

lc525 (commented):
The PR as a whole looks good. We might need to change the logic of when topic deletion happens (or the microservice responsible for the topic deletion in the first place), depending on the result of a couple of tests I've suggested in comments.

apis/mlops/scheduler/scheduler.proto
```diff
@@ -17,6 +17,7 @@ message Model {
   ModelSpec modelSpec = 2;
   DeploymentSpec deploymentSpec = 3;
   StreamSpec streamSpec = 4;
```
lc525, line 19: Just noticed the existence of "StreamSpec" above, holding options related to dataflow (the topic names) which as far as I know are never used. However, let's keep things as they are; we might repurpose the StreamSpec in case we need to configure Core 2 streaming in any way.

scheduler/pkg/kafka/gateway/infer.go
RobertSamoilescu: The scenarios above seem to be covered by [this](https://github.com/SeldonIO/seldon-core/blob/02a7a8e6457285d53cc7700b207f815e6e90603f/scheduler/pkg/kafka/gateway/client.go#L162-L170).

lc525 (approved):
Could we test one further thing here before merging?

1. Create pipeline with at least two models, both with `cleanTopicsOnDelete: true`
2. Send inference requests to pipeline
3. Delete one of the models
4. Send some more inference requests to pipeline (those are likely to break or stall)
5. Recreate the deleted model and wait for it to become available, together with the pipeline
6. Send some inference requests to the pipeline (those should succeed), and if not we should understand why and what else needs to be fixed

We should consider this as an integration test to be automated.

RobertSamoilescu:
> Could we test one further thing here before merging?
> 
> 1. Create pipeline with at least two models, both with `cleanTopicsOnDelete: true`
> 2. Send inference requests to pipeline
> 3. Delete one of the models
> 4. Send some more inference requests to pipeline (those are likely to break or stall)
> 5. Recreate the deleted model and wait for it to become available, together with the pipeline
> 6. Send some inference requests to the pipeline (those should succeed), and if not we should understand why and what else needs to be fixed
> 
> We should consider this as an integration test to be automated.

Tested the scenario with 3 models chained `a -> b -> c`:
* when all models are present, the result is returned as expected. 
* when we delete for example, model `b`, pipeline ready becomes `False`. If we send a request to the pipeline, the request is hanging until timeout. In dataflow engine the following message keeps appearing.
```
│ Exception in thread "seldon-mesh-seldon-dataflow-64188c55e8ef132e8e60879d48af746e-89ef91b6-9e23-4701-9450-9deede5d5952-StreamThread-1" org.apache.kafka.streams.errors.MissingSourceTopicException: One or more source topics were missing during r │
│     at org.apache.kafka.streams.processor.internals.StreamsRebalanceListener.onPartitionsAssigned(StreamsRebalanceListener.java:58)                                                                                                                 │
│     at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.invokePartitionsAssigned(ConsumerCoordinator.java:329)                                                                                                                       │
│     at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.onJoinComplete(ConsumerCoordinator.java:478)                                                                                                                                 │
│     at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.joinGroupIfNeeded(AbstractCoordinator.java:478)                                                                                                                              │
│     at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.ensureActiveGroup(AbstractCoordinator.java:389)                                                                                                                              │
│     at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.poll(ConsumerCoordinator.java:564)                                                                                                                                           │
│     at org.apache.kafka.clients.consumer.KafkaConsumer.updateAssignmentMetadataIfNeeded(KafkaConsumer.java:1220)                                                                                                                                    │
│     at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1179)                                                                                                                                                                │
│     at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1159)                                                                                                                                                                │
│     at org.apache.kafka.streams.processor.internals.StreamThread.pollRequests(StreamThread.java:1014)                                                                                                                                               │
│     at org.apache.kafka.streams.processor.internals.StreamThread.pollPhase(StreamThread.java:962)                                                                                                                                                   │
│     at org.apache.kafka.streams.processor.internals.StreamThread.runOnce(StreamThread.java:766)                                                                                                                                                     │
│     at org.apache.kafka.streams.processor.internals.StreamThread.runLoop(StreamThread.java:617)                                                                                                                                                     │
│     at org.apache.kafka.streams.processor.internals.StreamThread.run(StreamThread.java:579)  
```

Once we reload the delete model, `b`, if the request is still hanging, the response is returned. The error message above stops appearing once the model is reintroduced.

* Topics are deleted and recreated as expected when unloading and reloading the model if the flag `cleanTopicsOnDelete: true` is set.

PR #6356, opened 2025-03-27
feat(docs): Server native autoscaling

sakoush:

This PR adds docs for native server autoscaling and the associated implemented policies.

Fixes #

- [x] Link this page in TOC.
- [ ] Adjust content of kubernetes/autoscaling.md
- [x] Get gitbook to render math content


docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+
+{% hint style="info" %}
+**Note**: Native autoscaling of servers is required in the case of MMS as the models are dynamically loaded and unloaded onto these server replicas. In this case Core 2 would autoscale server replicas according to changes to the model replicas that are required. This is in contrast to single-model autoscaling approach explained [here](kubernetes/hpa-rps-autoscaling.md) where server and model replicas are both autoscaled using. Server autoscaling can be used in the case of single-model serving, simplifying the autoscaling process as users would only needs to manage the scaling logic for model replicas, only requiring one HPA manifest.
```
lc525, line 7:
```suggestion
**Note**: Native autoscaling of servers is required in the case of MMS as the models are dynamically loaded and unloaded onto these server replicas. In this case Core 2 would autoscale server replicas according to changes to the model replicas that are required. This is in contrast to single-model autoscaling approach explained [here](kubernetes/hpa-rps-autoscaling.md) where server and model replicas are independently scaled using HPA (but relying on the same metric). Server autoscaling can also be used in the case of single-model serving, simplifying the autoscaling process as users would only needs to manage the scaling logic for model replicas, only requiring one HPA manifest.
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
+# Server _Native_ Autoscaling
+
+## Overview
+Core 2 runs with long lived server replicas that can host multiple models (through multi-model serving, or MMS). These server replicas can be autoscaled natively to host these dynamically loaded models, allowing users to seamlessly optimize the infrastructure cost associated with their deployments. This document outlines the autoscaling policies and mechanisms that are available for autoscaling server replicas. These policies are designed to ensure that the server replicas are increased (scaled up) or decreased (scaled down) in response to changes in the number of model replicas that are required. In other words if a given model is scaled up, the system will scale up the server replicas to host these models. Similarly, if a given model is scaled down, the system will scale down the server replicas that are hosting these models.
```
lc525, line 4:
```suggestion
Core 2 runs with long lived server replicas, each able to host multiple models (through multi-model serving, or MMS). The server replicas can be autoscaled natively by Core 2 in response to dynamic changes in the requested number of model replicas, allowing users to seamlessly optimize the infrastructure cost associated with their deployments. 

This document outlines the autoscaling policies and mechanisms that are available for autoscaling server replicas. These policies are designed to ensure that the server replicas are increased (scaled up) or decreased (scaled down) in response to changes in the number replicas requested for each model. In other words if a given model is scaled up, the system will scale up the server replicas in order to host the new model replicas. Similarly, if a given model is scaled down, the system _may_ scale down the number of replicas of the server hosting the model, depending on other models that are loaded on the same server replica.
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+
+{% hint style="info" %}
+**Note**: Native autoscaling of servers is required in the case of MMS as the models are dynamically loaded and unloaded onto these server replicas. In this case Core 2 would autoscale server replicas according to changes to the model replicas that are required. This is in contrast to single-model autoscaling approach explained [here](kubernetes/hpa-rps-autoscaling.md) where server and model replicas are both autoscaled using. Server autoscaling can be used in the case of single-model serving, simplifying the autoscaling process as users would only needs to manage the scaling logic for model replicas, only requiring one HPA manifest.
+{% endhint %}
+
+## Requirements
+
+To enable autoscaling of server replicas, the following requirements need to be met:
+1. Setting `minReplicas` and `maxReplicas` in the `Server` CR. This will define the minimum and maximum number of server replicas that can be created.
+2. Setting `autoscaling.autoscalingServerEnabled` to `true` (default) during installation of Core 2. This will enable the autoscaling of server replicas.
```
lc525, line 14:
```suggestion
2. Setting the `autoscaling.autoscalingServerEnabled` value to `true` (default) during installation of the Core 2 `seldon-core-v2-setup` helm chart. If not installing via helm, setting the `ENABLE_SERVER_AUTOSCALING` environment variable to `true` in the `seldon-scheduler` podSpec  (via either SeldonConfig or a SeldonRuntime podSpec override) will have the same effect. This will enable the autoscaling of server replicas.
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+apiVersion: mlops.seldon.io/v1alpha1
+kind: Server
+metadata:
+  name: mlserver
+  namespace: seldon
+spec:
+  replicas: 2
+  minReplicas: 1
+  maxReplicas: 4
+maxReplicas: 4
```
lc525, line 28: duplicate `maxReplicas` (incorrect indent)?

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+
+{% hint style="info" %}
+**Note**: Not setting `minReplicas` and/or `maxReplicas` will also effectively disable autoscaling of server replicas. In this case, the user will need to manually scale the server replicas by setting the `replicas` field in the `Server` CR. This allows external autoscaling mechanisms to be used e.g. HPA. In future versions of Core 2, we might relax this requirement and allow native autoscaling of server replicas without setting both `minReplicas` and `maxReplicas`.
+{% endhint %}
+
+## Server Scale Up
+
+### Overview
+
+When we want to scale up a model replicas, the associated servers might not have enough capacity (replicas) available. In this case we need to scale up the server replicas to match the number required by our models.
```
lc525, line 40:
```suggestion
When we want to scale up the number of replicas for a model, the associated servers might not have enough capacity (replicas) available. In this case we need to scale up the server replicas to match the number required by our models.
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+
+There is a period of time where the new server replicas are being created and the new model replicas are being loaded onto these server replicas. During this period, the system will ensure that the existing server replicas are still serving load so that there is no downtime during the scale up process. This is achieved by using partial scheduling of the new model replicas onto the new server replicas. This ensures that the new server replicas are gradually loaded with the new model replicas and that the existing server replicas are still serving load. Check the [Partial Scheduling](../models/scheduling.md) document for more details.
+
+## Server Scale Down
+
+### Overview
+
+When we want to scale down a model replicas, the associated servers might be left unused. In this case, extra server pods (especially if they use GPUs) are wasted causing infrastructure cost for the user.
```
lc525, line 56:
```suggestion
Once we have scaled down the number of replicas for a model, some of the corresponding server replicas might be left unused (depending on whether those replicas also hosted other models or not). In this case, the extra server pods are wasting resources and causing additional infrastructure cost (especially if they have expensive resources such as GPUs attached)
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+
+Scaling down servers in sync with models is not straight forward in the case of multi-model serving. Scaling down one model does not necessarily mean that we also need to scale down the corresponding server replica as this server replica might be still serving load for other models. 
+
+Therefore we need to define some heuristics that can be used to scale down servers if we think that they are not properly used.
+
+{% hint style="info" %}
+**Note**: Scaling down the number of replicas for a model server does not necessarily mean that the system is going to remove a specific replica that we want.
```
lc525, line 63:
```suggestion
**Note**: Scaling down the number of replicas for an inference server does not necessarily mean that the system is going to remove a specific replica that we want. 
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+Scaling down servers in sync with models is not straight forward in the case of multi-model serving. Scaling down one model does not necessarily mean that we also need to scale down the corresponding server replica as this server replica might be still serving load for other models. 
+
+Therefore we need to define some heuristics that can be used to scale down servers if we think that they are not properly used.
+
+{% hint style="info" %}
+**Note**: Scaling down the number of replicas for a model server does not necessarily mean that the system is going to remove a specific replica that we want. 
+
+As currently we have the model server deployed as StatefulSets, scaling down the number of replicas will mean that we are removing a pod with the largest index.
+
+The system will rebalance afterwards, where the models from this draining server replica will be rescheduled but this requires some wait time to achieve. This draining process should allow for no downtime as models are being rescheduled onto other server replicas before the draining server replica is removed.
```
lc525, line 67:
```suggestion
The system will rebalance afterwards, where the models from this draining server replica will be rescheduled but this requires some wait time to achieve. This draining process is done without incurring downtime as models are being rescheduled onto other server replicas _before_ the draining server replica is removed.
```

docs-gb/kubernetes/server-autoscaling.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Policies
+
+1. **Empty Server Replica**:
+    
+In the simplest case we can remove a server replica if it does not host any models. This guarantees that there is no load on a particular server replica before removing it.
+
+This policy works best in the case of single model serving where the server replicas are only hosting a single model. In this case, if the model is scaled down, the server replica will be empty and can be removed.
+
+However in the case of MMS it can lead to less optimal packing of models onto server replicas. This is because the system will not automatically pack models onto the smaller set of replicas. This can lead to more server replicas being used than necessary. This can be mitigated by the lightly loaded server replicas policy.
```
lc525, line 78:
```suggestion
However in the case of MMS, only reducing the number of server replicas when a server no longer hosts any models can lead to a suboptimal packing of models onto server replicas. This is because the system will not automatically pack models onto the smaller set of replicas. This can lead to more server replicas being used than necessary. This can be mitigated by the lightly loaded server replicas policy.    
```

lc525 (approved):
lgtm with some minor suggestions

docs-gb/kubernetes/server-autoscaling.md
sakoush: done in 
01e2376

PR #6383, opened 2025-04-09
feat: allow removal of kafka pipeline topics

RobertSamoilescu:

This PR allows the user to specify whether the input and output kafka topics associated with a pipeline should be deleted when the pipeline is unloaded. This can be achieved by setting the `cleanTopicsOnDelete` flag in the manifest file. Example of manifest file:
```yaml
kind: Pipeline
metadata:
  name: iris-pipeline
spec:
  dataflow:
    cleanTopicsOnDelete: true
  steps:
    - name: iris
  output:
    steps:
    - iris
```
If the flag is missing, it defaults to `cleanTopicsOnDelete: false`.

Fixes #


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/KafkaAdmin.kt
```diff
@@ -104,6 +104,26 @@ class KafkaAdmin(
         return null
     }
 
+    suspend fun deleteTopics(steps: List<PipelineStepUpdate>): Exception? {
+        try {
+            steps
+                .flatMap { step -> step.sourcesList + step.sink + step.triggersList }
```
lc525, line 110: Why are we also adding `step.triggersList` here?

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -285,12 +288,22 @@ class PipelineSubscriber(
...
+                    .withMessage("kafka streams topic deletion error")
+        }
+
         client.pipelineUpdateEvent(
             makePipelineUpdateEvent(
                 metadata = metadata,
                 operation = PipelineOperation.Delete,
-                success = true,
-                reason = "pipeline removed",
+                success = pipelineError == null,
```
lc525, line 305: A note that `success == false` will transition the Pipeline to "PipelineFailed" on the scheduler side and prevent this pipeline from being removed from k8s. However, in this particular case we have removed the pipeline, and it's just that we weren't able to delete the topics. I think this is fine, as long as we retry deleting the topics when we delete a pipeline already in the "PipelineFailed" state.

lc525 (approved):
lgtm

- To test what happens when dataflow-engine returns `success=false` in `handleDelete` and whether deleting the pipeline again from k8s triggers the same code-path again.

RobertSamoilescu:
> lgtm
> 
> * To test what happens when dataflow-engine returns `success=false` in `handleDelete` and whether deleting the pipeline again from k8s triggers the same code-path again.

As discussed offline, the pipeline is deleted from the operator since the scheduler does not wait for the confirmation from the dataflow engine whether the deletion has succeeded or not. Thus, no error is returned to the operator from the grpc call.

PR #6413, opened 2025-04-22
Feat/cyclic pipelines

RobertSamoilescu:

This PR allows the user to define cyclic pipelines. We define a new specification `allowCycles` which is a boolean flag with default value of `false`. This flag is added because careful consideration is needed when working with cyclic pipelines since it can easily lead to infinite loops. Thus, this flag ensures that the user wants to use a cyclic pipeline and is aware of the implications.

Example of CRD:
```
kind: Pipeline
metadata:
  name: cyclic-pipeline
spec:
  allowCycles: true
  steps:
    - name: a
    - name: b
      inputs:
        - a.outputs
        - c.outputs
    - name: c
      inputs:
        - b.outputs
  output:
    steps:
    - c
```

Fixes #


lc525 (approved):
lgtm;

PR #6419, opened 2025-04-23
feat: custom kafka stream join window

RobertSamoilescu:
Currently, the `joinWindowMs` for Kafka Streams provided in the manifest file are not passed to the dataflow engine. Dataflow engine uses a predefined value for the length of the joining window (60s). For a pipeline with cycles, this might results in an infinite loop. 

This PR allows the user to pass a custom value for `joinWindowMs` which will be then used by the dataflow engine for the given pipeline, and thus infinite loop scenarios as above can be avoided.

Fixes #


lc525 (approved):
lgtm, thank you for implementing this!

PR #6445, opened 2025-05-12
feat: statefulsets to deployments for servers

RobertSamoilescu:

Fixes #


lc525 (commented):
Let's add a feature flag for enabling deployments for servers, and default to statefulsets for now.

k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
```diff
@@ -385,6 +385,30 @@ metadata:
...
   namespace: '{{ .Release.Namespace }}'
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mlserver-models
+  namespace: '{{ .Release.Namespace }}'
+spec:
+  accessModes:
+  - ReadWriteOnce
```
lc525, line 395: This allows reads/writes to the underlying PV **from a single k8s node**. This means things will start failing if the server replicas are spread across multiple nodes (as they should be in a HA setup). The problem is that the correct access mode (ReadWriteMany) is not directly supported by many cloud providers (GKE for example doesn't support it). Therefore we may need to think of a different solution for this (or storage of Model artifacts in Core 2 general)

lc525:
> * still need to decide about the persistent volume on the agent which is cleared at initialization. This PR will create a singe pv which will be shared by all replicas (before each replica had its own pv).

Indeed. This also means that a new server replica coming up after models are loaded on a different replica will clear the model storage for everyone, right? We'll need to deal with this as well.

RobertSamoilescu:
> > * still need to decide about the persistent volume on the agent which is cleared at initialization. This PR will create a singe pv which will be shared by all replicas (before each replica had its own pv).
> 
> Indeed. This also means that a new server replica coming up after models are loaded on a different replica will clear the model storage for everyone, right? We'll need to deal with this as well.

That is correct. That's why I decided to move to shared volumes. Also avoids some race conditions when it comes to loading a new model and clearing the models dir.

k8s/Makefile
```diff
@@ -8,23 +8,57 @@ HELM_RUNTIME_BASE := helm-charts/seldon-core-v2-runtime/templates
...
+	sed -i 's/HACK_REMOVE_ME//' ${HELM_COMPONENTS_BASE}/_components-deployments.tpl
+	sed 's/labelz:/labels:\n\ \ \ /' ${HELM_SERVERS_BASE}/_servers-deployments.tpl > ${HELM_SERVERS_BASE}/._servers-deployments.tpl
+	sed -i 's/annotationz:/annotations:\n\ \ \ /' ${HELM_SERVERS_BASE}/._servers-deployments.tpl
+	sed -i 's/HACK_REMOVE_ME//' ${HELM_SERVERS_BASE}/._servers-deployments.tpl
+	sed -zi 's#\(apiVersion: rbac.authorization.k8s.io/v1\nkind: ClusterRole.*\- get\n\-\-\-\)#{{- if and (not .Values.controller.skipClusterRoleCreation) (or .Values.controller.clusterwide .Values.controller.watchNamespaces) -}}\n\1\n{{- end }}#' ${HELM_COMPONENTS_BASE}/_components-deployments.tpl
+	sed -zi 's#\(apiVersion: rbac.authorization.k8s.io/v1\nkind: ClusterRoleBinding.*ServiceAccount\n  name: seldon-v2-controller-manager\n  namespace:[^-]\{27\}\n\-\-\-\)#{{- if or .Values.controller.clusterwide .Values.controller.watchNamespaces -}}\n\1\n{{- end }}#' ${HELM_COMPONENTS_BASE}/_components-deployments.tpl
+	mv ${HELM_SERVERS_BASE}/._servers-deployments.tpl ${HELM_SERVERS_BASE}/_servers-deployments.tpl
+
+.PHONY: merge-configs
+merge-configs:
```
lc525, line 45: It's quite nice to keep the two types of settings separated and then merge them as you've done here. +1

k8s/Makefile
```diff
@@ -8,23 +8,57 @@ HELM_RUNTIME_BASE := helm-charts/seldon-core-v2-runtime/templates
...
+	sed -i '1s;^;{{- define "setup.statefulsets" }}\n;' ${HELM_COMPONENTS_BASE}/_components-statefulsets.tpl && echo '{{- end }}' >> ${HELM_COMPONENTS_BASE}/_components-statefulsets.tpl
+	sed -i '1s;^;{{- define "setup.deployments" }}\n;' ${HELM_COMPONENTS_BASE}/_components-deployments.tpl && echo '{{- end }}' >> ${HELM_COMPONENTS_BASE}/_components-deployments.tpl
+	@{ \
+	echo '{{- if .Values.useServerWithDeployment }}'; \
```
lc525:
```suggestion
	echo '{{- if .Values.useDeploymentsForServers }}'; \
```

k8s/helm-charts/seldon-core-v2-servers/values.yaml
```diff
@@ -1,3 +1,5 @@
+useServerWithDeployment: false
```
lc525:
```suggestion
useDeploymentsForServers: false
```

k8s/helm-charts/seldon-core-v2-servers/values.yaml
```diff
@@ -12,4 +14,4 @@ triton:
   maxReplicas: 1
   statefulSetPersistentVolumeClaimRetentionPolicy:
     whenDeleted: Retain
-    whenScaled: Retain
+    whenScaled: Retain
```
lc525: nit: no newline at EOF

k8s/Makefile
lc525: Suggested the name change for increased clarity; It will need to be modified in a couple of other places as well.

k8s/kustomize-experimental/helm-components-sc/patch_mlserver.yaml
```diff
@@ -0,0 +1,104 @@
+apiVersion: mlops.seldon.io/v1alpha1
```
lc525, line 1: As a note, I see that now there is quite a bit of duplication of those settings across kustomize and kustomize-experimental, which means that any change would need to happen twice. I think at this stage (experimental) this is fine, and we can try and see if we can have the common parts refactored out when we're moving the Deployments feature out of experiment stage.

k8s/kustomize/helm-components-sc/patch_controller.yaml
```diff
@@ -37,3 +37,5 @@ spec:
             value: '{{ hasKey .Values.controller "logLevel" | ternary .Values.controller.logLevel .Values.logging.logLevel }}'
           - name: WATCH_NAMESPACES
             value: '{{ join "," .Values.controller.watchNamespaces }}'
+          - name: USE_SERVER_WITH_DEPLOYMENT
+            value: '{{ .Values.useServerWithDeployment }}'
```
lc525: no newline at EOF

operator/config/crd/bases/mlops.seldon.io_serverconfigs.yaml
```diff
@@ -8229,7 +8229,6 @@ spec:
                 type: array
             required:
             - podSpec
-            - volumeClaimTemplates
```
lc525, line 8232: isn't this still used for the statefulset case?

operator/config/serverconfigs-experimental/kustomization.yaml
```diff
@@ -0,0 +1,18 @@
...
+- triton.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: agent
+  newName: seldonio/seldon-agent
+  newTag: latest
+- name: mlserver
+  newName: seldonio/mlserver
+  newTag: 1.6.1
```
lc525: Please move to 1.7.1

operator/config/serverconfigs/mlserver.yaml
```diff
@@ -96,6 +96,8 @@ spec:
           resourceFieldRef:
             containerName: mlserver
             resource: requests.memory
+      - name: SELDON_USE_SERVER_WITH_DEPLOYMENT
```
lc525: For consistency, if you decide to make the suggested name change, those should probably be `SELDON_USE_DEPLOYMENTS_FOR_SERVERS`

operator/controllers/mlops/server_controller.go
```diff
@@ -40,9 +40,10 @@ import (
 // ServerReconciler reconciles a Server object
 type ServerReconciler struct {
 	client.Client
-	Scheme    *runtime.Scheme
-	Scheduler *scheduler.SchedulerClient
-	Recorder  record.EventRecorder
+	Scheme                  *runtime.Scheme
+	Scheduler               *scheduler.SchedulerClient
+	Recorder                record.EventRecorder
+	UseServerWithDeployment bool
```
lc525: name change candidate

operator/controllers/mlops/server_controller.go
```diff
@@ -148,7 +160,7 @@ func serverReady(status mlopsv1alpha1.ServerStatus) bool {
 }
 
 func (r *ServerReconciler) updateStatusFromError(ctx context.Context, logger logr.Logger, server *mlopsv1alpha1.Server, err error) {
-	server.Status.CreateAndSetCondition(mlopsv1alpha1.StatefulSetReady, false, err.Error())
+	server.Status.CreateAndSetCondition(mlopsv1alpha1.DeploymentReady, false, err.Error())
```
lc525: isn't this code used by both StatefulSets and Deployments? (same question for the next change in this file)

operator/main.go
```diff
@@ -92,13 +92,14 @@ func getWatchNamespaceConfig(namespace, watchNamespaces string, clusterwide bool
...
-		clusterwide          bool
-		logLevel             string
+		metricsAddr             string
+		enableLeaderElection    bool
+		probeAddr               string
+		namespace               string
+		watchNamespaces         string
+		clusterwide             bool
+		logLevel                string
+		useServerWithDeployment bool
```
lc525: name change candidate

lc525 (approved):
lgtm as a general approach, and can be merged as an experimental feature that we'll continue testing. I've suggested a couple of possible improvement areas. The ones you agree are relevant but which should have separate PRs should get their own tasks in Jira.

PR #6446, opened 2025-05-13
(docs) Update the dependencies table

Rajakavitha1:
Incorporate Chandras suggestions.
One suggestion from side is  https://docs.seldon.ai/seldon-core-2/installation/installation#seldon-core-2-dependencies - Don't add envoy and rclone as dependencies as they can confuse the customers although we are explaining them below. They are part of Core 2 components we ship so customers don't have to tweak anything 

Fixes #


paulb-seldon (approved):
I shortened the note, apart from that happy to merge

PR #6498, opened 2025-06-02
feat(dataflow): dataflow scalability

RobertSamoilescu:

This PR enhances the scalability of the pipelines/dataflow engine by allowing multiple replicas of a pipeline to run across multiple dataflow engine instances. Previously, each pipeline could only run on a single instance (referred to here as a pipeline replica), due to a replication factor of 1 enforced by the load balancer. This meant that increasing the number of dataflow engine instances did not increase pipeline replication or scalability.

Key Improvements:

* **Pipeline Replication Logic**
  * By default, the number of pipeline replicas is set to min(# of dataflow engine instances, # of Kafka partitions). Replicating beyond this threshold is avoided, as it would lead to idle resources.

* **Configurable Replicas via SeldonConfig and SeldonRuntime**

  * SeldonConfig sets the upper limit based on Kafka partitions (e.g., max 4 for 4 partitions).

  * SeldonRuntime allows for finer control by setting the maximum number of total pipeline instances (e.g., up to #partitions × #pipelines).
  
   For example, with 4 Kafka partitions and 3 pipelines, the theoretical maximum is 12 replicas. If SeldonRuntime limits it to 9, only 9 instances will be launched. Importantly, these limits are not overwritten in the deployment manifest—they are used to cap the number of running dataflow engine instances.

* **Dynamic Scaling of Pipeline Replicas**

  * The number of dataflow engine replicas automatically adjusts as pipelines are added or removed.

  E.g., if SeldonRuntime allows 20 replicas, and 3 pipelines with 4 partitions each are running (12 total replicas), adding a new pipeline would spin up 4 more replicas. Similarly, removing a pipeline will reduce the count by the corresponding number of replicas.

* **Robust Pipeline Scheduling with Vector Clocks**

  * Pipeline scheduling now accounts for asynchronous operations and potential race conditions during concurrent pipeline creation, deletion, or rebalancing.

  * Each pipeline now maintains a vector clock (logical timestamp) that is incremented with every event (create/delete/rebalance).

  * This timestamp is sent with scheduler messages and returned by the dataflow engine in acknowledgments.

  * The scheduler uses this timestamp to detect and ignore stale messages, preventing invalid state transitions (e.g., processing a deletion acknowledgment after a creation confirmation).

  * This mechanism is especially critical when scaling out dataflow engine replicas, which can trigger frequent rebalancing and race conditions across multiple servers subscribing to the scheduler.

Fixes #


k8s/helm-charts/kafka-nodepool.yaml
lc525, line 1: This should not be part of the helm-charts. Kafka configuration in production is the responsibility of end-clients, and on our end the example install for pre-prod/testing relies on ansible. A strimzi configuration that is used by ansible already exists here: https://github.com/SeldonIO/seldon-core/blob/v2/kafka/strimzi/templates/pool.yaml

k8s/helm-charts/kafka.yaml
lc525, line 1: This should not be part of the helm-charts. A strimzi configuration that is used by ansible already exists here: https://github.com/SeldonIO/seldon-core/blob/v2/kafka/strimzi/templates/cluster.yaml

k8s/helm-charts/seldon-core-v2-setup/templates/seldon-v2-components.yaml
```diff
@@ -790,10 +790,6 @@ spec:
         env:
         - name: MODELGATEWAY_NUM_WORKERS
           value: '{{ .Values.modelgateway.workers }}'
-        - name: KAFKA_DEFAULT_REPLICATION_FACTOR
```
lc525: Not all end-users rely on helm (but rather, they have k8s manifests derived from `k8s/yaml/*`). They will want the system to continue working if applying the previous config, or we will need to carefully coordinate an upgrade path. Removing those environment vars looks like a breaking change on our part, that I would like to avoid if possible. Is that the case?

Furthermore, the same env vars are kept for `seldon-dataflow-engine`. I'm probably missing something here, so let's discuss when we go over the PR

scheduler/pkg/kafka/dataflow/conflict_resolution.go
```diff
@@ -0,0 +1,140 @@
...
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/seldonio/seldon-core/apis/go/v2/mlops/chainer"
+
+	"github.com/seldonio/seldon-core/scheduler/v2/pkg/store/pipeline"
+)
+
+type ConflictResolutioner struct {
+	vectorClock          map[string]uint64                             // key: pipeline name, value: vector clock timestamp
```
lc525: Isn't there a data race on those maps?; The maps should likely be protected by a RWMutex, and the field-level access done via methods on the type.

operator/controllers/reconcilers/common/validator.go
```diff
@@ -0,0 +1,59 @@
...
+	maxReplicas := numPartitions
+	if pipelineCount != 0 {
+		maxReplicas = numPartitions * pipelineCount
+	}
+
+	logger.Info("Maximum replicas for dataflow engine", "max_replicas", maxReplicas)
+
+	if component.Replicas != nil && *component.Replicas > maxReplicas {
+		component.Replicas = &maxReplicas
+		logger.Info("Adjusted dataflow engine replicas to max", "replicas", maxReplicas)
```
lc525: This should probably be published as an Event on the SeldonConfig or SeldonRuntime CR, rather than just logging it in the operator.

operator/controllers/mlops/seldonconfig_controller.go
```diff
@@ -53,3 +70,27 @@ func (r *SeldonConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+func ValidateSeldonConfig(
+	ctx context.Context,
+	clt client.Client,
+	seldonConfig *mlopsv1alpha1.SeldonConfig,
+	namespace *string,
+	logger logr.Logger,
+) error {
+	for _, component := range seldonConfig.Spec.Components {
+		logger.Info("Checking component", "component_name", component.Name)
```
lc525: Consider using Debug here to avoid noisy logs. Any error should contain the name of the component where the validation failed. For now, that's only `dataflow-engine`. Please also consider whether there are other places where Info should be converted into Debug

operator/controllers/mlops/seldonruntime_controller.go
```diff
@@ -240,6 +244,37 @@ func (r *SeldonRuntimeReconciler) mapSeldonRuntimesFromSeldonConfig(_ context.Co
...
+	logger := log.FromContext(ctx).WithName("mapSeldonRuntimesFromPipeline")
+
+	pipeline, ok := obj.(*mlopsv1alpha1.Pipeline)
+	if !ok {
+		logger.Error(fmt.Errorf("unexpected type %T", obj), "expected Pipeline")
+		return nil
+	}
+
+	var seldonRuntimes mlopsv1alpha1.SeldonRuntimeList
+	if err := r.Client.List(ctx, &seldonRuntimes); err != nil {
```
lc525: We have at most one SeldonRuntime per namespace. I suggest we use something like `r.Client.List(ctx, &seldonRuntimes, client.InNamespace(pipeline.Namespace))` so that we don't iterate through all SeldonRuntimes for each pipeline.

operator/controllers/reconcilers/seldon/runtime_reconciler.go
```diff
@@ -27,9 +31,88 @@ type SeldonRuntimeReconciler struct {
...
+		maxReplicas = numPartitions * pipelineCount
+	}
+
+	if component.Replicas != nil && *component.Replicas > maxReplicas {
+		component.Replicas = &maxReplicas
+		recorder.Eventf(
+			runtime,
+			v1.EventTypeWarning,
+			"DataflowEngineReplicasAdjusted",
+			fmt.Sprintf("Dataflow engine replicas adjusted to %d based on KafkaConfig and Pipeline count", maxReplicas),
```
lc525, line 72: Might be worth being slightly more explicit in the log message about where the KafkaConfig lies in an installed cluster. It is in `SeldonConfig`, (overwrite-able by `SeldonRuntime`) right?

lc525 (approved):
lgtm;

Because now adding a pipeline may involve creating more dataflow-engine replicas, together with the associated movement of some existing pipelines onto the new replica (likely causing some unavailability for "victim" pipelines), I would like us to assess (via some control-plane stress tests with pipelines) the overall impact of adding one or more pipelines onto the availability of other pipelines.

We may need to write a dedicated k6 scenario for this. However, this assessment can be done as part of INFRA-1507 (internal)

**Next feature** for dataflow scalability (future PRs): allow for dedicated dataflow-engines (or tag-based?) even before reaching the max number of replicas (catering for the situation where you have "whale" pipelines and "small" pipelines, with more of the small ones being able to share the same underlying resources. Just something to think about for now.

PR #6506, opened 2025-06-09
fix: dataflow message reconcile

RobertSamoilescu:

This PR implements a robust pipeline scheduling with vector clocks:

  * Pipeline scheduling now accounts for asynchronous operations and potential race conditions during concurrent pipeline creation, deletion, or rebalancing.

  * Each pipeline now maintains a vector clock (logical timestamp) that is incremented with every event (create/delete/rebalance).

  * This timestamp is sent with scheduler messages and returned by the dataflow engine in acknowledgments.

  * The scheduler uses this timestamp to detect and ignore stale messages, preventing invalid state transitions (e.g., processing a deletion acknowledgment after a creation confirmation).

  * This mechanism is especially critical when scaling out dataflow engine replicas, which can trigger frequent rebalancing and race conditions across multiple servers subscribing to the scheduler.

Fixes #


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
```diff
@@ -189,6 +189,11 @@ class Pipeline(
             steps: List<PipelineStepUpdate>,
             kafkaDomainParams: KafkaDomainParams,
         ): Pair<Topology, Int> {
+            // Sort the steps by the sink to ensure the same
+            // order when building the topology amongst multiple
+            // replicas. The scheduler doesn't send the same message
+            // because the steps are created from iterating over a map
+            val sortedSteps = steps.sortedBy { it.sink.topicName }
```
lc525, line 196: This is probably unrelated to the vector clock changes, but happy for it to stay in this PR as well (we'd want to include it as a minimum in the next patch release)

scheduler/pkg/kafka/dataflow/server.go
```diff
@@ -373,6 +400,9 @@ func contains(slice []string, val string) bool {
 }
 
 func (c *ChainerServer) rebalance() {
+	c.mu.Lock()
```
lc525, line 403: In this case, I would rather go back to the finer-grained (per pipeline) locking. This is because when you have lots of pipelines, we might end up holding the lock here for a long time.

lc525 (approved):
lgtm, with one suggestion to revert to finer-grained locking in rebalance()

scheduler/pkg/kafka/dataflow/server.go
RobertSamoilescu: As previously discussed, a race condition can occur when a pipeline rebalance and model deletion are executed concurrently.

The rebalance process works as follows:
* It first reads the list of currently running pipelines.
* It then acquires a lock to coordinate pipeline migrations (via delete/create operations) across dataflow engines.

However, if a model deletion is processed after the rebalance reads the running pipelines but before it acquires the lock, the following issue arises:
* The rebalance may  resurrect a deleted pipeline by migrating it (delete-then-create) to another engine.
* As a result, a pipeline that should have been terminated remains active.

PR #6519, opened 2025-06-11
fix(Scheduler): No dataflow engines available for terminated pipelines

MiguelAAe:
**What this PR does / why we need it**: Terminating and terminated pipeline statuses aren't currently saved in the db when a pipeline Scheduler receives an unload request for a given pipeline (its status of deleted true is saved). On a restart the pipeline is restored with status terminate which causes the Scheduler to go through the pipeline state machine of `terminate` -> `terminating` -> `terminated` and since this process is faster than stablishing a connection to Dataflow from the ChainerServer we print `no dataflow engines available to handle pipeline`. 

In short: restarts of the schedulers causes it to print warn logs of `no dataflow engines available to handle pipeline` of deleted pipelines 

Fixes #
* Suppress warning logs of no dataflow engines available to handle pipelines with status terminated
* Terminating and Terminated pipeline statuses are now saved in badger db

Tests #
* creation of pipelines, deletion and restart verified do not produce `no dataflow engines available to handle pipeline`
* creation of pipelines, deletion and restart verified do not produce `no dataflow engines available to handle pipeline` when the scheduler is updated to this fix.

### DIFF
--- a/scheduler/pkg/store/pipeline/store.go
+++ b/scheduler/pkg/store/pipeline/store.go
@@ -114,7 +114,6 @@ func (ps *PipelineStore) InitialiseOrRestoreDB(path string, deletedResourceTTL u
 func (ps *PipelineStore) restorePipeline(pipeline *Pipeline) {
 	logger := ps.logger.WithField("func", "restorePipeline")
 	ps.mu.Lock()
-	logger.Infof("Adding pipeline %s with state %s", pipeline.GetLatestPipelineVersion().String(), pipeline.GetLatestPipelineVersion().State.Status.String())
 	err := ps.modelStatusHandler.addPipelineModelStatus(pipeline)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


scheduler/pkg/store/pipeline/store.go
```diff
@@ -394,7 +394,8 @@ func (ps *PipelineStore) setPipelineStateImpl(name string, versionNumber uint32,
 					// to set the status of the new version of the pipeline to ready
 					evts = append(evts, ps.terminateOldUnterminatedPipelinesIfNeeded(pipeline)...)
 				}
-				if !pipeline.Deleted && ps.db != nil {
+				if ps.db != nil {
+					ps.logger.Infof("saving pipeline %s to db with status %s", pipeline.Name, status.String())
```
lc525: I would suggest switching this log from info to debug. Please mark the PR as non-draft when you think it's ready :-)

PR #6522, opened 2025-06-12
feat(modelgateway): Added support for custom worker timeout

RobertSamoilescu:

This PR adds support for setting a custom worker timeout in modelgateway through envs:
```yaml
env:
  - name: MODELGATEWAY_WORKER_TIMEOUT_MS
    value: "20000"
``` 

Fixes #

### DIFF
--- a/scheduler/pkg/kafka/gateway/infer.go
+++ b/scheduler/pkg/kafka/gateway/infer.go
@@ -35,7 +35,7 @@ const (
 	pollTimeoutMillisecs        = 10000
 	DefaultNumWorkers           = 8
 	EnvVarNumWorkers            = "MODELGATEWAY_NUM_WORKERS"
-	DefaultWorkerTimeoutMs      = 10 * 60 * 1000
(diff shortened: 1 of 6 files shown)


scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -35,6 +35,8 @@ const (
 	pollTimeoutMillisecs        = 10000
 	DefaultNumWorkers           = 8
 	EnvVarNumWorkers            = "MODELGATEWAY_NUM_WORKERS"
+	DefaultWorkerTimeoutMs      = 10 * 60 * 1000
```
lc525: Can we default this to 2 minutes instead? I feel that 10 minutes was/is way too long as a default. We could also leave it at 10 minutes and force everyone to reduce it according to their usecase, but I'm for a more sensible default.

lc525 (approved):
lgtm, suggesting a smaller (but still very generous) default. Do you see any reason why we should keep it to 10min?

PR #6538, opened 2025-06-17
feat(modelgw): modelgw scalability

RobertSamoilescu:
# Why
## Issues

In the current implementation, `modelgw` can only scale up to the number of topic partitions. This limitation exists because model creation events are broadcast to all `modelgw` instances.

## Motivation

The goal of this PR is to enable `modelgw` to scale beyond the number of partitions. Specifically, it introduces a mechanism to scale up to `#partitions × maxNumConsumers` (with `maxNumConsumers` defaulting to 100).

To achieve this, we introduce a load balancer that distributes worker assignments across existing `modelgw` replicas. For each model, a `consumerGroupId` is computed and mapped to min(#partitions, #existingModelGwReplicas). Within each replica, the model is handled by a consumer that belongs to the designated `consumerGroupId`. This approach ensures that no consumer group has members subscribed to different topics.

# What
## Summary of changes

Key changes include:
* introduced load balancer for `modelgw`
* limit scaling up to `#partitions x maxNumConsumers`
* `maxNumConsumers` is provided as an env to `modelGw` and `scheduler`

## Checklist


scheduler/cmd/scheduler/main.go
```diff
@@ -224,7 +224,7 @@ func main() {
 	}
 
 	// scheduler <-> dataflow grpc
-	dataFlowLoadBalancer := util.NewRingLoadBalancer(1)
+	dataFlowLoadBalancer := util.NewRingLoadBalancer(4)
```
lc525: this needs to be rebased on top of #6498

scheduler/cmd/scheduler/main.go
```diff
@@ -276,12 +276,17 @@ func main() {
 	)
 
 	// scheduler <-> controller grpc
+	modelGwLoadBalancer := util.NewRingLoadBalancer(4)
```
lc525:
```suggestion
	modelGwLoadBalancer := util.NewRingLoadBalancer(4)
	// scheduler <-> controller grpc
```

Here, I assume "4" is a temporary number used for testing

lc525:
Happy with the overall approach, I am aware further work is needed, but it relies on the dataflow scalability being merged first.

scheduler/cmd/scheduler/main.go
RobertSamoilescu: Yes

vtaskow (approved):

PR #6557, opened 2025-06-23
feat(Scheduler): Deal with model replicas being set to 0

MiguelAAe:
# Why

Seldon Core allows the creation of models from the crd defitions to have `replicas`, `minReplicas` and `maxReplicas` specifications. However assignment of 0 **replicas** is converted to 1 replica.

This has led to unusual behavior when customers want to downscale from 3 model replicas to 0; the model spec will change to 0 replicas but its status will still be 3 model replicas and the model will not be fully unloaded from the cluster.

## Motivation
As more customers see the value of scaling to 0 this has become a necessary feature in Seldon Core

## Summary of changes
* Change `GetValidatedScalingSpec` function to allow only automatically set replicas to 1 when replicas feature aren't present and allow setting replicas to 0. (this function affects both models and servers)
* new tests `TestGetValidatedScalingSpec`
* new model status `ModelScaledDown` 
* new diagram for Loading a model

## Checklist
- [x] Added/updated documentation

### DIFF
--- a/operator/apis/mlops/v1alpha1/utils_test.go
+++ b/operator/apis/mlops/v1alpha1/utils_test.go
@@ -114,13 +114,25 @@ func TestGetValidatedScalingSpec(t *testing.T) {
 			wantErr: "",
 		},
 		{
-			name:        "success - unset replica params defaults to 1 when min replicas is set",
+			name:        "success - unset replica params defaults to value of min replicas",
 			replicas:    nil,
-			minReplicas: ptr.Int32(1),
+			minReplicas: ptr.Int32(2),
+			maxReplicas: nil,
+			expected: &ValidatedScalingSpec{
+				Replicas:    2,
+				MinReplicas: 2,
+				MaxReplicas: math.MaxUint32,
+			},
+			wantErr: "",
+		},
+		{
+			name:        "success - unset replica params defaults to 1 when min replicas is set to 0",
+			replicas:    nil,
+			minReplicas: ptr.Int32(0),
 			maxReplicas: nil,
 			expected: &ValidatedScalingSpec{
 				Replicas:    1,
-				MinReplicas: 1,
+				MinReplicas: 0,
 				MaxReplicas: math.MaxUint32,
 			},
 			wantErr: "",


operator/apis/mlops/v1alpha1/model_types_test.go
```diff
@@ -451,6 +451,182 @@ func TestAsModelDetails(t *testing.T) {
...
+			wantErr: "",
+		},
+		{
+			name:        "unset replica params defaults to 1",
+			replicas:    nil,
+			minReplicas: nil,
+			maxReplicas: nil,
+			expected: &ValidatedScalingSpec{
+				Replicas:    1,
+				MinReplicas: 0,
```
lc525: Should this be valid, with `Replicas > MaxReplicas` ? I would have considered this case to validate to `{Replicas: 1, MinReplicas: 0, MaxReplicas: math.MaxUint32}` (essentially, MaxReplicas unconstrained by Core 2, there will of course be k8s limitations well before MaxUint32)

operator/apis/mlops/v1alpha1/model_types_test.go
```diff
@@ -451,6 +451,182 @@ func TestAsModelDetails(t *testing.T) {
...
+		},
+		{
+			name:        "unset replica params defaults to 1",
+			replicas:    nil,
+			minReplicas: i32(1),
+			maxReplicas: nil,
+			expected: &ValidatedScalingSpec{
+				Replicas:    1,
+				MinReplicas: 1,
+				MaxReplicas: 0,
```
lc525: Same question here: `MinReplicas > MaxReplicas`, & `Replicas > MaxReplicas`

operator/apis/mlops/v1alpha1/model_types_test.go
```diff
@@ -451,6 +451,182 @@ func TestAsModelDetails(t *testing.T) {
...
+			},
+			wantErr: "",
+		},
+		{
+			name:        "min replica to 1 and replica to 0 should convert to min replica",
+			replicas:    i32(0),
+			minReplicas: i32(1),
+			maxReplicas: nil,
+			expected: &ValidatedScalingSpec{
```
lc525: We could just consider this invalid due to `replicas < minReplicas`. Perhaps there is an argument that we should special-case replicas=0 so that users can scale to zero without intervening in the set {min,max}Replicas.

lc525 (commented):
I've raised some questions as to the sane/unsurprising behaviour in particular cases, mostly focusing on the results of the introduced unit test. The approach as a whole looks good.

lc525:
Bringing @domsolutions on as a reviewer, due to overlap with work done on _server_ scale-to-zero behaviour. @domsolutions, would you be able to also comment on what should be the right behaviour from your perspective?

operator/apis/mlops/v1alpha1/model_types_test.go
```diff
@@ -451,6 +451,182 @@ func TestAsModelDetails(t *testing.T) {
...
+				Replicas:    1,
+				MinReplicas: 1,
+				MaxReplicas: 0,
+			},
+			wantErr: "",
+		},
+		{
+			name:        "min replicas scaling",
```
lc525: @domsolutions This is the same approach as the one suggested in #6613

operator/apis/mlops/v1alpha1/model_types_test.go
domsolutions: Yeah I agree, doesn't make much sense having a validated max of 0 and replicas going beyond that, `math.MaxUint32` sounds like a sensible approach

operator/apis/mlops/v1alpha1/model_types_test.go
domsolutions: I think max should also go to `math.MaxUint32`  also

operator/apis/mlops/v1alpha1/model_types_test.go
domsolutions: In my [PR](https://github.com/SeldonIO/seldon-core/pull/6613) I considered this an invalid spec... IMHO it's confusing for the customer what the replica count would validate to be, if they explictiy set the `replicas` field and it resolves to something else without erroring

operator/apis/mlops/v1alpha1/model_types_test.go
domsolutions: Yes, I think this make sense from a common sense perspective, customer explicitly states they want a minimum amount

operator/apis/mlops/v1alpha1/utils.go
```diff
@@ -43,7 +50,7 @@ func GetValidatedScalingSpec(replicas *int32, minReplicas *int32, maxReplicas *i
 	if maxReplicas != nil && *maxReplicas > 0 {
 		validatedSpec.MaxReplicas = uint32(*maxReplicas)
 		if validatedSpec.Replicas > validatedSpec.MaxReplicas {
-			return nil, fmt.Errorf("number of replicas %d must be <= min replicas  %d", validatedSpec.Replicas, validatedSpec.MaxReplicas)
+			return nil, fmt.Errorf("number of replicas %d must be <= max replicas  %d", validatedSpec.Replicas, validatedSpec.MaxReplicas)
```
domsolutions: Can we add a check that `min <= max`

operator/apis/mlops/v1alpha1/model_types_test.go
```diff
@@ -451,6 +451,182 @@ func TestAsModelDetails(t *testing.T) {
...
+				Replicas:    2,
+				MinReplicas: 2,
+			},
+			wantErr: "",
+		},
+		{
+			name:        "max replicas scaling",
+			replicas:    nil,
+			minReplicas: nil,
+			maxReplicas: i32(2),
```
domsolutions: can use `ptr.Int32()` insead

operator/apis/mlops/v1alpha1/model_types_test.go
MiguelAAe: addressed in 0c35cb9ff0ffa2e7d47053c3d820e6304b2493f9

operator/apis/mlops/v1alpha1/model_types_test.go
MiguelAAe: addressed 0c35cb9ff0ffa2e7d47053c3d820e6304b2493f9

operator/apis/mlops/v1alpha1/model_types_test.go
MiguelAAe: added as an error https://github.com/SeldonIO/seldon-core/pull/6557/files#diff-a42dfb1172acd1ddb53a34e6d235a8110c6415a933a99213acb96f41e4a7ebeeR53

operator/apis/mlops/v1alpha1/utils.go
MiguelAAe: added in a38c14c1be4d4cd5471ae42baf86d7124b6e05b4

apis/mlops/scheduler/scheduler.proto
```diff
@@ -156,6 +156,7 @@ message ModelStatus {
       ModelTerminated = 5;
       ModelTerminateFailed = 6;
       ScheduleFailed = 7;
+      ModelScaledDown = 8;
```
tyndria, line 160: **praise**: that's useful to know as we're touching model statuses in our project as well

operator/scheduler/server_test.go
```diff
@@ -46,7 +46,7 @@ func TestServerNotify(t *testing.T) {
 					},
 					Spec: v1alpha1.ServerSpec{
 						ScalingSpec: v1alpha1.ScalingSpec{
-							Replicas: getIntPtr(2),
+							Replicas: ptr.Int32(2),
```
tyndria, line 51: 👍 
@vtaskow `"github.com/gotidy/ptr"` - sth we should start using it as well 👀

scheduler/pkg/store/memory_status.go
```diff
@@ -87,6 +87,8 @@ func updateModelState(isLatest bool, modelVersion *ModelVersion, prevModelVersio
 			modelState = ModelFailed
 			modelReason = stats.lastFailedReason
 			modelTimestamp = stats.lastFailedStateTime
+		} else if modelVersion.GetDeploymentSpec() != nil && stats.replicasAvailable == 0 && modelVersion.GetDeploymentSpec().Replicas == 0 && modelVersion.GetDeploymentSpec().MinReplicas == 0 {
```
tyndria, line 91: **question**: As I'm still learning about Core, I wonder if you could please explain why do we have to do this check `stats.replicasAvailable == 0`? Why this one `modelVersion.GetDeploymentSpec().Replicas == 0 && modelVersion.GetDeploymentSpec().MinReplicas == 0` is not enough to say that model is scaled down?

Thanks 🙏

operator/apis/mlops/v1alpha1/utils.go
```diff
@@ -20,33 +23,49 @@ type ValidatedScalingSpec struct {
 func GetValidatedScalingSpec(replicas *int32, minReplicas *int32, maxReplicas *int32) (*ValidatedScalingSpec, error) {
 	var validatedSpec ValidatedScalingSpec
```
tyndria: **thought/suggestion (non-blocking)**: I wonder if it's valid to set default values (1, 0, MaxUint32) in the beginning of this function and override those if values are provided as an arguments? 

```
validatedSpec := ValidatedScalingSpec{
	Replicas:    1,
	MinReplicas: 0,
	MaxReplicas: math.MaxUint32,
}
```

In this case we would avoid these clauses, however, I'm not sure if it's valid from the K8S perspective:
```
else {
	validatedSpec.MaxReplicas = math.MaxUint32
}
```

tyndria: **P.S.** I've realised that according to my proposal the original test case would have to be implemented **explicitly** without relying on initial zero values.

scheduler/pkg/store/memory_status.go
MiguelAAe: There is a small gap in which you can have a model to be in the desire state of having 0 replicas and still have replicas in the server. We need to ensure that our desire state matches the one present in the cluster. 

for example you could scale down your model to 0 replicas but it may take some time for some processes to kick in and scaled the model to 0. during that time the model is still active.  Hence a model isn't scaled down until its current state matches user desire state.

tyndria: Got it, thanks!

operator/apis/mlops/v1alpha1/utils.go
MiguelAAe: It is possible to set values explicitly and still pass all test cases, below is with more explicitly but It feels like the code is a bit more complicated, any thoughts @domsolutions 

```go
func GetValidatedScalingSpec(replicas *int32, minReplicas *int32, maxReplicas *int32) (*ValidatedScalingSpec, error) {
	validatedSpec := ValidatedScalingSpec{
		Replicas:    1,
		MinReplicas: 0,
		MaxReplicas: math.MaxUint32,
	}

	if replicas == nil && minReplicas == nil && maxReplicas == nil {
		return &validatedSpec, nil
	}

	if replicas != nil && *replicas > 0 {
		validatedSpec.Replicas = uint32(*replicas)
	} else if replicas != nil && *replicas == 0 && minReplicas != nil && *minReplicas == 0 {
		validatedSpec.Replicas = uint32(0)
	} else if minReplicas != nil && *minReplicas > 0 {
		if replicas != nil && *replicas < *minReplicas {
			return nil, fmt.Errorf("number of replicas %d cannot be less than minimum replica %d", *replicas, *minReplicas)
		}
		// set replicas to the min replicas when replicas is not set explicitly
		validatedSpec.Replicas = uint32(*minReplicas)
	}

	if minReplicas != nil && *minReplicas > 0 {
		validatedSpec.MinReplicas = uint32(*minReplicas)
		if validatedSpec.Replicas < validatedSpec.MinReplicas {
			return nil, fmt.Errorf("number of replicas %d must be >= min replicas %d", validatedSpec.Replicas, validatedSpec.MinReplicas)
		}
	}

	if maxReplicas != nil && *maxReplicas > 0 {
		validatedSpec.MaxReplicas = uint32(*maxReplicas)
		if validatedSpec.MinReplicas > validatedSpec.MaxReplicas {
			return nil, fmt.Errorf("min number of replicas %d must be <= max number of replicas %d", validatedSpec.MinReplicas, validatedSpec.MaxReplicas)
		}
		if validatedSpec.Replicas > validatedSpec.MaxReplicas {
			return nil, fmt.Errorf("number of replicas %d must be <= max replicas %d", validatedSpec.Replicas, validatedSpec.MaxReplicas)
		}
	}

	return &validatedSpec, nil
}
```

domsolutions: IMHO I think it's a good idea, makes it immediatly clear what the default values are w/o trying to figure out what the code is doing.. I'd probably suggest having a seprate private func to provide the default values to make it even more clear... but think this is a minor thing so happy with whatever you decide.

domsolutions (approved):
LGTM

operator/apis/mlops/v1alpha1/utils_test.go
```diff
@@ -0,0 +1,165 @@
...
+		{
+			name:        "error - min replica is bigger than max replicas",
+			replicas:    ptr.Int32(6),
+			minReplicas: ptr.Int32(6),
+			maxReplicas: ptr.Int32(4),
+			expected:    nil,
+			wantErr:     "min number of replicas 6 must be <= max number of replicas 4",
+		},
+		{
+			name:        "success - replicas stays at 0 when min replicas and max replicas is 4",
```
lc525:
```suggestion
			name:        "success - replicas stays at 0 when min replicas is 0 and max replicas is 4",
```

operator/apis/mlops/v1alpha1/utils_test.go
```diff
@@ -0,0 +1,165 @@
...
+			maxReplicas: nil,
+			expected: &ValidatedScalingSpec{
+				Replicas:    1,
+				MinReplicas: 0,
+				MaxReplicas: math.MaxUint32,
+			},
+			wantErr: "",
+		},
+		{
+			name:        "success - unset replica params defaults to 1 when min replicas is set",
```
lc525: Suggestion: this test could target the case where unset replica params defaults to minReplicas when minReplicas is set. I suggest having minReplicas set to 2 to showcase that. It would also be set to 0 if minReplicas is 0, and that can be another test case.

operator/apis/mlops/v1alpha1/utils_test.go
```diff
@@ -0,0 +1,165 @@
...
+			maxReplicas: nil,
+			expected: &ValidatedScalingSpec{
+				Replicas:    1,
+				MinReplicas: 1,
+				MaxReplicas: math.MaxUint32,
+			},
+			wantErr: "",
+		},
+		{
+			name:        "success - unset min and replicas defaults to 1 when max replicas is set",
```
lc525:
```suggestion
			name:        "success - unset minReplicas defaults to 0, unset replicas defaults to 1 when max replicas is set",
```

scheduler/docs/design.md
```diff
@@ -60,14 +60,15 @@ A model state can be in a set of states.
...
-	ScheduleFailed
+    ModelStateUnknown ModelState = iota
+    ModelProgressing
+    ModelAvailable
+    ModelFailed
+    ModelTerminating
+    ModelTerminated
+    ModelTerminateFailed
+    ScheduleFailed
+    ModelScaledDown
```
lc525, line 71: Those docs changes are only tangentially related to the code changes (namely, just updating with the added state). However, let's keep them in this PR because they are very useful. Thank you for creating the diagram!

operator/apis/mlops/v1alpha1/utils_test.go
MiguelAAe: by default we always set replicas to 1 when replicas is not set

lc525: but the code [here](https://github.com/SeldonIO/seldon-core/pull/6557/files#diff-a1f4f216e2d18cbbaaf58eed2d66c27d4da91a960ccdf8942f8b59dce84c73b1R42-R43) seems to suggest otherwise and set it to minReplicas. If replicas is not set and minReplicas is set (the case in the test), the PR is setting replicas to minReplicas

PR #6560, opened 2025-06-26
feat(Agent): add readiness probes

lc525:

At the moment, using Core 2 in conjunction with k8s Pod Disruption Budgets (PDBs) doesn't work as expected, because on the inference server pod side, the pod is considered "Ready" from a PDB perspective while the agent is in draining mode.

This means that while we cleanly tear down an inference server (blocking new connections but continuing to serve all outstanding requests), PDB may continue to terminate other inference server replicas without affecting the disruption budget.

The consequence of this is that in the extreme, all inference server replicas serving particular models may be terminated (leading to unavailability), while also fulfilling all the PDB constraints.

As part of the changes, we refactor Agent code for readability and improve the way we wait and deal with internal and external dependent services.

Fixes #INFRA-1392 (Add agent readiness probes)

Please review per-commit and read the commit log message before reviewing, because each commit is self-standing.


scheduler/pkg/agent/agent_debug.go
```diff
@@ -47,68 +47,68 @@ func NewAgentDebug(logger log.FieldLogger, port uint) *agentDebug {
 	}
```
vtaskow, line 49: Line 43 lights up for me due to the return type not being exported for an exported function

scheduler/pkg/agent/agent_debug.go
```diff
@@ -123,6 +123,6 @@ func (cd *agentDebug) ReplicaStatus(ctx context.Context, r *pbad.ReplicaStatusRe
 		i++
 	}
 	return &pbad.ReplicaStatusResponse{
-		AvailableMemoryBytes: uint64(cd.stateManager.GetAvailableMemoryBytes()),
+		AvailableMemoryBytes: uint64(ad.stateManager.GetAvailableMemoryBytes()),
```
vtaskow: redundant type conversion

scheduler/pkg/agent/client_test.go
```diff
@@ -107,7 +107,7 @@ func (f FakeDependencyService) Name() string {
 
 func addVerionToModels(models []string, version uint32) []string {
 	modelsWithVersions := make([]string, len(models))
-	for i := 0; i < len(models); i++ {
+	for i := range len(models) {
```
vtaskow, line 123: :+1:

scheduler/pkg/agent/agent_svc_manager.go
```diff
@@ -273,84 +295,138 @@ func (am *AgentServiceManager) createConnection() error {
...
+		}()
 	}
 
-	// agent debug service
-	if err := isReadyChecker(isStartup, am.agentDebugService, logger, "Agent debug service not ready",
-		maxElapsedTime,
-	); err != nil {
-		return err
-	}
+	go func() {
```
vtaskow, line 408: question: I believe it's possible to unblock here when none of the internal sub services are ready in a case when add 2 + both model repo and inference server are ready so the wait group counter will become zero, is that right? would it make more sense to add 2 + len(internalSubServices) first and then start all of them?

scheduler/pkg/agent/agent_svc_manager.go
```diff
@@ -273,84 +295,138 @@ func (am *AgentServiceManager) createConnection() error {
...
-	); err != nil {
-		return err
+	criticalSubservicesNotReady := make([]string, 0)
+	for notification := range readyNotifications {
+		if notification.err != nil {
+			logger.WithError(notification.err).Errorf("Giving up on waiting for %s service to become ready; service marked as failed", notification.subserviceName)
+			if am.isCriticalSubserviceFailure(isStartup, notification.subServiceType) {
+				criticalSubservicesNotReady = append(criticalSubservicesNotReady, notification.subserviceName)
+			} else {
+				logger.Warnf("Non-critical subservice no longer ready: %s", notification.subserviceName)
```
vtaskow: nit: this suggests that a subservice was ready but is it wrong to say that when it's on a start up? same question for #L423

scheduler/pkg/agent/agent_svc_manager.go
lc525: But this goroutine and the wait can only execute _after_ `wg.Add(len(internalSubServices))` was called, right? So even if the counter becomes 0 before the second add, this should still behave as expected. However, I agree it's better to just add before starting any goroutines, to avoid confusion, so I'll do so.

scheduler/pkg/agent/agent_debug.go
lc525: I suspect it wasn't redundant at some point, but then `GetAvailableMemoryBytes` got refactored. Will fix.

scheduler/pkg/agent/agent_svc_manager.go
lc525: Not sure I understood this. We're under the `if notification.err != nil` branch, so we know this is a service that failed to start. We then check whether that service is "critical" according to the stage we're in (where critical means something different at startup compared to normal checks). If the error is not critical, (but it's still an error!), we log that.

scheduler/pkg/agent/agent_svc_manager.go
vtaskow: Ah, you are right, my bad!

scheduler/pkg/agent/agent_svc_manager.go
lc525: will fix (following discussion)

PR #6600, opened 2025-07-08
feat(pipelinegw): pipelinegw scalability to number of partitions

RobertSamoilescu:
# Why
## Issues

In the current implementation, `pipelinegw` can scale to any number of replicas. This is possible because when a request is sent through the load balancer to one of the replicas, the pipeline is created on the first hit, and the associated output topic is added to a consumer belonging to a uniquely identified consumer group.

However, this approach is inefficient: each consumer in the unique consumer group reads messages from the same Kafka pipeline output topic, but ends up ignoring most of them—processing only those messages that correspond to the requests sent by its respective `pipelinegw` replica.

## Motivation

Enable efficient scalability for `pipelinegw` — this is the first step in that direction. A following PR will implement pipeline rebalancing among `pipelinegw` replicas. 

# What
## Summary of changes

This PR introduces a more efficient scaling mechanism for `pipelinegw`, up to the number of partitions. Key changes include:

- Consumers are now created when the pipeline is loaded, rather than on each request.
- On-request loading is still supported for models, primarily for benchmarking purposes.
- All consumers for a given pipeline belong to the **same** consumer group. This avoids the issue of fetching and discarding unrelated Kafka messages.
- Partition assignment and rebalancing are handled via Kafka's rebalance callbacks.
- Each `requestId` includes the partition from which the final response should be fetched. This ensures that both `modelgw` and `dataflow-engine` read from and write to the same partition that was chosen by `pipelinegw` when the message was originally written to the input topic.
- In `modelgw`, we write the model response to the same partition from which the request was read.
- In `dataflow-engine`, we determine the correct partition to write to based on the `requestId`, since the `StreamPartitioner` has access only to the key and value.
- `client.go` file was moved to avoid circular imports

## Checklist

### DIFF
--- a/operator/controllers/reconcilers/seldon/runtime_reconciler.go
+++ b/operator/controllers/reconcilers/seldon/runtime_reconciler.go
@@ -32,6 +32,10 @@ type SeldonRuntimeReconciler struct {
 }
 
 func ParseInt32(s string) (int32, error) {
+	if s == "" {
+		return 1, nil
+	}
+
 	i64, err := strconv.ParseInt(s, 10, 32)
 	return int32(i64), err
 }


scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -249,10 +249,10 @@ func (iw *InferWorker) produce(
 			logger.Debugf("Adding kafka header for topic %s %s:%s", topic, h.Key, string(h.Value))
 		}
 	}
-	logger.Infof("Produce response to topic %s", topic)
+	logger.Infof("Produce response to topic %s on partition %d", topic, job.msg.TopicPartition.Partition)
```
lc525: This should probably be `Debugf`

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -197,14 +215,52 @@ func (km *KafkaManager) Infer(
...
+	partition := partitions[rand.Intn(len(partitions))]
+	logger.Debugf("Using partition %d for resource %s", partition, resourceName)
+
 	// Use composite key to differentiate multiple pipelines (i.e. mirror) using the same message
-	compositeKey := getCompositeKey(resourceName, requestId, ".")
+	// Note that we add the partition to the key to ensure that the message will be sent to
+	// a partition for which the consumer is subscribed. For modelgw, it is enought to send the
+	// message to the same partition as the one we read from. For dataflow engine on the other hand,
+	// we need to read the partition from the request id.
+	compositeKey := getCompositeKey(strconv.Itoa(int(partition)), resourceName, requestId, ".")
```
lc525, line 238: This is an elegant way out of the fundamental problem of each pipeline-gw needing to be both a producer into a partition and a consumer from the same exact partition, whith dataflow-engine controlling this latter partition no.

scheduler/pkg/kafka/pipeline/multi_topic_consumer.go
```diff
@@ -56,13 +58,17 @@ func NewMultiTopicsKafkaConsumer(
...
-	err := consumer.createConsumer()
+	err := consumer.createConsumer(logger)
 	return consumer, err
 }
 
-func (c *MultiTopicsKafkaConsumer) createConsumer() error {
+func (c *MultiTopicsKafkaConsumer) createConsumer(logger log.FieldLogger) error {
 	consumerConfig := kafka_config.CloneKafkaConfigMap(c.config.Consumer)
 	consumerConfig["group.id"] = c.id
+	consumerConfig["go.events.channel.enable"] = true
```
lc525: This channel should not be enabled at the same time with the Poll() method. There will be cases where the Poll simply looses events if this setting is on (because the corresponding events were sucked into the `events.channel`)

scheduler/pkg/kafka/pipeline/multi_topic_consumer.go
```diff
@@ -56,13 +58,17 @@ func NewMultiTopicsKafkaConsumer(
...
+	consumerConfig["go.application.rebalance.enable"] = true // Required for Events()
+	consumerConfig["partition.assignment.strategy"] = "roundrobin"
```
lc525, line 68: Have you considered [stickyAssignor](https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/consumer/StickyAssignor.html)? We should look into it more broadly

scheduler/pkg/kafka/pipeline/multi_topic_consumer.go
```diff
@@ -141,10 +147,17 @@ func (c *MultiTopicsKafkaConsumer) subscribeTopics(cb kafka.RebalanceCb) error {
 	return c.consumer.SubscribeTopics(topics, cb)
 }
 
+func extractRequestId(partitionAndRequestId string) string {
+	parts := strings.SplitN(partitionAndRequestId, "|", 2)
```
lc525: Wasn't the separator ["."](https://github.com/SeldonIO/seldon-core/pull/6600/files#diff-7de2207a5b2f5c03825af0697582778bf69507adff4abf12661f3f774bb8369cR263) here?

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -267,13 +324,48 @@ func createResponseErrorPayload(modelName string, response []byte) []byte {
...
+			pipeline.consumer.partitions = make([]int32, len(e.Partitions))
+			for i, partition := range e.Partitions {
+				pipeline.consumer.partitions[i] = partition.Partition
+			}
+		case kafka.RevokedPartitions:
+			km.mu.Lock()
+			km.mu.Unlock()
+
+			logger.Debug("Rebalance: Revoked partitions:", e.Partitions)
+			err := consumer.Unassign()
```
lc525, line 327: Here, aren't we assuming that all partitions have been unassigned? (rather than just e.Partitions?)

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -157,35 +160,50 @@ func getPipelineKey(resourceName string, isModel bool) string {
...
-				}
-			}()
-		}
+func (km *KafkaManager) LoadPipeline(resourceName string, isModel bool) (*Pipeline, error) {
+	km.mu.RLock()
+	defer km.mu.RUnlock()
+	return loadPipeline(resourceName, isModel, &km.pipelines)
+}
+
+func (km *KafkaManager) StorePipeline(resourceName string, isModel bool) error {
```
lc525: This is much more than just "Store", because it essentially does "LoadOrCreateConsumersAndStore". Given that we can wait on the Consumers via the wg, could this be turned into one function that uses the standard `LoadOrStore` operation, plus the logic for creating the consumers and them subscribing to the right topics?

lc525 (approved):
I think the overall approach is solid, but I've suggested a couple refinements and small fixes.

scheduler/pkg/kafka/pipeline/kafkamanager.go
RobertSamoilescu: We are dealing only with partitions per consumers ... or maybe I don't understand the question ...

scheduler/pkg/kafka/pipeline/multi_topic_consumer.go
RobertSamoilescu: Will leave it as `roundrobin` for now, since `sticky` give the following error:
```
            Unsupported partition.assignment.strategy: sticky
            {
                code: -186,
                str: "Unsupported partition.assignment.strategy: sticky",
                fatal: false,
                retriable: false,
                txnRequiresAbort: false,
            }

```

scheduler/pkg/kafka/pipeline/multi_topic_consumer.go
RobertSamoilescu: Yes. Forgot to updated it when I changed to ".". Thank you!

PR #6613, opened 2025-07-15
Infra 1492/scheduler waits for 0 replicas

domsolutions:
# Why

There is a bug with the scaling specification for Server CRs:
 - If Server spec looked like (where `replicas` is not set):
 ```yaml
spec:
  minReplicas: 2
```
 the controller would notify the scheduler to wait for 2 servers (agents) to connect to it. But the controller would only deploy 1 replica, so the scheduler would wait until the max timeout of 10 minutes. The fix now ensures the recplias count == minReplica if replicas is not set and deploys the correct amount of StatefulSets.

## Issues

There was no way of adding some test cases to the integration tests as there was a tight coupling between the scheduling client implementation and the controller. Meaning the scheduler was also need to be part of the integrationn test. To address this, changed the `schedulerClient` to implement a `Client` interface and changed the controller to depend on `Client` instead of `schedulerClient`. From doing this, was able to generate mocks of the scheduler client and mock the responses on the tests.

# What
## Summary of changes

- Within `Reconcile` of the controller, the scaling spec is validated the the server replica count is now set to the calculated replica number from validating the spec and taking the min/max into account

## Checklist

### CODE REVIEW

**Comment by domsolutions** [comment]
Lets merge this [PR](https://github.com/SeldonIO/seldon-core/pull/6557/files) first, as there will be conflicts, and I'll rebase and resolve the conflicts

File: operator/Makefile [hunk 1]
```diff
@@ -85,7 +85,7 @@ lint: ${.GOLANGCILINT_PATH}/golangci-lint
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" POD_NAMESPACE=default go test ./... -coverprofile cover.out
```
lc525: why was this change required?

operator/scheduler/client.go
```diff
@@ -59,13 +79,13 @@ type SchedulerClient struct {
 // For this will need to know ports (hardwire for now to 9004 and 9044 - ssl comes fom envvar - so always
 // the same for all schedulers
 
-func NewSchedulerClient(logger logr.Logger, client client.Client, recorder record.EventRecorder) *SchedulerClient {
+func NewSchedulerClient(logger logr.Logger, client client.Client, recorder record.EventRecorder) Client {
```
lc525: As a matter of style/approach (we can discuss if you have strong opinions one way or the other): I would rather return the concrete type here, but keep (and define) the interface on the Operator side (the "consumer" of this object). This is because in some more advanced integration tests we will want an actual `SchedulerClient` rather than a mock. This is the same approach as discussed [here](https://go.dev/wiki/CodeReviewComments#interfaces), but of course if there are good reasons to return the interface we could do so. I just don't necessarily see them here.

lc525: We do use (manually generated) mocks for parts of the scheduler in other tests (e.g. see [here](https://github.com/SeldonIO/seldon-core/blob/v2/scheduler/pkg/agent/server_test.go)). Just leaving this here as a reference, we'll want to aim for a consistent mock generation across the codebase where possible (it's probably wise to move to auto-generated mocks).

operator/Makefile
domsolutions: It was needed due to

```golang
var (
	SeldonNamespace = getEnvOrDefault("POD_NAMESPACE", "seldon-mesh")
)
```

Where if `POD_NAMESPACE` is not set, the namespace is set to `seldon-mesh`, but in the test we're using `default` namespace from

```golang
const (
	TestNamespace = "default"
)
```

So we get errors as it can't find the server spec as it's trying to retrieve it from a different ns. 

I do wonder if this highlights an issue though, as `SeldonNamespace` is used predominantly in tests and only in 2 non-test places (wonder if it was created for tests and accidentally leaked into actual code):

```golang
func GetSeldonConfigForSeldonRuntime(seldonConfigName string, client client.Client) (*SeldonConfig, error) {
	if seldonConfigName == "" {
		return nil, fmt.Errorf("SeldonConfig not specified and is required")
	}
	sc := SeldonConfig{}
	err := client.Get(context.TODO(), types.NamespacedName{Name: seldonConfigName, Namespace: constants.SeldonNamespace}, &sc)
	return &sc, err
}

func GetServerConfigForServer(serverConfig string, client client.Client) (*ServerConfig, error) {
	if serverConfig == "" {
		return nil, fmt.Errorf("ServerType not specified and is required")
	}
	sc := ServerConfig{}
	err := client.Get(context.TODO(), types.NamespacedName{Name: serverConfig, Namespace: constants.SeldonNamespace}, &sc)
	if err != nil {
		return nil, fmt.Errorf("failed to get ServerConfig: %w", err)
	}
	return &sc, err
}

```

I wonder if it should be hardcoded like this, as from looking at the operator's config, it supports monitoring cluster-wide i.e. multiple namespaces?

operator/scheduler/client.go
domsolutions: > We do use (manually generated) mocks for parts of the scheduler in other tests (e.g. see [here](https://github.com/SeldonIO/seldon-core/blob/v2/scheduler/pkg/agent/server_test.go)). Just leaving this here as a reference, we'll want to aim for a consistent mock generation across the codebase where possible (it's probably wise to move to auto-generated mocks).

I did notice those, I'd say it's pretty much standard Go practice to use auto-generated mocks for a few reasons:
- any change to the interface can easily be added to the mock just by running `go generate`
- probably most importantly, we have the ability to control what the mock returns for each call, and we can verify that it has been called with the correct params and the correct number of times etc... you could probably achieve that as well writing your own mocks, but I can't see a good reason for it

I agree we should have a standard approach and probably look at removing the old hand-generated mocks and upgrade the tests to use the auto-gen mocks.

domsolutions: > As a matter of style/approach (we can discuss if you have strong opinions one way or the other): I would rather return the concrete type here, but keep (and define) the interface on the Operator side (the "consumer" of this object). This is because in some more advanced integration tests we will want an actual SchedulerClient rather than a mock. This is the same approach as discussed [here](https://go.dev/wiki/CodeReviewComments#interfaces), but of course if there are good reasons to return the interface we could do so. I just don't necessarily see them here.

Yeah I think you're right, previous projects I've worked on we used [uberfx](https://uber-go.github.io/fx/index.html) DI framework and it'd require the constructor to return an interface so it would know how to wire up the app correctly, But the approach your suggesting allows the code to be less brittle to change, and doesn't restrict the scheduler client by bounding it to the interface, so will go with that.

operator/Makefile
lc525: Albeit nonintuitive and in need of better docs, I think the current behaviour is correct. Namely, the `SeldonConfig` and `ServerConfig` CRs need to be in the same namespace as the controller, irrespective of whether the controller is cluser-wide, in its own namespace, or in the same namespace as the "runtime". 

The point is that those overall configs can be centrally managed in the namespace of the controller. Then, users of core in each namespace can customise the settings (and override any needed pieces of the configs in `SeldonRuntime` and `Server` CRs.

So when a `SeldonRuntime` references a `SeldonConfig` by name (i.e `spec.seldonConfig`) we need to search that name in the namespace of the controller. Same with Servers.

operator/apis/mlops/v1alpha1/utils.go
```diff
@@ -20,6 +20,8 @@ type ValidatedScalingSpec struct {
 	MaxReplicas uint32
 }
 
+// TODO consider putting these rules via CEL into the CR definition for Server, this will allow customers earlier feedback
+// their CRs are invalid upon trying to apply them, instead of checking controller logs after applying.
```
lc525, line 24: The validation errors get returned and they are part of the reconcile loop, with the status of the CR being set accordingly (for example [here](https://github.com/SeldonIO/seldon-core/blob/c330727c2881c5a2b835f795bb92a320e5e26469/operator/controllers/mlops/server_controller.go#L101)) to the encountered error. So not as bad as needing to check controller logs, but I agree it would still be better as CEL validation rules and rejecting the update outright

scheduler/pkg/agent/k8s/mocks/mock_discovery_v1.go
```diff
@@ -1,12 +1,3 @@
-/*
-Copyright (c) 2024 Seldon Technologies Ltd.
-
-Use of this software is governed BY
-(1) the license included in the LICENSE file or
-(2) if the license included in the LICENSE file is the Business Source License 1.1,
-the Change License after the Change Date as each is defined in accordance with the LICENSE file.
-*/
-
 // Code generated by MockGen. DO NOT EDIT.
```
lc525: Previously the mock files had the project license. Given that these files are generated rather than written by us, I'm ambivalent whether we should include the license here, but could do so for consitency. Mostly trying to figure out whether we forgot to automatically add the license to the generated files, or this was intentional.

PR #6614, opened 2025-07-15
enable race detector

domsolutions:
# Why
## Motivation

# What
## Checklist


hodometer/Makefile
```diff
@@ -60,11 +60,11 @@ build: build-hodometer build-receiver
 
 .PHONY: test-hodometer
 test-hodometer:
-	$(GO_ENV) go test $(VERBOSE) ./cmd/hodometer/... ./pkg/hodometer/...
+	$(GO_ENV) go test $(VERBOSE) -race./cmd/hodometer/... ./pkg/hodometer/...
```
lc525:
```suggestion
	$(GO_ENV) go test $(VERBOSE) -race ./cmd/hodometer/... ./pkg/hodometer/...
```

lc525 (commented):
Added missing space as suggestion. Do you have a rough approximation wrt how much this slows down the tests? We run those make targets in the github CI and would want to keep the runtime small. 

If the delta is large, we should still run the tests with the race detector on, but I propose we only enable it if an environment variable is set and run it locally before rising PRs (until we're setting up local github runners).

operator/apis/mlops/v1alpha1/experiment_types_test.go
```diff
@@ -23,6 +23,8 @@ import (
 )
 
 func TestAsSchedulerExperimentRequest(t *testing.T) {
+	t.Parallel()
```
lc525, line 26: In general (not specific to this line or test case), tests where the internal code being tested relies on the event hub will require a new event hub instance per test case, or will most likely fail (because they might read the events corresponding to a different test case from the hub).

domsolutions: Yeah I did notice that, especially within the scheduler tests, I only applied parallel to the operator tests where it appeared the test cases didn't depend on each other.

domsolutions:
Review after pipleline scalability tests are complete to make it easier to isolate issues.

lc525 (approved):
lgtm, thank you for replying to my initial comments. I think the time trade-off of having the race detector on is acceptable.

PR #6619, opened 2025-07-18
feat: OpenAI API REST translation to OIP

RobertSamoilescu:
# Why
## Motivation

# What
## Checklist


scheduler/pkg/agent/rproxy.go
```diff
@@ -68,11 +77,25 @@ func addRequestIdToResponse(req *http.Request, res *http.Response) {
 	}
 }
 
+func (t *lazyModelLoadTransport) TranslateToOIP(req *http.Request, termination string) (*http.Request, error) {
```
lc525: This function acts like a "translate if needed" rather than "always translate". Therefore, I'd suggest names like `TranslateToOIPIfNeeded` and `TranslateFromOIPIfNeeded` to better reflect the behavior.

scheduler/pkg/agent/rproxy.go
```diff
@@ -29,10 +29,18 @@ import (
...
+	"github.com/seldonio/seldon-core/scheduler/v2/pkg/agent/translator"
+	"github.com/seldonio/seldon-core/scheduler/v2/pkg/agent/translator/openai"
 	"github.com/seldonio/seldon-core/scheduler/v2/pkg/metrics"
 	"github.com/seldonio/seldon-core/scheduler/v2/pkg/util"
 )
 
+const (
+	chatCompletionsPath   = "/chat/completions"
+	embeddingsPath        = "/embeddings"
+	imagesGenerationsPath = "/images/generations"
```
lc525, line 41: Because of using those custom paths, this works for inferences sent to models directly, but not for pipelines. Correct?

scheduler/pkg/agent/translator/base.go
```diff
@@ -0,0 +1,193 @@
...
+	mapContent := map[string]interface{}{}
+	err = json.Unmarshal([]byte(content), &mapContent)
+	if err != nil {
+		return "", fmt.Errorf("failed to unmarshal content: %w", err)
+	}
+
+	contentBytes, err := json.Marshal(mapContent)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal content: %w", err)
+	}
```
lc525: Why is this needed? Presumably if `json.Unmarshal` works, the content is valid-formatted JSON (I'm sure I'm missing something here, but I don't know what)

scheduler/pkg/agent/translator/openai/chat_completions.go
```diff
@@ -0,0 +1,562 @@
...
+
+		// Add parallel_tool_calls if present
+		if len(parallelToolCalls) > 0 {
+			strParallelToolCalls, err := marshalListContent(parallelToolCalls)
+			if err != nil {
+				return nil, fmt.Errorf("failed to marshal %s content: %v", parallelToolCallsKey, err)
+			}
+			inferenceRequestInputs = append(
+				inferenceRequestInputs,
+				translator.ConstructStringTensor(parallelToolCallsKey, strParallelToolCalls),
```
lc525: For OIP, wouldn't we want each tool call from parallel tool calls as a separate tensor?

scheduler/pkg/agent/translator/openai/chat_completions.go
```diff
@@ -0,0 +1,562 @@
...
+				translator.ConstructStringTensor(toolChoiceKey, strToolChoice),
+			)
+		}
+	}
+
+	return map[string]interface{}{
+		inputsKey: inferenceRequestInputs,
+		// There is an inconsistency in the naming of the parameters field
+		// across the runtimes. The API runtime uses `llm_parameters` for all
+		// model types (not just LLMs), while local runtime uses `kwargs`
```
lc525: Would it be simpler to just solve the inconsistency in the two runtimes (i.e pick one)? Is there a fundamental reason why the local runtime would be better with `kwargs`?

scheduler/pkg/agent/translator/openai/chat_completions_test.go
```diff
@@ -0,0 +1,1115 @@
...
+						},
+					},
+					{
+						"name":     "tool_calls",
+						"shape":    []int{7},
+						"datatype": "BYTES",
+						"data": []string{
+							"",
+							"",
+							"[\"{\\\"function\\\":\\\"foo_function\\\",\\\"id\\\":\\\"foo_id\\\",\\\"type\\\":\\\"function\\\"}\"]",
```
lc525: (somewhat unrelated to this PR): There's something strange/wrong with the StringCodec and the way we're using it to serialize json. Escaping json should not require an additional level of escaping for each additional json nesting level.

scheduler/pkg/agent/translator/openai/chat_completions_test.go
```diff
@@ -0,0 +1,1115 @@
...
+				},
+				"parameters": map[string]interface{}{
+					"llm_parameters": map[string]interface{}{},
+					"kwargs":         map[string]interface{}{},
+				},
+			},
+		},
+		{
+			name: "image-input",
+			openAIContent: map[string]interface{}{
```
lc525: It would be cleaner to replace all of those `interface{}` with `any`.

scheduler/pkg/agent/translator/base.go
```diff
@@ -0,0 +1,193 @@
+package translator
```
lc525: Suggest to move the entire translator package outside `agent/` and into `scheduler/pkg/`

scheduler/pkg/agent/translator/openai/chat_completions.go
```diff
@@ -0,0 +1,562 @@
...
+	parametersKey        = "parameters"
+	llmParametersKey     = "llm_parameters"
+	localParametersKey   = "kwargs"
+	modelNameKey         = "model_name"
+	idKey                = "id"
+)
+
+func (t *OpenAIChatCompletionsTranslator) TranslateToOIP(req *http.Request) (*http.Request, error) {
+	// Convert OpenAI API request to JSON
+	jsonBody, err := translator.ConvertRequestToJsonBody(req)
```
lc525: looks like on receiving the request, we don't support compression (and assume the request is plaintext). However, we do support compression on the response side. Is there a reason for this assymetry? (not sure of the support here before the translation layer work)

scheduler/pkg/agent/rproxy.go
```diff
@@ -102,16 +125,28 @@ func (t *lazyModelLoadTransport) RoundTrip(req *http.Request) (*http.Response, e
...
+	// Translate the request to OIP format if needed
+	termination, err := translator.GetPathTermination(req)
+	if err != nil {
+		t.logger.WithError(err).Warn("Failed to get path termination for request")
+		return nil, err
 	}
+	req, err = t.TranslateToOIP(req, termination)
 	if err != nil {
+		t.logger.WithError(err).Warn("Failed to translate request to OIP format")
+		return nil, err
```
lc525: For RoundTripper, an err return should be used only to indicate the inability to generate a response (including a response with an error status code), not a logical failure in the processing done for the response. So it would be useful for any errors that are worth exposing to the end-user to be returned in HTTP responses with error status codes rather than just logging them locally. (see docs for RoundTripper.RoundTrip)

lc525 (commented):
lgtm; Some questions around:

scheduler/pkg/agent/translator/base.go
RobertSamoilescu: You are correct.

scheduler/pkg/agent/translator/openai/chat_completions.go
RobertSamoilescu: This is different. This is just a parameter to activate parallel tool calling (i.e., calling multiple tools at the same time). This can be set to "true" or "false". Don't remember why this formatting was chosen, but that's what the LLM module is expecting.

scheduler/pkg/agent/translator/openai/chat_completions.go
RobertSamoilescu: There is an inconsistency in the LLM module across multiple runtimes. For example, API runtime uses `"llm_parameters"` whether we are talking about a chat completion task or embedding task, Local runtime uses `"kwargs"`, Local Embeddings runtime uses `"embedding_parameters"`. I agree that those inconsistencies should be solved in the LLM module, but there is some effort to it (update runtime, update docs, release). For now I will keep it as it is.

scheduler/pkg/agent/translator/openai/chat_completions.go
RobertSamoilescu: From what I am aware, the gzip middleware is only available for responses in FastAPI (see [here](https://fastapi.tiangolo.com/advanced/middleware/#gzipmiddleware))

scheduler/pkg/agent/translator/base.go
```diff
@@ -183,11 +183,7 @@ func translateLine(line string) (string, error) {
 		return "", fmt.Errorf("failed to unmarshal content: %w", err)
 	}
 
-	contentBytes, err := json.Marshal(mapContent)
-	if err != nil {
-		return "", fmt.Errorf("failed to marshal content: %w", err)
-	}
-
 	// Reconstruct the line with the original SSE format
+	contentBytes, _ := json.Marshal(mapContent)
```
lc525: So the question here was less about the `Marshal` failing, but about the transformation:

`content` ------> `json.Unmarshal` --------> `mapContent`  -------> `json.Marshal` --------> `contentBytes`

why can't we use `content` directly in the Sprintf below?

RobertSamoilescu: We can. I think I forgot to remove it at some point when I was testing. Thanks for pointing it out

PR #6636, opened 2025-07-31
fix(agent): model availability during inference pod deletion  

domsolutions:
## Motivation

When we have 2 server replicas deployed and we have 2 models **iris-1** and **iris-2** and **iris-1** is deployed to both servers and **iris-2** is deployed to one of the replicas, if we delete one of the server pods:
- delete pod containing only **iris-1**: agent notifies sheduler during draining process, scheduler notifies envoy to stop sending reqs to this pod, pod shutsdown, and no inference reqs should fail. Once replica spins up again, sheduler should schedule model onto this replica
- delete pod  containg both models: agent notifies scheduler during draining process, scheduler moves **iris-2** to other server replica, notfies envoy to stop sending reqs to terminated pod and to send reqs to other pod for **iris-2** reqs. Once replica spins up again, will schedule models again and notify envoy of new routes via scheduler

No inference reqs should fail during this process but instead we get  many connection timeout 503 errors when running a load test of 10 reqs/s for both models via:

```shell
CONSTANT_RATE_DURATION_SECONDS=6000 REQUEST_RATE=10 MODEL_NAME="iris-1" MODEL_TYPE="iris" INFER_HTTP_ENDPOINT="http://localhost:8000" k6 run scenarios/model_constant_rate.js

CONSTANT_RATE_DURATION_SECONDS=6000 REQUEST_RATE=10 MODEL_NAME="iris-2" MODEL_TYPE="iris" INFER_HTTP_ENDPOINT="http://localhost:8000" k6 run scenarios/model_constant_rate.js
```

Model/server specs:

```yaml
kind: Model
metadata:
  name: iris-1
  namespace: seldon-mesh
spec:
  replicas: 2
  requirements:
  - sklearn
  storageUri: gs://seldon-models/scv2/samples/mlserver_1.3.5/iris-sklearn


kind: Model
metadata:
  name: iris-2
  namespace: seldon-mesh
spec:
  replicas: 1
  requirements:
  - sklearn
  storageUri: gs://seldon-models/scv2/samples/mlserver_1.3.5/iris-sklearn


kind: Server
metadata:
  name: mlserver
  namespace: seldon-mesh
spec:
  maxReplicas: 2
  minReplicas: 1
  replicas: 2
  serverConfig: mlserver
    whenDeleted: Retain
    whenScaled: Retain
```

There appears to be multiple problems contributing to this issue:

1. When the inference pod received the `ctrl+c` signal, there was a panic on closing an already closed channel, this causes the whole graceful shutdown to be skipped, including gracefulling shutting down the HTTP server which receives the inference reqs, which would cause inference reqs to error 
2. The draining process is triggered via a `preStop` hook which calls `/terminate` on a HTTP server. This will then syncronsously call scheduler to notify envoy to stop forwarding reqs. This draining process is only triggered if it receives successisve calls from `kubelet` within a small time window, if at least 1 call was sent too slowly, the draining process will not commence and eventually `kubelet` will trigger `ctrl+c` without draining the agent and reqs will fail
3. The agent HTTP server which receives inference reqs, on spin up phase, was claiming to be ready to receive reqs when it may not have been, which would lead to failed reqs
4. The agent on spin up, connects to the scheduler and notifies it's loaded a model, before it has successfully been assigned an IP address which is accessible to other components in the namespace via DNS. This leads to envoy not being able to route reqs as it has a stale copy of the previous DNS record for the pod

## Summary of changes

In the order of the issues above:
1. Removed signalling failure of any services by closing the `done` channel. Instead create a new error channel  `errChan`. Instead of closing channel, any errors are sent on this channel and a subscriber listens to both the `ctrl+c` and `errChan` and will trigger shutdown if it receives on either channel.
2. The documented in code reason for the 3 pre-stop calls is to ensure it's a genuine shutdown sequence initiated by k8s. This seems over cautious IMHO so I've removed the triple call condition, and will trigger agent drain on the first call, successive calls will have no effect.
3. Removed marking the `reverseHTTPProxy` as ready before starting the HTTP server. Instead provide a `BaseContext` to the `http.Server` which calls this once the server has spun up. Within the clousre provided by `BaseContext` we mark the server as ready.
4. Once the pod has been marked as ready, we need to wait for k8s to assign the pod's IP to the `endpointslice` resource, this will make it available to DNS queries, so envoy can resolve the host. We make a GET request to k8s API to find out if we have an IP, if we don't, we watch on a channel, and wait for an update where the IP has been assigned. Once we receive a success from the channel, we unblock, and proceed to consume model command operations from the scheduler to load the model. Additionally I reduced the DNS refresh poller on envoy from the default `5 seconds` to `2 seconds` to mitigate stale records. 

After all of the above changes were made, initially no errors occured. After further testing, I eventually noticed a very few amount of reqs were still failing, once the pod had spun up. I imagine this is from incoming reqs falling into the time window where envoy has not updated the DNS record for the replica. To mitigate this I introduced an exponetial backoff retry in envoy, to only retry on connection issues or `5xx` response. This reduced the errors, but some were still failing. From checking the stats UI handler on envoy, envoy was allowing some of the backoff retries through, but blocking others as they exceeded the max default allowed retries of `3`. I introduced a cricuit breaker max retry threshold of `5` which then allowed all of the retries through. 

## Checklist

### DIFF
--- a/scheduler/cmd/agent/main.go
+++ b/scheduler/cmd/agent/main.go
@@ -303,7 +303,7 @@ func main() {
 	// Now we are ready start config listener
 	err = rcloneClient.StartConfigListener()
 	if err != nil {
-		errChan <- fmt.Errorf("failed to initialise rclone config listener: %w", err)
+		logger.WithError(err).Fatal("Failed to initialise rclone config listener")
 	}
 
 	// Start grpc connection to scheduler and handle incoming events


operator/controllers/mlops/seldonruntime_controller.go
```diff
@@ -108,6 +108,7 @@ func (r *SeldonRuntimeReconciler) handleFinalizer(ctx context.Context, logger lo
 // +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="discovery.k8s.io",resources=endpointslices,verbs=list;watch
```
domsolutions, line 111: controller needs this RBAC to allow server replicas to check their IP is visible

scheduler/pkg/agent/agent_svc_manager.go
```diff
@@ -548,6 +560,23 @@ func (am *AgentServiceManager) HandleSchedulerSubscription() error {
 	// This connection may break and will be retried, but we define the agent as "started"
 	// once we are able to successfully connect once.
 	am.isStartup.Store(false)
+
+	if am.agentConfig.runningInK8s {
```
domsolutions, line 562: We must only run this after `am.isStartup.Store(false)` as this allows the readiness check return 200 and for k8s to assign our IP to endpoints resource

scheduler/pkg/agent/readyservice/server.go
```diff
@@ -117,21 +116,25 @@ func (ready *ReadyService) GetHTTPHandler() *http.Handler {
...
-		// Check if the agent and its sub-services are ready
-		if am.Ready() {
-			w.WriteHeader(http.StatusOK)
-			_, err := w.Write([]byte("The agent and its dependent sub-services are ready"))
-			if err != nil {
-				ready.logger.WithError(err).Error("Failed to write response for readiness check")
-			}
-		} else {
-			http.Error(w, "The agent is not ready", http.StatusServiceUnavailable)
+	}
```
domsolutions, line 126: This is a no-op change, just a refactor for readability

scheduler/cmd/agent/main.go
```diff
@@ -290,33 +291,29 @@ func main() {
...
 		promMetrics,
+		clientset,
 	)
 
 	// Wait for required services to be ready
 	err = agentService.WaitReadySubServices(true)
 	if err != nil {
-		logger.WithError(err).Fatal("Failed to wait for all agent dependent services to be ready")
-		close(done)
+		errChan <- fmt.Errorf("failed to waiting for agent dependent sub-services to be ready: %w", err)
```
lc525: This send into the errChan channel is on the synchronous path, but `termSignalErrHandler` has not been called (yet). Therefore, the (fatal) error is just buffered, and execution continues. I think it would be better to shut down at this stage rather than continuing with `rcloneClient.StartConfigListener`.

I am aware that previously we also continued here rather than stopping. But if there's a fatal error, I think the right thing to do would be to exit early.

scheduler/pkg/agent/agent_svc_manager.go
```diff
@@ -548,6 +558,23 @@ func (am *AgentServiceManager) HandleSchedulerSubscription() error {
...
+		// In the case of not ready, caller will backoff retry indefinitely.
+		logger.Info("Checking pod has IP published in endpoints")
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		defer cancel()
+		// will block on k8s watch until IP assigned
+		err := am.extendedClient.HasPublishedIP(ctx, fmt.Sprintf("%s-%d", am.agentConfig.serverName, am.agentConfig.replicaIdx), "")
+		if err != nil {
+			return fmt.Errorf("failed waiting to check if pod's IP is published to endpoints: %v", err)
+		}
+		logger.Info("Found IP in endpoints")
```
lc525: nit: could we use `logger.Debug` here?

scheduler/pkg/agent/drainservice/server.go
```diff
@@ -26,8 +25,6 @@ import (
 
 const (
 	terminateEndpoint = "/terminate"
-	eventWindowMs     = 100
-	eventsTarget      = 3 // e.g. the number of containers in a pod
```
lc525, line 30: I've explained in the general comment why I think this structure was chosen, with waiting for calls to the pre-stop hook from all containers before continuing. The original PR introducing this behaviour is here: https://github.com/SeldonIO/seldon-core-v2/pull/644 (note it's a different repo). It's probably not needed most of the time, but it essentially tries to identify a pod termination vs a call to the pre-stop hook of a particular container (say, just rclone using lots of resources and beeing preempted).

scheduler/pkg/agent/drainservice/server.go
```diff
@@ -145,19 +138,13 @@ func (drainer *DrainerService) handleTerminate(w http.ResponseWriter, _ *http.Re
...
+	defer drainer.muTriggered.Unlock()
+
+	if !drainer.triggered {
+		drainer.triggered = true
+		now := time.Now()
+		drainer.triggeredWg.Done()
 		drainer.drainingFinishedWg.Wait()
-	} else {
-		drainer.events = 0
+		drainer.logger.Infof("Drainer service completed handling %s endpoint, took %s", terminateEndpoint, time.Since(now))
```
lc525: As a note, here we'll see this in the logs only if the termination grace period was not exceeded. But showing this info will certainly allow us to understand how close we're getting to the end of the grace period when under load

scheduler/pkg/agent/k8s/extended_client.go
lc525, line 1: As a note, we should understand if setting `spec.publishNotReadyAddresses=True` on the Service would simplify this (i.e not require us to watch the endpointslice for readiness) for all cases. But right now this change brings us to a significantly better place wrt 5xx errors.

domsolutions: Ah, didn't know that was configurable.. sounds like it would simplify it, but would need to test. I'll create a ticket for it.

scheduler/cmd/agent/main.go
domsolutions: Yep makes sense, FYI prior to my change it wouldn't have continued as the `Fatal` would have triggered an exit with error code 1, but the code mistakingly was then attempting to close the done channel after which it would never reach. I've just changed it to log as a fatal error now.

lc525: We would want to do the same thing for `rcloneClient.StartConfigListener()` below this, right (i.e error out immediately)?

PR #6664, opened 2025-08-06
fix: enable link checker on docs-gb for PRs to master and v2

monica-seldon:
# Why
## Motivation

# What
## Checklist

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -27,4 +27,3 @@ _template
 \.txt$
 \.pdf$
 \.csv$
-^file://


.github/workflows/daily-link-checker.yml
```diff
@@ -1,41 +1,20 @@
-name: Check all Markdown links daily
+name: Nightly Link Checker
```
tyndria: **suggestion** Minor and a bit funny to have a file named `daily-link-checker.yml`, but then the workflow names `Nightly Link Checker` :)

I propose to 
- **name the workflow `GB Docs Link Checker`** as it's consistent with other workflow names
Image

- **name the file `gb-docs.yml`**, as we don't want to rename the file if we decided to change the schedule of the job

.github/workflows/daily-link-checker.yml
tyndria, line 1: **thought**: there is an old link checker for _old Core 1 docs_ [here](https://github.com/SeldonIO/seldon-core/blob/master/.github/workflows/docs.yml) with name `V1 Docs Test`. It seems we would have to remove it after Core 1 docs migration?

I'll cc @Rajakavitha1 to discuss that after she came back

.github/workflows/daily-link-checker.yml
```diff
@@ -1,41 +1,20 @@
...
-  workflow_dispatch:
-
-permissions:
-  contents: read
-  issues: write
 
 jobs:
-  check-docs-gb:
+  lychee-check:
     runs-on: ubuntu-latest
```
tyndria: **todo**: I believe we shouldn't run this action for the forks, so we need to add this line:
```
if: github.repository == 'seldonio/seldon-core'
steps:
  ...
```

.github/workflows/daily-link-checker.yml
tyndria: **question/thought**: question for @Rajakavitha1  - Is it correct to use this file for both Core 1 and Core 2 docs? What if there is a conflict in the `.lycheeignore` file? For example, we should ignore a file matching the pattern in Core 1 docs, but we shouldn't ignore it in Core 2 docs?

Rajakavitha1: Hi @tyndria  the file ignore is consistent across both Core 1 and Core 2 docs. I guess it is okay to use it.

.lycheeignore
```diff
@@ -0,0 +1,40 @@
...
+^production-environment/
+^product-tour/
+^demos/
+^architecture/
+^operations/
+^upgrading/
+^README.md
+^multi-tentant-environment.md
+^openshift-environment.md
+^help-and-support/
```
Rajakavitha1: Why are we ignoring these?

.lycheeignore
```diff
@@ -0,0 +1,40 @@
...
+^http://default-broker.seldon-logs
+^https://www.tensorflow.org/tfx/serving/api_rest#request_format_2
+^https://otexts.org/fpp2/holt-winters.html
+^https://www.pcmag.com/encyclopedia
+^https://www.microsoft.com/en-us/download/details.aspx?
+^https://opensea.io/
+^https://www.techopedia.com/definition/31996/hash-partitioning
+^https://docs.github.com/
+_template
+^https://127\.0\.0\.1(:\d+)?/
```
Rajakavitha1, line 13: may be you can add the http:// as well. It was reported https://github.com/SeldonIO/seldon-core/actions/runs/16800884316?pr=6664

Image

Rajakavitha1 (commented):
Thank you @monica-seldon . Just a couple of changes

.lycheeignore
monica-seldon: These entries were already present in the previous ignore file. I haven’t made any changes to them @Rajakavitha1

Rajakavitha1: Can we please not ignore these files.

.lycheeignore
tyndria: @Rajakavitha1 did you mean here `http://127.0.0.1` or just `http://`?

Monica added `^http://`, but I don't think it's valid to ignore **all links starting from http://** as there could be links we would like to be really working

cc @monica-seldon

Rajakavitha1: > @Rajakavitha1 did you mean here `http://127.0.0.1` or just `http://`?
> 
> Monica added `^http://`, but I don't think it's valid to ignore **all links starting from http://** as there could be links we would like to be really working
> 
> cc @monica-seldon

Hi @monica-seldon  and @tyndria  could you please ensure that http://127.0.0.1 is not reported as broken link.

Thank you.

monica-seldon: Hi @Rajakavitha1 @tyndria — I checked the latest link check report and http://127.0.0.1 is not being reported as broken.

PR #6675, opened 2025-08-06
feat: pipeline loadbalancer

RobertSamoilescu:
# Why
## Motivation

This PR enhances the horizontal scalability of the pipeline gateway by evolving the scaling model from `#partitions` to `#partitions × min(max_num_consumers, #pipelines)`. 

# What
## Summary of changes

* **Removal of the pipeline-gateway service**: The pipeline-gateway service was removed and the corresponding permanent cluster.

* **Per-pipeline clustering**: A dedicated cluster is now created for each pipeline. Each cluster's endpoints are defined by the IPs of the pipeline-gateway pods assigned to that pipeline.

* **xDS event hub update**: Events to xDS are now emitted after load balancing decisions are made. Previously, events were sent immediately after a pipeline was added, but this approach is no longer feasible because the responsible pipeline-gateway replicas aren't known at that point.

* **Handling request rebalancing in infer**: During inference, if a partition rebalance occurs mid-processing, the request may be canceled—this happens when the consumer is no longer responsible for the partition where the request was published. An error is returned in such cases. In the future, we may introduce a retry mechanism to handle such scenarios more gracefully.

* **Pipeline gateway stream identification**: pipeline-gateway instances now identify themselves via a boolean flag when connecting to the scheduler. This allows the scheduler to differentiate between gateway and non-gateway streams and apply load balancing logic appropriately.

## Checklist

### DIFF
--- a/operator/controllers/reconcilers/seldon/runtime_reconciler.go
+++ b/operator/controllers/reconcilers/seldon/runtime_reconciler.go
@@ -61,13 +61,6 @@ func getEnvVarValue(podSpec *v1.PodSpec, name string, defaultValue string) strin
 	return defaultValue
 }
 
-func min(a, b int32) int32 {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 func replicaCalc(resourceCount, maxConsumers, partitions int32) int32 {
 	if resourceCount == 0 {
 		return 1
--- a/scheduler/pkg/util/loadbalancer.go
+++ b/scheduler/pkg/util/loadbalancer.go
@@ -34,13 +34,6 @@ func NewRingLoadBalancer(numPartitions int) *RingLoadBalancer {
 	}
 }
 
-func min(a, b int) int {
-	if a < b {
-		return a
-	}
-	return b
-}
-
 func (lb *RingLoadBalancer) AddServer(serverName string) {
 	lb.ring = lb.ring.AddNode(serverName)
 	lb.nodes[serverName] = true


apis/mlops/scheduler/scheduler.proto
```diff
@@ -401,6 +401,8 @@ message PipelineStatusRequest {
 
 message PipelineSubscriptionRequest {
   string subscriberName = 1; //Name of the subscription caller
+  string subscriberIp = 2; // IP of the subscription caller
+  bool isPipelineGateway = 3; // If true, the subscription is for a pipeline gateway
```
lc525, line 405: To be added as an internal issue (jira), and dealt-with in a separate PR, as previously discussed:

Rather than having lots of special-casing in the subscription request message (with us needing to modify the protobuf each time special handling is needed), I suggest we make the identification ID-based. For this purpose, we should introduce a `subscriberID` field. This field can be set to a v5 UUID, and the scheduler should be aware of a set of specific (constant) v5 UUIDs representing subscribers for which special handling is needed.

apis/mlops/scheduler/scheduler.proto
```diff
@@ -401,6 +401,8 @@ message PipelineStatusRequest {
 
 message PipelineSubscriptionRequest {
   string subscriberName = 1; //Name of the subscription caller
+  string subscriberIp = 2; // IP of the subscription caller
```
lc525, line 404: At the time the scheduler receives the subscription request, this info is probably accessible directly via `https://pkg.go.dev/google.golang.org/grpc/peer#FromContext` applied to the grpc context.

lc525: Were you considering complex scenarios such as multiple pipeline subscribers (potentially pipeline-gateway pods) connecting via a single proxy? (in that case, the `FromContext` would only return the IP of the proxy).

operator/controllers/reconcilers/seldon/runtime_reconciler.go
```diff
@@ -40,80 +49,124 @@ func ParseInt32(s string, defaultVal int32) (int32, error) {
...
+			if env.Name == name {
+				return env.Value
+			}
+		}
+	}
+
+	return defaultValue
+}
+
+func min(a, b int32) int32 {
```
lc525: please use the [builtin min function](https://go.dev/ref/spec#Min_and_max), available since go 1.21, instead

operator/controllers/reconcilers/seldon/runtime_reconciler.go
```diff
@@ -40,80 +49,124 @@ func ParseInt32(s string, defaultVal int32) (int32, error) {
...
+	}
+	return partitions * min(resourceCount, maxConsumers)
+}
+
+func validateScaleSpec(
 	component *mlopsv1alpha1.ComponentDefn,
 	runtime *mlopsv1alpha1.SeldonRuntime,
 	commonConfig common.ReconcilerConfig,
 	namespace *string,
+	envVar string,
```
lc525: I suggest an argument rename to reflect the fact that `envVar` should contain the max number of consumers.

operator/controllers/reconcilers/seldon/runtime_reconciler.go
```diff
@@ -40,80 +49,124 @@ func ParseInt32(s string, defaultVal int32) (int32, error) {
...
+	maxReplicas := replicaCalc(resourceCount, maxConsumers, numPartitions)
 	if component.Replicas != nil && *component.Replicas > maxReplicas {
 		component.Replicas = &maxReplicas
 		recorder.Eventf(
 			runtime,
 			v1.EventTypeWarning,
-			"DataflowEngineReplicasAdjusted",
-			fmt.Sprintf("Dataflow engine replicas adjusted to %d based on KafkaConfig and Pipeline count", maxReplicas),
+			eventReason,
+			fmt.Sprintf("%s replicas adjusted to %d based on KafkaConfig and resource count", component.Name, maxReplicas),
```
lc525: The message should reflect that the number of requested replicas was above the maximum possible for the given configuration, and only then that it was adjusted.

operator/controllers/reconcilers/seldon/runtime_reconciler.go
```diff
@@ -40,80 +49,124 @@ func ParseInt32(s string, defaultVal int32) (int32, error) {
...
+func validateScaleSpec(
 	component *mlopsv1alpha1.ComponentDefn,
 	runtime *mlopsv1alpha1.SeldonRuntime,
 	commonConfig common.ReconcilerConfig,
 	namespace *string,
+	envVar string,
+	defaultMaxConsumers int32,
+	eventReason string,
+	resourceListObj client.ObjectList,
+	countResources func(client.ObjectList) int,
```
lc525, line 80: From what I see in the callers of this function, countResources always just calls `len` on `resourceListObj.Items` (the equivalent local variable names in this function). Have you added this to allow for more flexibility in the future (i.e custom filtering of what is considered and what is not?). 

I think that's a great idea (for example we might need to exclude pipelines that are "pinned" to particular `dataflow-engine` replicas).

scheduler/pkg/server/pipeline_status_test.go
```diff
@@ -210,3 +211,276 @@ func TestPipelineStatusEvents(t *testing.T) {
...
+
+			// trigger rebalance and allow events to propagate
+			s.pipelineGwRebalance()
+			time.Sleep(500 * time.Millisecond)
+
+			var msr *pb.PipelineStatusResponse
+			select {
+			case next := <-stream.msgs:
+				msr = next
+			default:
```
lc525: For new tests that wait for event propagation, I recommend the following pattern:

* sleep for a relatively short duration to allow for event propagation (the one above at 500ms is just fine)
* in the select, rather than having a `default` case (which is selected immediately if no event is present in another case), have a second case with a longer timeout (i.e `case <- time.After(2*second)`. This can prevent flaky tests, especially when running under constrained resources (like CI).

scheduler/pkg/server/pipeline_status_test.go
```diff
@@ -210,3 +211,276 @@ func TestPipelineStatusEvents(t *testing.T) {
...
+			time.Sleep(500 * time.Millisecond)
+
+			// Read all messages from the streams before rebalance
+			for i, stream := range streams {
+			drainLoop:
+				for {
+					select {
+					case msg := <-stream.msgs:
+						log.Infof("Drained message from stream %d: %v", i, msg)
+					default:
```
lc525: Similar to comment above regarding allowing for a longer event propagation time but without having all of it as a prepended sleep.

lc525 (approved):
lgtm, with some comments/suggestions related to implementation details. The overall approach is well thought-out and should be a solid base to build upon!

apis/mlops/scheduler/scheduler.proto
RobertSamoilescu: As discussed offline, we'll leave it like this for now.

PR #6689, opened 2025-08-18
feat(kafka): Schema registry

MiguelAAe:
# Why
Adding optional Schema Registry support enables seamless integration with the Confluent ecosystem, including connectors and third-party producers/consumers. This enhancement provides schema evolution capabilities, data contract enforcement, and broader interoperability while maintaining full backward compatibility with our existing Protobuf message format. Schema Registry usage can be toggled via configuration, allowing teams to adopt it incrementally based on their integration requirements.

## Issues
- Current Kafka transport cannot integrate with Confluent connectors
- Third-party systems struggle to consume our Protobuf messages without schema metadata
- No mechanism for schema evolution or compatibility checks

## Motivation
Enable broader ecosystem adoption and simplify integration patterns for external teams while preserving existing functionality.

# What
## Summary of changes
- Added `ProtobufByteArraySerializer` and `ProtobufByteArrayDeserializer` classes to handle Schema Registry wire format
- Implemented configurable serde factory that switches between raw Protobuf and Schema Registry modes
- Added Schema Registry client integration for dynamic schema ID resolution
- Maintained backward compatibility with existing raw byte array serialization

## Tested
- [x] Confluent cloud Connection to a Schema Registry
- [x] Local Schema Registry
- [x] Different types of pipelines
- [x] Complex Pipelines with joints
- [x] Confluent Cloud Kafka cluster with Schema Registry
- [x] Failure modes with connection lost from Data flow, model gateway and pipeline gateway
- [x] Kafka Connect Protobuf Converter 

## Todo
- [ ] Update protobuf and kafka streams dependency to 8.0.0 to use auto gen protobuf code from the version in main @lc525 

## Checklist
- [x] Added/updated documentation

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/SchemaRegistry.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/SchemaRegistry.kt
@@ -71,13 +71,9 @@ class SchemaRegistrySerializerFactory(private val config: SchemaRegistryConfig) 
         }
     }
 
-    val requestSerializer: KafkaProtobufSerializer<ModelInferRequest> by lazy {
-        createAndConfigureSerializer()
-    }
+    val requestSerializer: KafkaProtobufSerializer<ModelInferRequest> = createAndConfigureSerializer()
 
-    val responseSerializer: KafkaProtobufSerializer<ModelInferResponse> by lazy {
-        createAndConfigureSerializer()
-    }
+    val responseSerializer: KafkaProtobufSerializer<ModelInferResponse> = createAndConfigureSerializer()
 
     private fun <T : Message> createAndConfigureSerializer(): KafkaProtobufSerializer<T> {
         return try {


scheduler/data-flow/build.gradle.kts
```diff
@@ -28,8 +28,10 @@ dependencies {
     implementation("io.klogging:slf4j-klogging:0.11.1")
 
     // Kafka
-    implementation("org.apache.kafka:kafka-streams:7.7.0-ccs")
-    testImplementation("org.apache.kafka:kafka-streams-test-utils:7.7.0-ccs")
+    implementation("org.apache.kafka:kafka-streams:7.9.0-ccs")
```
MiguelAAe: this value needs to be updated to 8.0.0 for compatibility with our current protobuf version in order to use confluent schema registry lib
https://mvnrepository.com/artifact/org.apache.kafka/kafka-streams?repo=confluent-packages

scheduler/data-flow/build.gradle.kts
```diff
@@ -38,14 +40,14 @@ dependencies {
     runtimeOnly("io.grpc:grpc-netty-shaded:1.73.0")
     implementation("com.google.protobuf:protobuf-java") {
         version {
-            strictly("[4.27.2,)")
-            prefer("4.27.2")
+            strictly("[3.25.5,)")
```
MiguelAAe: downgrade of protobuf for the time being, needs to be rolled back up to 4.27.2 once we have upgraded kafka streams to a higher version

k8s/helm-charts/seldon-core-v2-setup/values.yaml.template
```diff
@@ -36,6 +36,11 @@ security:
         caPath: /tmp/certs/kafka/client/ca.crt
         brokerCaPath: /tmp/certs/kafka/broker/ca.crt
         endpointIdentificationAlgorithm:
+  schemaRegistry:
+    client:
+      URL:
+      username:
+      password:
```
MiguelAAe: @lc525 do we need any kind of special loading of secrets here?

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -526,3 +545,36 @@ func (km *KafkaManager) consume(pipeline *Pipeline) error {
...
+	v2Res := &inference_schema.ModelInferRequest{}
+	err := proto.Unmarshal(payload, v2Res)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal request to dataplane model: %w", err)
+	}
+
+	schemaConfig := protobuf.NewSerializerConfig()
+	schemaConfig.NormalizeSchemas = true
+
+	ser, err := protobuf.NewSerializer(km.schemaRegistryClient, serde.ValueSerde, schemaConfig)
```
MiguelAAe, line 571: Future version could load a protobuf serialiser as dependency instead of a schema registry client

apis/Makefile
```diff
@@ -24,7 +24,9 @@ build: build-go build-kotlin build-v2-python update-copyright
 # the latest version of protoc + language-specific plugins.
 
 # https://github.com/protocolbuffers/protobuf/releases
-PROTOC_VERSION ?= 27.2
+PROTOC_VERSION ?= 25.5
+
+PROTOC_VERSION_SCHEMA_REGISTRY ?= 25.0
```
lc525: Once #6795 is merged, this can be rolled back

k8s/helm-charts/seldon-core-v2-setup/templates/_components-deployments.tpl
```diff
@@ -844,6 +849,10 @@ spec:
       - configMap:
           name: seldon-tracing
         name: tracing-config-volume
+      - name: confluent-schema-volume
```
lc525: I suggest we make the naming across various places in the manifest vendor-neutral. Here, it could be kafka-schema-volume

scheduler/cmd/modelgateway/main.go
```diff
@@ -57,6 +59,8 @@ var (
 	logLevel               string
 	tracingConfigPath      string
 	healthProbeServicePort int
+	// TODO: add file watcher cfg using koanf and in the future read all file config in one file
+	k = koanf.New(".")
```
lc525, line 63: Note: we're using `spf13/viper` in other packages, so it might be worth harmonizing at some point in the future. Leave as is for now.

scheduler/data-flow/build.gradle.kts
```diff
@@ -61,11 +63,12 @@ dependencies {
     implementation("io.ktor:ktor-server-content-negotiation:3.0.1")
     implementation("io.ktor:ktor-serialization-kotlinx-json:3.0.1")
     implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.7.3")
+    implementation("com.charleskorn.kaml:kaml:0.61.0")
 
     testImplementation(kotlin("test"))
     testImplementation("org.junit.jupiter:junit-jupiter-params:5.10.3")
     testImplementation("io.strikt:strikt-core:0.34.1")
-    testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("org.assertj:assertj-core:3.25.1")
```
lc525, line 71: Where is assertj used? I haven't seen it imported in the added tests, but it's likely I've missed it.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/SchemaRegistry.kt
```diff
@@ -0,0 +1,252 @@
...
+
+    private fun <T : Message> createAndConfigureSerializer(): KafkaProtobufSerializer<T> {
+        return try {
+            val serializer = KafkaProtobufSerializer<T>()
+            serializer.configure(config.toSerializerProperties(), false)
+
+            try {
+                serializer.schemaRegistryClient.allSubjects
+            } catch (e: Exception) {
+                logger.info("failed to ping schema registry client")
```
lc525: Should this be logged as an error?

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/SchemaRegistry.kt
```diff
@@ -0,0 +1,252 @@
...
+    org.apache.kafka.common.serialization.Serializer<ByteArray> {
+    private val logger = Logger(ProtobufWireFormatSerializer::class)
+
+    override fun serialize(
+        topic: String?,
+        data: ByteArray?,
+    ): ByteArray? {
+        if (data == null) return null
+
+        val topicSuffix = topic?.splitToSequence(".")?.last()
```
lc525, line 151: This is implemented quite nicely!

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Types.kt
```diff
@@ -24,7 +20,8 @@ typealias RequestId = String
 typealias TRecord = ByteArray
 typealias TopicsAndTensors = Pair<Set<TopicName>, Set<TensorName>>
 
-class SamePartitionForwarder : StreamPartitioner<String, TRecord> {
+class SamePartitionForwarder<TRecord> : StreamPartitioner<String, TRecord> {
```
lc525: Do we need to make this generic? Here, aren't we just defining a type parameter named `TRecord` that shadows the type alias declared above? It would be sufficient to call it `<T>` to avoid confusions. Is the added generic because in serialization this is no longer always used as a pure ByteArray?

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/kafka/SchemaRegistryTest.kt
```diff
@@ -0,0 +1,260 @@
...
+
+        @Test
+        fun `topic with input in the name and extra dots`() {
+            val topic = "pipeline.test.input..outputs"
+
+            val lastSuffix = topic.splitToSequence(".").last()
+
+            expectThat(lastSuffix).isEqualTo("outputs")
+        }
+    }
```
lc525: This test is a bit superfluous, because it just tests that the stdlib's `last()` works correctly. I suggest we remove it.

scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -23,7 +23,10 @@ import (
 	"time"
 
 	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
-	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
+	"github.com/confluentinc/confluent-kafka-go/v2/schemaregistry"
+	"github.com/confluentinc/confluent-kafka-go/v2/schemaregistry/serde"
+	"github.com/confluentinc/confluent-kafka-go/v2/schemaregistry/serde/protobuf"
+	grpcretry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
```
lc525: This type of renaming is good in terms of good practices, but for large PRs I would rather it gets split into its own PR so as to not add noise to the diff when reviewing the actual changes. It's fine as it is for this PR, just something I'm proposing for the future.

scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -402,6 +420,43 @@ func (iw *InferWorker) grpcRequest(ctx context.Context, job *InferWork, req *v2.
 	return nil
 }
 
+func (iw *InferWorker) serializeModelInferRespWithSchemaRegistry(topic string, payload []byte) ([]byte, error) {
+	iw.logger.Debugf("first 10 bytes before schema serialisation")
+	if len(payload) > 10 {
+		for _, b := range payload[:10] {
+			iw.logger.Debugf("%02x", b)
+		}
+	}
```
lc525: Suggest it would be more suitable to log this at the "trace" level.

scheduler/pkg/kafka/gateway/worker.go
```diff
@@ -402,6 +420,43 @@ func (iw *InferWorker) grpcRequest(ctx context.Context, job *InferWork, req *v2.
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to serialise response to dataplane model with schema id: %w", err)
+	}
+
+	iw.logger.Debugf("first 10 bytes after schema serialisation")
+	if len(payload) > 10 {
+		for _, b := range payload[:10] {
+			iw.logger.Debugf("%02x", b)
+		}
+	}
```
lc525, line 466: Same comment as above

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -526,3 +545,40 @@ func (km *KafkaManager) consume(pipeline *Pipeline) error {
...
 	return nil
 }
+
+func (km *KafkaManager) serializeModelInferReqWithSchemaRegistry(topic string, payload []byte) ([]byte, error) {
+	km.logger.Debugf("first 10 bytes before schema serialisation")
+	if len(payload) > 10 {
+		for _, b := range payload[:10] {
+			km.logger.Debugf("%02x", b)
+		}
+	}
```
lc525: Suggest trace level for logging this and the last 10 bytes

scheduler/pkg/kafka/schema/schema.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// TrimSchemaID trims the magic byte, schema id and message index of a kafka message that was sent with a schema ID
+func TrimSchemaID(payload []byte) []byte {
+	// If it's Schema Registry format (magic byte 0x0)
+	if len(payload) < 6 {
+		return payload
+	}
+	if payload[0] == 0x0 {
+		// Skip magic byte (1) + schema ID (4) + message index (0) = 6 bytes
```
lc525: Assuming that this should read `+ message index (1)`

scheduler/cmd/modelgateway/main.go
MiguelAAe: Ok no worries, Koanf was chosen due to its lightweight and modular architecture and reduced dependencies but I didn't see it was used in the operator.

scheduler/data-flow/build.gradle.kts
MiguelAAe: it is used here: https://github.com/SeldonIO/seldon-core/pull/6689/files#diff-d25de7909966a2b5d6d865fe31a8fbe7a9061dfa37f4473dd2053b9c7784834bR49

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/SchemaRegistry.kt
MiguelAAe: yes it should

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Types.kt
MiguelAAe: I think this could be reverted back, I was having some linting issues and I believe the Deprecated tag is enough

scheduler/cmd/modelgateway/main.go
lc525: Yes, we can ultimately move to koanf (i.e remove the dependency on `spf13/viper`) but it wouldn't be high priority.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Types.kt
lc525: Just to be clear, we can leave as is but was suggesting we transform to `SamePartitionForwarder<T> : StreamPartitioner<String, T>`. Why would `Deprecated` be needed? In the previous version, it would just be a class inheriting from a generic where all type parameters have been specified

scheduler/pkg/kafka/pipeline/kafkamanager.go
MiguelAAe: addressed in b1b98e5fe1ff26451bfbe7633dcdb9b932f7f1da

scheduler/pkg/kafka/schema/schema.go
MiguelAAe: addressed in b1b98e5fe1ff26451bfbe7633dcdb9b932f7f1da

k8s/helm-charts/seldon-core-v2-setup/templates/_components-deployments.tpl
MiguelAAe: is the name for the secret fine? https://github.com/SeldonIO/seldon-core/pull/6689/files#diff-d4285188a6cb5438bdcfded165f13b11cc1a40623fe1ebc2f56bd10136a196b2R855

MiguelAAe: left the secret `confluent-schema` the same to be more in line with the secret `confluent-kafka-sasl `

PR #6690, opened 2025-08-21
fix(pipeline-gw): failed incoming reqs when partitions not available 

domsolutions:
# Why

Some inference reqs would fail when no partitions were available to the kafka consumer due to partition revoking, this can happen for many reasons i.e. kafka broker overloaded, connection issues between consumer and broker, consumer too slow etc. 

## Summary of changes

- We now wait for max of `10 seconds` for at least one partition to be available, and then proceed with inference req. If we timeout, we respond with `500`
- Envoy will now backoff retry failed reqs for pipelines which either failed with a `5xx` error or there were connection issues

## Checklist

### DIFF
--- a/scheduler/pkg/kafka/pipeline/broadcaster.go
+++ b/scheduler/pkg/kafka/pipeline/broadcaster.go
@@ -1,40 +1,41 @@
+/*
+Copyright (c) 2024 Seldon Technologies Ltd.
+
+Use of this software is governed by
+(1) the license included in the LICENSE file or
+(2) if the license included in the LICENSE file is the Business Source License 1.1,
+the Change License after the Change Date as each is defined in accordance with the LICENSE file.
+*/
+
 package pipeline
 
 import "sync"
 
 type Broadcaster struct {
-	mu        sync.RWMutex
-	listeners []chan struct{}
+	mu    sync.Mutex
+	ready chan struct{}
 }
 
 func NewBroadcaster() *Broadcaster {
-	return &Broadcaster{
-		listeners: make([]chan struct{}, 0),
-	}
+	return &Broadcaster{}
 }
 
 func (b *Broadcaster) Subscribe() <-chan struct{} {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
-	ch := make(chan struct{})
-	b.listeners = append(b.listeners, ch)
-	return ch
-}
-
-func (b *Broadcaster) HasListeners() bool {
-	b.mu.RLock()
-	defer b.mu.RUnlock()
-
-	return len(b.listeners) > 0
+	if b.ready == nil {
+		b.ready = make(chan struct{})
+	}
+	return b.ready
 }
 
 func (b *Broadcaster) Broadcast() {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
-	for _, ch := range b.listeners {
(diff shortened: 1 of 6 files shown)


scheduler/pkg/kafka/pipeline/httpserver.go
```diff
@@ -182,7 +182,6 @@ func (g *GatewayHttpServer) infer(w http.ResponseWriter, req *http.Request, reso
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
 		} else {
-			w.WriteHeader(http.StatusOK)
```
domsolutions, line 185: This was a superfluous call as header is already written within `w.Write(resJson)` and was causing the logs to fill up with warnings 
```
http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/github.com/gorilla/mux/otelmux/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:83)      
```

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -327,7 +350,7 @@ func (km *KafkaManager) Infer(
 	}
 	go func() {
 		evt := <-deliveryChan
-		logger.Infof("Received delivery event %s", evt.String())
+		logger.Debugf("Received delivery event %s", evt.String())
```
domsolutions, line 353: I put a lot of logs to `debug` instead of `info` as when doing load tests it was very hard to try and find the errors amongst the noise, should also give us a slight performce improvement as logrus allocates a lot (probably should look at replacing our logger with something more performant like zap at some point) and imagine most customers have it deployed at `info`

scheduler/pkg/kafka/pipeline/broadcaster.go
```diff
@@ -0,0 +1,41 @@
...
+(2) if the license included in the LICENSE file is the Business Source License 1.1,
+the Change License after the Change Date as each is defined in accordance with the LICENSE file.
+*/
+
+package pipeline
+
+import "sync"
+
+type Broadcaster struct {
+	mu    sync.Mutex
```
RobertSamoilescu: I would suggest to replace `sync.Mutex` with `sync.RWMutex`

scheduler/pkg/kafka/pipeline/broadcaster.go
```diff
@@ -0,0 +1,41 @@
...
+	ready chan struct{}
+}
+
+func NewBroadcaster() *Broadcaster {
+	return &Broadcaster{}
+}
+
+func (b *Broadcaster) Subscribe() <-chan struct{} {
```
RobertSamoilescu, line 23: First try to read lock using `RLock` & `RUnlock`. Only if the channel is not ready, then create the channel. This would allow multiple subscribers to subscribe simultaneously if the channel is already initialized.

RobertSamoilescu (approved):

scheduler/pkg/kafka/pipeline/broadcaster.go
domsolutions: Fair point, I wasn't too concerned about the lock contention as it's not a hot path, at least it shouldn't be, but you're right, shouldn't introduce contention where we can avoid it.

PR #6695, opened 2025-08-22
fix(pipeline-gw): re-publish in-flight reqs due to partition revoke

domsolutions:
# Motivation

When partitions get revoked, we can no loger wait for reqs already published to kafka as we don't have access to the partition and we return an error to the client. This PR will now re-attempt to publish the req once a partition becomes available. 

## Summary of changes

- a new sentinel error `errPartitionRevoked` which infer will treat as a special case and re-attempt to publish the req a max of once, it will wait for a max of 10 seconds for a partition and then error out, at which point envoy backoff retry will manage the retry
- we commit and outstanding offsets when the partitions are revoked

## Checklist


RobertSamoilescu (approved):
LGTM!

PR #6728, opened 2025-09-01
feat(pipeline-gw): health probes

domsolutions:
# Motivation

We require health probes on `pipeline-gw` for k8s to know if the service is not healthy and if it should be restarted for example. 

# Summary of changes

1. Startup probe added which will pass once the connection to the scheduler is established and HTTP/gRPC servers are up
2. Readiness probe added which will pass only when HTTP/gRPC servers are up and kafka producer/consumer is healthy
3. Liveness probe added which will pass only when HTTP/gRPC servers are up and kafka producer/consumer is healthy

Importantly the readiness/liveness probes will not fail if the connection to the scheduler is lost, as they can still serve inference reqs, just no new pipelines can be loaded/unload (eventually the container will os.Exit if it can't establish a connection after backoff retry)

Startup probe additionally checks for a heathy connection to the scheduler, as this is required on spin up to receive the pipelines from the scheduler to serve. 

A failure on the readiness probe will cause the pod's IP not to be available via k8s DNS (i.e. `/endpointslice`). This won't actually affect anything as `envoy` does not use DNS to route traffic to `pipeline-gw` as `envoy` is configured to use the pod's IP. This probe has just been added for completeness. 

Additionally fixed panic on shutdown when closing already closed `done` channel.

## Checklist


scheduler/pkg/health-probe/health.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type service struct {
+	probes ProbeType
+	cb     ProbeCallback
+}
+
+func NewManager() Manager {
+	return &manager{
+		svcs: make([]service, 0),
```
lc525: a nit here, but it would be nicer to have the svcs be a map keyed by ProbeType, with the value being an array of ProbeCallbacks. This would mean that when you need to fire a probe (say, ProbeReadiness), you just iterate through all the callbacks of that probe. It will make no difference to efficiency as I don't imagine us having thousands of callbacks :-)

scheduler/cmd/pipelinegateway/main.go
```diff
@@ -222,5 +235,47 @@ func main() {
...
+		return nil
+	}, health.ProbeStartUp)
+
+	// note this will not attempt connection handshake until req is sent
+	conn, err := grpc.NewClient(fmt.Sprintf(":%d", grpcPort),
+		grpc.WithConnectParams(grpc.ConnectParams{
+			Backoff: backoff.DefaultConfig,
+		}),
+		grpc.WithKeepaliveParams(util.GetClientKeepAliveParameters()),
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
```
lc525: Will this fail if the end-user has enabled TLS on the GRPC service? (see [here](https://github.com/SeldonIO/seldon-core/blob/1ea918a23aa52431c99c7cf4c2faee52b03db2db/scheduler/pkg/kafka/pipeline/grpcserver.go#L78))

lc525 (commented):
lgtm; One significant question on whether the grpc health check callback will still work as expected for customers that have enabled TLS.

scheduler/cmd/pipelinegateway/main.go
domsolutions: Good spot, it indeed failed, also failed on the HTTP probe when TLS was enabled. Instead I created a non-TLS HTTP server which will call out to both HTTP/gRPC with the correct TLS config if it's enabled via the env var. Tested with TLS on/off and looks good.

scheduler/cmd/pipelinegateway/main.go
```diff
@@ -182,45 +184,163 @@ func main() {
...
 	if err := httpServer.Stop(); err != nil {
 		logger.WithError(err).Error("Failed to stop http server")
 	}
 	logger.Infof("Shutting down scheduler client")
 	grpcServer.Stop()
-	pipelineSchedulerClient.Stop()
+	schedulerClient.Stop()
+}
+
+func waitForTermSignalOrErr(logger *log.Logger, errChan <-chan error) {
```
domsolutions, line 244: Additionally fixed panic on shutdown when closing already closed `done` channel.

scheduler/cmd/pipelinegateway/main.go
```diff
@@ -86,16 +97,7 @@ func init() {
...
-
-func makeSignalHandler(logger *log.Logger, done chan<- bool) {
-	exit := make(chan os.Signal, 1)
-	signal.Notify(exit, os.Interrupt, syscall.SIGTERM)
-
-	<-exit
-
-	logger.Info("shutting down due to SIGTERM or SIGINT")
-	close(done)
+	flag.IntVar(&healthProbeServicePort, flagHealthPort, defaultHealthProbePort, "K8s health probe port")
```
lc525:
```suggestion
	flag.IntVar(&healthProbeServicePort, flagHealthPort, defaultHealthProbePort, "Health probe port")
```
Suggestion because Core 2 will happily run in something like docker compose if needed, it's just that we don't really support those deployment types for production atm.

PR #6745, opened 2025-09-02
feat(model-gw): health probes

domsolutions:
# Motivation

We require health probes on `model-gw` for k8s to know if the service is not healthy and if it should be restarted for example.

# Summary of changes

1. startup probe only checks if connection to scheduler is successful
2. liveness/readiness probes will check if the kafka producer(s) are all not closed and all consumer(s) are still active

A failure on the readiness probe will cause the pod's IP not to be available via k8s DNS (i.e. `/endpointslice`). This won't actually affect anything as no other pods use the DNS  to communicate with the pod. This probe has just been added for completeness.

As part of this PR, also fixed the panic on shutdown which occurs when closing the already closed `done` channel .Graceful shutdown should now work.

## Checklist


k8s/helm-charts/seldon-core-v2-setup/templates/_components-deployments.tpl
```diff
@@ -917,6 +918,8 @@ spec:
           value: "9004"
         - name: SELDON_SCHEDULER_TLS_PORT
           value: "9044"
+        - name: HEALTH_PROBE_PORT
+          value: "8000"
```
lc525: Noting that we changed the port from 9999 on pipeline-gw to 8000 for this service. Doesn't
really matter in the end as long as the ports are available in the container though.

scheduler/pkg/kafka/gateway/manager.go
```diff
@@ -157,6 +158,22 @@ func (cm *ConsumerManager) getInferKafkaConsumer(modelName string, create bool)
...
 }
 
+func (cm *ConsumerManager) Healthy() error {
+	cm.mu.RLock()
+	defer cm.mu.RUnlock()
+
+	for _, consumer := range cm.consumers {
+		if !consumer.producerIsActive() {
+			return fmt.Errorf("producer %s not active", consumer.producer.String())
+		}
```
lc525, line 168: We can leave this as is now, but it might be worth considering later if there are cases where we
don't make the whole pod not-ready/not live if only one consumer is no longer active for some
time (the processing by other consumers will go on unperturbed). However, I think it's fine now
because we don't really retry creating consumers.

domsolutions: Yeah possibly exceeding some threshold... btw, it looks like we don't call `Close`  on the consumers when shutting down the pod.. that may cause issues with not saving the latest offset and replaying msgs and kafka _may_ think there's still a consumer (although there might be some health checks going on under the hood with the confluent pkg). Does appear we do close the  `producer` though.

k8s/helm-charts/seldon-core-v2-setup/templates/_components-deployments.tpl
domsolutions: ah, there was no special reason for that, let me make it consistent

PR #6746, opened 2025-09-02
feat(dataflow): pipeline parallel loading

RobertSamoilescu:
# Why
## Issues

With the recent improvements to **dataflow-engine** scalability, an existing technical debt item has become more pressing.

Currently, when the scheduler sends `LoadPipeline` gRPC messages to a dataflow-engine replica, these messages are handled serially. Each pipeline load waits for its associated Kafka Streams application to become Ready (including consumer group rebalancing, remote state download, etc.) before the next pipeline is loaded. This creates unnecessary delays when loading multiple pipelines.

## Motivation

This PR introduces a mechanism that allows a dataflow-engine replica to load multiple pipelines **in parallel**, while capping the maximum number of in-flight loads (configurable, e.g., 8). This ensures better concurrency without overwhelming the system.

# What
## Summary of changes

* Initialize a dispatcher with nThreads (configurable).
* Associate each pipeline with its own task queue, where a task can be a create or delete operation.
* Each task queue is processed sequentially, but task queues run in parallel.
* Task queues are conflated: only the most recent submitted task is executed.
  * Example: if submissions are create → delete → create → create → delete (common during rebalancing), only the last delete is processed.
* Introduce a background cleaner that:
  * Closes queues and waits for their tasks to finish (to avoid memory leaks).
  *  Runs every 5 seconds.
  * Cleans up queues marked for deletion after 30s once their deletion tasks have completed.

## Checklist

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
@@ -141,7 +141,6 @@ class Pipeline(
 
     companion object {
         private val logger = noCoLogger(Pipeline::class)
-        const val STATE_DIR_CONFIG = "state.dir"
 
         fun forSteps(
             metadata: PipelineMetadata,
@@ -153,11 +152,8 @@ class Pipeline(
         ): Pair<Pipeline?, PipelineStatus.Error?> {
             val (topology, numSteps) = buildTopology(metadata, steps, kafkaDomainParams)
             val pipelineProperties = localiseKafkaProperties(kafkaProperties, metadata, numSteps, kafkaConsumerGroupIdPrefix, namespace)
-            pipelineProperties[STATE_DIR_CONFIG] = "/tmp/kafka-streams/${metadata.id}"
-
             var streamsApp: KafkaStreams?
             var pipelineError: PipelineStatus.Error?
-
             try {
                 streamsApp = KafkaStreams(topology, pipelineProperties)
             } catch (e: Exception) {
--- a/scheduler/Dockerfile.dataflow
+++ b/scheduler/Dockerfile.dataflow
@@ -27,6 +27,11 @@ RUN mkdir -p /tmp/certs/kafka/broker/ && \
     cp /etc/ssl/certs/ca-bundle.crt /tmp/certs/kafka/broker/ca.crt && \
     chmod -R 777 /tmp/certs/
 
+
+RUN mkdir -p /tmp/kafka-streams \
+    && chown 1000:1000 /tmp/kafka-streams \
+    && chmod 750 /tmp/kafka-streams
+
 USER default
 
 WORKDIR /app


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Cli.kt
```diff
@@ -38,6 +38,7 @@ object Cli {
     val logLevelKafka = Key("log.level.kafka", enumType(*Level.values()))
     val namespace = Key("pod.namespace", stringType)
     val dataflowReplicaId = Key("dataflow.replica.id", stringType)
+    val dispatcherNThreads = Key("dispatcher.n.threads", intType)
```
lc525: I think that this setting should have a more end-user friendly name, relating to what it does rather than to implementation details. I suggest something like `pipeline.ctlops.threads`

scheduler/data-flow/src/main/resources/local.properties
```diff
@@ -1,6 +1,7 @@
 log.level.app=INFO
 log.level.kafka=WARN
 dataflow.replica.id=seldon-dataflow-engine
+dispatcher.n.threads=8
```
lc525: Should also be configured via helm/SeldonConfig, so we'll need to add the setting as an environment var for the container

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -65,9 +77,118 @@ class PipelineSubscriber(
...
+        val processingJob: Job,
+        var isMarkedForDeletion: Boolean = false,
+        var deletionScheduledAt: Long = 0L,
+    )
+
+    init {
+        // Start background cleanup task
+        scope.launch {
+            while (true) {
+                delay(5000L)
```
lc525: I think this is too aggressive (short), especially as it is a "stop the world" (control-plane wise) cleanup operation. I would do the cleanup every 5 minutes or so. I don't think the added memory consumption would be significant, even with thousands of pipelines. However, if I'm missing something let's discuss.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -65,9 +77,118 @@ class PipelineSubscriber(
...
+    private fun startProcessing(
+        scope: CoroutineScope,
+        queue: Channel<Task>,
+    ): Job {
+        return scope.launch {
+            for (task in queue) {
+                try {
+                    task.run()
+                } catch (e: Exception) {
+                    val reason = "Task failed permanently: ${e.message}"
```
lc525: The reason here might end up being displayed in k8s (either as an event on the CR or in the CR status). I would replace "Task" with "Pipeline {create|delete} operation failed:", with create or delete being shown depending on the failed op.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -157,137 +278,30 @@ class PipelineSubscriber(
...
-                // Calling stop() here may be superfluous (depending on the state in which the pipeline is in),
-                // but we want to ensure that we clean up the KafkaStreams state of the pipeline because
-                // otherwise we have issues in re-starting it.
-                // Calling stop() on an already stopped pipeline is safe.
-                previous.stop()
+        queuesMutex.withLock {
+            val existingQueueInfo = queues[metadata.id]
+            if (existingQueueInfo == null) {
+                // Create new queue and processing job
+                val queue = Channel<Task>(Channel.CONFLATED)
```
lc525, line 285: It's nice that this option exists and is built-in

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -157,137 +278,30 @@ class PipelineSubscriber(
...
-        // not in themselves errors, they are not expected at this stage. If the pipeline
-        // is not running here then it can't be marked as ready.
-        if (pipelineStatus !is PipelineStatus.Started) {
-            pipelineStatus.hasError = true
-        }
-        pipelineStatus.log(logger, Level.DEBUG)
-        client.pipelineUpdateEvent(
-            makePipelineUpdateEvent(
-                metadata = metadata,
+        queues[metadata.id]?.queue?.send(
```
lc525: Is it worth having this under the mutex? I'm not sure what happens if the queue disappears between fetching `queues[metadata.id]` and calling `.queue.send` on it.

RobertSamoilescu: you are right. That should be under lock.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
```diff
@@ -152,8 +153,11 @@ class Pipeline(
         ): Pair<Pipeline?, PipelineStatus.Error?> {
             val (topology, numSteps) = buildTopology(metadata, steps, kafkaDomainParams)
             val pipelineProperties = localiseKafkaProperties(kafkaProperties, metadata, numSteps, kafkaConsumerGroupIdPrefix, namespace)
+            pipelineProperties[STATE_DIR_CONFIG] = "/tmp/kafka-streams/${metadata.id}"
```
lc525: Why was this change required? In `localiseKafkaProperties` we call `.withAppId`, which already passes a hash of metadata.id to the application ID, alongside other parts. By default, the state dir is `/tmp/kafka-streams/<app-id>`.

lc525 (approved):
One added comment querying the need to set the kafka app state dir manually.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Pipeline.kt
RobertSamoilescu: Solved by modifying the Dockerfile

PR #6747, opened 2025-09-02
fix(scheduler): race conditions

domsolutions:
## Issues

Since enabling the race detector in the pipeline, more tests have been fialing with either unexpected results or race conditions. This PR might not fix them all as the detector does report false negatives (have run pipeline 3x successfully).

Left comments below to describe the issues.

## Motivation

# What
## Checklist

### DIFF
--- a/scheduler/pkg/envoy/xdscache/seldoncache.go
+++ b/scheduler/pkg/envoy/xdscache/seldoncache.go
@@ -60,7 +60,7 @@ type SeldonXDSCache struct {
 	Clusters               *util.CountedSyncMap[Cluster]
 	Pipelines              *util.CountedSyncMap[PipelineRoute]
 	Routes                 *util.CountedSyncMap[Route]
-	muClustersToAdd        sync.RWMutex
+	mu                     sync.RWMutex
 	clustersToAdd          map[string]struct{}
 	clustersToRemove       map[string]struct{}
 	pipelineGatewayDetails *PipelineGatewayDetails
@@ -291,6 +291,9 @@ func (xds *SeldonXDSCache) ClusterResources() []types.Resource {
 }
 
 func (xds *SeldonXDSCache) RouteResources() []types.Resource {
+	xds.mu.RLock()
+	defer xds.mu.RUnlock()
+
 	defaultRoutes, mirrorRoutes := makeRoutes(xds.Routes, xds.Pipelines)
 	return []types.Resource{defaultRoutes, mirrorRoutes}
 }
@@ -318,8 +321,8 @@ func (xds *SeldonXDSCache) UpdateRoutes(nodeId string) error {
 }
 
 func (xds *SeldonXDSCache) AddClusters() error {
-	xds.muClustersToAdd.Lock()
-	defer xds.muClustersToAdd.Unlock()
+	xds.mu.Lock()
+	defer xds.mu.Unlock()
 
 	var clientSecret *Secret
 	if xds.tlsActive {
@@ -334,7 +337,6 @@ func (xds *SeldonXDSCache) AddClusters() error {
 		if ok {
 			resource := makeCluster(cluster.Name, cluster.Endpoints, cluster.Grpc, clientSecret)
 			resources[cluster.Name] = resource
-
 		}
 		delete(xds.clustersToAdd, clusterName)
 	}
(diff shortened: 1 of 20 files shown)


scheduler/pkg/kafka/dataflow/server_test.go
```diff
@@ -452,12 +460,9 @@ func TestPipelineEvents(t *testing.T) {
 				s.mu.Unlock()
 			}
 
-			err = s.pipelineHandler.SetPipelineState(test.loadReq.Name, test.loadReq.Version, test.loadReq.Uid, test.status, "", sourceChainerServer)
+			err = s.pipelineHandler.SetPipelineState(test.loadReq.Name, test.loadReq.Version, test.loadReq.Uid, test.status, "", "some-source")
```
domsolutions, line 463: having event source as `sourceChainerServer` means the chain server will ignore the msg as not upgrade the subscriber stream i.e. `dataflow-engine`. Changed to `some-source` to cause event to emit

scheduler/pkg/store/memory_status.go
```diff
@@ -105,29 +106,45 @@ func updateModelState(isLatest bool, modelVersion *ModelVersion, prevModelVersio
 	}
 }
 
-func (m *MemoryStore) FailedScheduling(modelVersion *ModelVersion, reason string, reset bool) {
```
domsolutions, line 108: From the changes, as we now return a deep copy of the model to prevents races, this function will no longer work as it modified `*ModelVersion` which will now point to a copy so the memory store won't be updated. Instead we can just look up the model in the store and update.

scheduler/pkg/store/memory.go
```diff
@@ -160,17 +160,26 @@ func (m *MemoryStore) UpdateModel(req *pb.LoadModelRequest) error {
 func (m *MemoryStore) getModelImpl(key string) *ModelSnapshot {
 	model, ok := m.store.models[key]
 	if ok {
-		return &ModelSnapshot{
-			Name:     key,
-			Versions: model.versions, // TODO make a copy for safety?
```
domsolutions, line 165: Note there was a TODO to make a copy.

We have to make a copy as we're getting race condtions throughout the code, where the model is returned to the caller which points to the model within the memory store map, but is no longer protected by the mutex, and another goroutine modifies the model producing the race.

scheduler/pkg/store/test_memory.go
```diff
@@ -0,0 +1,48 @@
...
+// to modify the statuses.
+func NewTestMemory(
+	logger log.FieldLogger,
+	store *LocalSchedulerStore,
+	eventHub *coordinator.EventHub) *TestMemoryStore {
+	m := NewMemoryStore(logger, store, eventHub)
+	return &TestMemoryStore{m}
+}
+
+func (t *TestMemoryStore) DirectlyUpdateModelStatus(model ModelID, state ModelStatus) error {
```
domsolutions: This was created to facilitate tests which were already using a method `SetModelState` which was created purely for tests (but available to non-tests), which bypassed the locking and allowed modifying the model state. This isn't a great solution, creating a struct purely for tests which is accessible to non-tests.. open to ideas for improvements. Tried putting this witinin a test which is fine, but then it's not accessible outside the pkg where it's needed.

scheduler/pkg/envoy/xdscache/seldoncache.go
```diff
@@ -291,6 +291,9 @@ func (xds *SeldonXDSCache) ClusterResources() []types.Resource {
 }
 
 func (xds *SeldonXDSCache) RouteResources() []types.Resource {
+	xds.mu.RLock()
```
lc525, line 294: :+1:  For the race detector, it's not obvious that those functions are only called from IncrementalProcessor, and the locks are there. I think it should be harmless to add them as there would be no contention on acquire/release, and it makes the code simpler to reason about.

scheduler/pkg/kafka/dataflow/server_test.go
lc525: The ignore is there to stop endless loops of events

scheduler/pkg/store/memory_status.go
```diff
@@ -105,29 +106,45 @@ func updateModelState(isLatest bool, modelVersion *ModelVersion, prevModelVersio
...
-		UnavailableReplicas: modelVersion.GetModel().GetDeploymentSpec().GetReplicas() - availableReplicas,
+	model, ok := m.store.models[modelID]
+	if !ok {
+		return fmt.Errorf("model %s not found", modelID)
 	}
-	// make sure we reset server but only if there are no available replicas
-	if reset {
-		modelVersion.SetServer("")
+
+	for _, modelVersion := range model.versions {
```
lc525: It's very likely that the model version that fails scheduling is the last one, so it might be worth to start from the end.

PR #6756, opened 2025-09-09
feat(scheduler): health probes

domsolutions:
## Motivation

We require health probes on `scheduler` for k8s to know if the service is not healthy and if it should be restarted for example.

## Summary of changes

- readiness/liveness probes on gRPC servers: agent server, scheduler server
- created new gRPC health service `HealthCheckService` which both servers implement the interface to provide their health response
- HTTP health server which gRPC calls out to each health endpoint

If TLS is enabled on the data-plane side, the gRPC health reqs need to have the correct cert to successfully make req. The TLS config was baked into the gRPC structs, this was taken out and injected into the structs so the health probes could have access. A new func `CreateControlPlaneTLSOptions` was created to manage creating the TLS cert manager if it's enabled via the necessary env vars. Other service were refactored as part of this PR to use this new func.

## Checklist

### DIFF
--- a/hodometer/cmd/hodometer/main.go
+++ b/hodometer/cmd/hodometer/main.go
@@ -14,8 +14,9 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/seldonio/seldon-core/components/tls/v2/pkg/tls"
 	"github.com/sirupsen/logrus"
+
+	"github.com/seldonio/seldon-core/components/tls/v2/pkg/tls"
 
 	"github.com/seldonio/seldon-core/hodometer/v2/pkg/hodometer"
 )


scheduler/cmd/scheduler/main.go
```diff
@@ -341,3 +393,68 @@ func main() {
...
+}
+
+func initHealthProbe(tlsOptions *tls.TLSOptions, log *log.Logger, config gRPCHealthProbes, healthSrvPort int) (*health_probe.HTTPServer, error) {
+	manager := health_probe.NewManager()
+	for _, probe := range config.probes {
+		var cert *tls.CertificateStore
+		if !probe.plaintText {
+			cert = tlsOptions.Cert
+		}
+		if err := createGRPCHealthProbe(cert, probe.port, manager); err != nil {
```
lc525, line 413: Because all the gRPC server sub-services are local, there was also the option of simply storing some readiness state in the associated data structures, without adding gRPC requests. However, perhaps your idea here is better from the perspective of testing those servers like an external client would. With readiness and liveness probes, we are adding almost 2 RPS background load to the system (beyond the keepalives), but I'm satisfied that the processing for each is minimal enough not to matter. Comparatively, the keepalive RPS grows with the number of Servers, and can become significant if you have hundreds of Servers.

scheduler/pkg/coordinator/hub.go
```diff
@@ -68,6 +68,12 @@ func NewEventHub(l log.FieldLogger) (*EventHub, error) {
 	return &hub, nil
 }
 
+func (h *EventHub) IsClosed() bool {
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+	return h.closed
```
lc525, line 74: Right now, the hub only becomes closed from a `defer` statement in `main.go`. So the check is not strictly necessary, but we can keep it for completeness.

PR #6757, opened 2025-09-09
feat(dataflow): pipeline status update

RobertSamoilescu:
# Why
## Issues

With improved scalability of dataflow-engine being merged, an existing dataflow-engine technical-debt item becomes more important.

In the current implementation, a dataflow-engine replica is the one responsible for setting the pipeline state, and this only occurs during the loading phase of the pipeline: if creating the Kafka streams application succeeds (including the associated rebalancing of consumers on the targeted Kafka topics), then the success message is sent back to the scheduler. If this creation operation fails, then the failure is similarly sent. However, no updates occur when the Kafka streams state changes (i.e transitions back to rebalancing due to a new consumer being added to the group, moves to a failure state because of bugs or connectivity issues to Kafka brokers – say a failure to refresh an auth token, etc).

## Motivation

This PR implements continuous updates of the pipeline status when the Kafka streams application status changes.

# What
## Summary of changes

* Update is sent from the pipeline to the pipeline subscriber with the current status
* Pipeline subscriber creates an update task which is added to the task queues for the given pipeline only if the queued task is not newer (conflict resolution based on timestamp)
* the task is processed from the queue and sends the update status to the scheduler
* two new `PipelineUpdateMessage` were added, `Rebalance` and `Ready`, to inform the scheduler if the pipeline is rebalancing or is ready/failed.  

## Checklist


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/PipelineSubscriber.kt
```diff
@@ -318,11 +320,47 @@ class PipelineSubscriber(
...
+        timestamp: Long,
+        status: PipelineStatus,
+    ) {
+        queuesMutex.withLock {
+            val queueInfo = queues[metadata.id]
+            if (queueInfo != null) {
+                // We need to make sure we don't remove any messages
+                // from the scheduler, thus we have to fetch the
+                // last element in the queue and put it back in case
+                // it has a higher timestamp - there is no peak function
```
lc525:
```suggestion
                // it has a higher timestamp - there is no peek function
```

apis/mlops/scheduler/scheduler.proto
```diff
@@ -425,6 +425,7 @@ message PipelineVersionState {
     PipelineTerminate = 5;
     PipelineTerminating = 6;
     PipelineTerminated = 7;
+    PipelineRebalancing = 8;
```
lc525, line 429: With the introduction of this new state, I believe some of the state transitions in `pkg/store/pipeline/store.go` may also need to be updated. At a minimum, the `PipelineRebalancing` state needs to be considered as a "running" state [here](https://github.com/lc525/seldon-core/blob/6db0eafc991db056f573d838c77047cfa41378c8/scheduler/pkg/store/pipeline/store.go#L305) for the purposes of PipelineGw rebalancing logic. Would you be able to take a look?

RobertSamoilescu: Yes, you are right. Thanks for pointing this out!

PR #6766, opened 2025-09-11
feat(dataflow-engine): health probes

domsolutions:
## Motivation

We require health probes on `dataflow-engine` for k8s to know if the service is not healthy and if it should be restarted for example.

## Summary of changes

- startup probe to ensure connection to `scheduler` is established
- readiness/liveness probes tests the coroutines are still active and running on the pipeline subsciber 

## Checklist

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Main.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Main.kt
@@ -9,6 +9,7 @@ the Change License after the Change Date as each is defined in accordance with t
 
 package io.seldon.dataflow
 
+import com.natpryce.konfig.Configuration
 import io.klogging.noCoLogger
 import io.seldon.dataflow.health.GrpcHealthCheck
 import io.seldon.dataflow.health.HealthServer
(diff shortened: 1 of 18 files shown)


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/health/ServiceHealthCheck.kt
```diff
@@ -0,0 +1,60 @@
...
+    override val name = "service"
+    private val logger = noCoLogger(ServiceHealthCheck::class)
+
+    override suspend fun check(): HealthStatus {
+        return try {
+            // Check if the main coroutine scope is still active
+            val scopeActive = pipelineSubscriber.scope.isActive
+
+            // Check if the dispatcher is still functional
+            val dispatcherActive = pipelineSubscriber.dispatcher.isActive
```
domsolutions, line 31: @RobertSamoilescu @lc525 does it make sense to check this, I presume this always has to be active for the pipeline subscriber to be healthy? Same question for line 28

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Main.kt
```diff
@@ -120,19 +126,43 @@ object Main {
...
+
+        // Liveness checks - internal service health only (no external dependencies)
+        healthService.addLivenessCheck(serviceHealthCheck)
+
+        // Readiness checks - ready to handle requests
+        healthService.addReadinessCheck(serviceHealthCheck)
+
+        healthServer.start()
+
+        addShutdownHandler(subscriber, healthServer)
```
RobertSamoilescu: Maybe move the entire health server set up in a separate function.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/Main.kt
```diff
@@ -120,19 +126,43 @@ object Main {
...
+        val serviceHealthCheck = ServiceHealthCheck(subscriber)
+
+        // Startup checks - ensure critical connections are established
+        healthService.addStartupCheck(gRPCSchedulerCheck)
+
+        // Liveness checks - internal service health only (no external dependencies)
+        healthService.addLivenessCheck(serviceHealthCheck)
+
+        // Readiness checks - ready to handle requests
+        healthService.addReadinessCheck(serviceHealthCheck)
```
RobertSamoilescu: The is currently an unused method called `addHealthCheck` which adds both of them. Consider using that or remove it.

scheduler/data-flow/src/test/kotlin/io/seldon/dataflow/health/GrpcHealthCheckTest.kt
```diff
@@ -0,0 +1,178 @@
...
+                        callback: Runnable,
+                    ) {}
+
+                    override fun <RequestT, ResponseT> newCall(
+                        methodDescriptor: io.grpc.MethodDescriptor<RequestT, ResponseT>,
+                        callOptions: io.grpc.CallOptions,
+                    ) = throw UnsupportedOperationException("Mock implementation")
+
+                    override fun resetConnectBackoff() {}
+                }
```
RobertSamoilescu: This looks like duplicate code. Consider using the `createMockChannel` instead with the appropriate args.

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/health/ServiceHealthCheck.kt
RobertSamoilescu: I think this is ok.

PR #6767, opened 2025-09-11
feat(pipelinegw): pipeline status in pipelinegw

RobertSamoilescu:
# Why
## Issues

Pipeline loading on `pipeline-gw` is never confirmed, meaning we cannot guarantee that the consumer has successfully started. As a result, pipeline inference may hang indefinitely if the consumer for the pipeline output topic fails to start.

## Motivation

Enhance pipeline readiness reporting by exposing more accurate status information from `pipeline-gw` replicas. This allows us to better determine whether a pipeline is truly ready to serve traffic.

# What
## Summary of changes

* Added a new readiness condition for the pipeline PIPELINE GATEWAY.
  * A pipeline is now considered ready only when all conditions are met: MODELS READY, PIPELINE GATEWAY READY, and DATAFLOW READY.
* Updated pipeline-gw components to return acknowledgements (ACKs) to the scheduler when loading/unloading operations succeed.
* The pipeline’s overall status is now derived from the responses of all pipeline-gw replicas where the pipeline is loaded.
  * This follows the same state machine and conflict resolution mechanism as the dataflow engine.
* Refactored the pipeline-gw client implementation for improved readability and maintainability.
* Fixed unloading from pipeline-gw
* Send ack from pipeline-gw to scheduler in a goroutine with retires.

## Note
In the future we should reconsider refactoring the rebalancing code for `dataflow`, `pipeline-gw`, `model-gw` as it shares most of the logic.

## Checklist


operator/scheduler/pipeline.go
```diff
@@ -197,33 +200,11 @@ func (s *SchedulerClient) SubscribePipelineEvents(ctx context.Context, grpcClien
...
-					)
-					pipeline.Status.CreateAndSetCondition(
-						v1alpha1.PipelineReady,
-						false,
-						pv.State.Reason,
-						pv.State.Status.String(),
-					)
-				}
+				// Handle status and pipeline-gw status update
+				reason := combineReasons(pv.State.PipelineGwReason, pv.State.Reason)
```
lc525, line 204: why do we need to combine the reasons if we're setting separate ready conditions below?

scheduler/pkg/envoy/xdscache/seldoncache.go
```diff
@@ -597,3 +616,16 @@ func (xds *SeldonXDSCache) CreateClusterForStreams(clusterPrefix string, streamN
...
 	return &cluster
 }
+
+// RemoveClusterForStreams removes a cluster created for streams by clusterPrefix and isGrpc flag.
+func (xds *SeldonXDSCache) RemoveClusterForStreams(clusterPrefix string, isGrpc bool, logger *logrus.Entry) {
+	xds.mu.Lock()
+	defer xds.mu.Unlock()
+
+	clusterName := getPipelineClusterName(clusterPrefix, isGrpc)
+	logger.Debugf("Removing cluster %s for streams", clusterName)
```
lc525: The message here is a bit confusing when taken out of context (i.e just seen in the logs). Should this say something like "Removing envoy cluster of pipeline-gw endpoints for %s"?

scheduler/pkg/kafka/pipeline/client.go
```diff
@@ -155,78 +157,199 @@ func getSubsriberIp() (string, error) {
...
+		)
+		return
+	}
+
+	message := fmt.Sprintf("Pipeline %s deleted", event.PipelineName)
+	ep.reportSuccess(chainer.PipelineUpdateMessage_Delete, pv, message, event.Timestamp)
+}
+
+func (ep *EventProcessor) handleCreateOrUpdatePipeline(event *scheduler.PipelineStatusResponse) {
+	pv, err := pipeline.CreatePipelineVersionWithStateFromProto(event.Versions[0])
```
lc525, line 296: Could we still log a warning in case the event contains multiple versions? We don't want to silently ignore unexpected input.

scheduler/pkg/server/pipeline_status.go
```diff
@@ -24,6 +27,71 @@ const (
 	addPipelineStreamEventSource = "pipeline.store.addpipelinestream"
 )
 
+func (s *SchedulerServer) PipelineStatusEvent(ctx context.Context, message *chainer.PipelineUpdateStatusMessage) (*chainer.PipelineUpdateStatusResponse, error) {
+	s.pipelineEventStream.mu.Lock()
+	defer s.pipelineEventStream.mu.Unlock()
+
+	logger := s.logger.WithField("func", "SubscribePipelineStatus")
```
lc525: Was this logger field name intentional? (it differs from the actual function name where it resides)

scheduler/pkg/server/pipeline_status.go
```diff
@@ -104,128 +172,176 @@ func (s *SchedulerServer) sendCurrentPipelineStatuses(
...
-		servers := s.pipelineGWLoadBalancer.GetServersForKey(consumerBucketId)
-		s.logger.Debugf("Servers for pipeline %s: %v", pipelineName, servers)
-		s.logger.Debug("Consumer bucket ID: ", consumerBucketId)
+		s.logger.Debugf(
+			"Rebalancing pipeline %s:%d with pipeline gateway state %s",
+			event.PipelineName, event.PipelineVersion, pv.State.PipelineGwStatus.String(),
+		)
 
-		// need to update envoy clusters
+		// no pipeline gateway available yet, publish event for envoy
```
lc525: How do we know at this point that there are no pipeline gateways available at this point? (I just found the comment confusing)

scheduler/pkg/server/pipeline_status.go
```diff
@@ -104,128 +172,176 @@ func (s *SchedulerServer) sendCurrentPipelineStatuses(
...
+	cr.CreateNewIteration(pv.Name, servers)
+
+	// send messages to each pipeline gateway stream
+	for _, pipelineSubscription := range streams {
+		s.logger.Debug("Processing pipeline subscription for ", pipelineSubscription.name)
+		server := pipelineSubscription.name
+		stream := pipelineSubscription.stream
+
+		if contains(servers, server) {
+			s.logger.Debug("Server contains model, sending status update for ", server)
```
lc525:
```suggestion
			s.logger.Debug("pipeline-gateway replica contains pipeline, sending status update for ", server)
```

scheduler/pkg/server/pipeline_status.go
```diff
@@ -243,15 +359,91 @@ func (s *SchedulerServer) sendPipelineEvents(event coordinator.PipelineEventMsg)
...
+	var pipelineEventSources = map[string]struct{}{
+		util.SourcePipelineStatusEvent: {},
+		util.SourceChainerServer:       {},
+	}
+	if _, ok := pipelineEventSources[event.Source]; ok {
+		return
+	}
+
+	// publish event for envoy with empty stream names since
+	// no pipeline gateway is available yet
```
lc525: again, a bit confused on the logic here. It would have made sense if this were under `if len(pipelineGwStreams) == 0` just below. But it's not, so how do we know that at this point we need to send to envoy an empty server list?

lc525: Essentially, I'm trying to understand if for every rebalance we, for a brief period, reset the envoy routes to contain no pipeline-gw servers in the upstream of the cluster associated to that pipeline

lc525 (approved):
lgtm, with some nits and some questions that we should discuss (it's very possible I'm slightly confused on the logic during rebalance)

operator/scheduler/pipeline.go
RobertSamoilescu: This is just the message to inform about the progress of the loading/unloading on pipelingw and dataflow. I didn't want to add a separate column, so I concatenated the reason from both and display them under the message column

PR #6786, opened 2025-09-17
enable Lychee for docs on v2

monica-seldon:
# Why
## Motivation

# What
## Checklist


.github/workflows/gb-docs.yml
```diff
@@ -0,0 +1,20 @@
...
+on:
+  schedule:
+    - cron: '0 0 * * *'  # every day at midnight UTC
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - v2
+
+jobs:
+  lychee-check:
```
lc525: I suggest we run this on PR merges that modify files in `docs-gb` (using something like the [paths-filter action](https://github.com/dorny/paths-filter)), and only run on cron much less frequent (say, once every two weeks) to detect stale external links.

Rajakavitha1:
HI @monica-seldon . The the PR is good to go for me.

.github/workflows/gb-docs.yml
```diff
@@ -0,0 +1,48 @@
...
+    - cron: '0 0 */14 * *'  # every two weeks at midnight UTC
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - v2
+
+jobs:
+  # Check docs changes for PRs only
+  check-docs-changes:
+    if: github.event_name == 'pull_request'
```
tyndria: @monica-seldon  I believe we also need to check that this job is running as part of the SeldonIO repo here, as we don't want this job to be executed in forks
```
github.repository == 'SeldonIO/seldon-core'
```

monica-seldon: Fixed! Added the `github.repository == 'SeldonIO/seldon-core'` check to all jobs in the workflow.
Now the lychee link checker will only run in our official repo, not in forks. This prevents unnecessary CI usage and addresses the security concern.
Thanks for catching that!

PR #6789, opened 2025-09-17
fix(operator): failed update status

RobertSamoilescu:
# Why
## Motivation

# What
## Summary of changes
Fixes errors in operator when an update or get is performed on an already deleted object.

## Checklist

### DIFF
--- a/operator/scheduler/model.go
+++ b/operator/scheduler/model.go
@@ -324,11 +324,6 @@ func (s *SchedulerClient) updateModelStatus(ctx context.Context, model *v1alpha1
 		return err
 	}
 
-	if !existingModel.ObjectMeta.DeletionTimestamp.IsZero() {
-		// Don't update status on deleted objects
-		return nil
-	}
-
 	prevWasReady := modelReady(existingModel.Status)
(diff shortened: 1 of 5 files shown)


operator/scheduler/model.go
```diff
@@ -317,38 +324,47 @@ func (s *SchedulerClient) updateModelStatus(ctx context.Context, model *v1alpha1
 		return err
 	}
 
+	if !existingModel.ObjectMeta.DeletionTimestamp.IsZero() {
+		// Don't update status on deleted objects
```
lc525: The condition reflects that the object is in the process of being deleted rather than already deleted. The comment should reflect that. I'm in two minds about this change, because it means that if a delete gets stuck, we give even less information to the end-user about what is going on.

lc525 (approved):
lgtm; The only change I'm a bit unsure of is the stopping of CR status updates while a model is being deleted. Let's briefly discuss the reasoning there before merging.

operator/scheduler/model.go
RobertSamoilescu: Removed this bit

PR #6795, opened 2025-09-23
fix(dataflow): deprecated use of kafka streams Transformer classes

lc525:
Dataflow used a deprecated API for executing transformations on KStreams. 
This prevented us from updating to kafka-streams 8.0.0

In this PR:

- I've replaced the (deprecated) [Transformer-based API](https://docs.confluent.io/platform/current/streams/javadocs/javadoc/org/apache/kafka/streams/kstream/Transformer.html) to the newer [Processor API](https://docs.confluent.io/platform/current/streams/javadocs/javadoc/org/apache/kafka/streams/processor/api/Processor.html). For all the transformations required by dataflow engine, a class inheriting from `FixedKeyProcessor` is enough: we do not transform the key of the kafka message in any way.
- I've added a reusable kafka header processor (`headers/PipelineHeaderProcessor.kt`) that can be reused by multiple transformation adjusting the headers with which a kafka message is published.
- I've added a `MockFixedKeyProcessorContext` class, aimed at being able to test the BatchProcessor update to the Processor API

Enabled by the code changes above, we update from `kafka-streams:7.7.0-ccs` to `kafka-streams:8.0.0-ccs`

#### Notes to reviewer:
Tested with local deployment of dataflow-engine dev build into kind cluster and running `samples/smoke-tests.sh`. Also verified that the resulting kafka messages continue to be produced with the right headers.


scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/StreamTransforms.kt
```diff
@@ -9,33 +9,51 @@ the Change License after the Change Date as each is defined in accordance with t
...
 }
 
-fun <T> KStream<T, TRecord>.headerSetter(
+fun <T> KStream<T, TRecord>.headerAdjust(
     pipelineName: String,
     pipelineVersion: String,
 ): KStream<T, TRecord> {
     return this
-        .transformValues(ValueTransformerSupplier { PipelineHeaderSetter(pipelineName, pipelineVersion) })
+        .processValues({
```
MiguelAAe, line 51: This way is quite tidy

PR #6799, opened 2025-09-23
feat(modelgw): modelgw status update

RobertSamoilescu:
# Why
## Issues

Model loading on `model-gw` is never confirmed, meaning we cannot guarantee that the consumer has successfully started. As a result, pipeline inference may hang indefinitely if there isn't any model-gw instance which can compute the inference requests for the particular pipeline step.

## Motivation
Enhance pipeline readiness reporting by exposing more accurate status information from `model-gw` replicas. This allows us to better determine whether a pipeline is truly ready to serve traffic.

# What
## Summary of changes

* Added a new readiness condition on the models - MODEL GATEWAY
** The condition MODELS READY on the Pipeline is true only if the MODEL GATEWAY is ready and the models are ready (loaded on the servers)
** Note that MODEL GATEWAY is not necessary to be true to server a model - it is only used for the pipeline readiness status
* Updated model-gw component to return acknowledgements (ACKs) to the scheduler when loading/unloading operations succeed
* *  It follows the same state machine as for pipeline-gw and dataflow-engine - to be unified in the future. Thus, to additional states were added for the model (i.e., ModelCreate, ModelTerminate)
* Refactored the model-gw client implementation for improved readability and maintainability.
* * Explicit operations for create/delete are now sent to the model-gw to "load" a model.
* Decoupled model-gw loading/unloading/rebalancing from the actual model status. Before the model status was considered when doing those operation - removes some edgecases like [this](https://github.com/SeldonIO/seldon-core/blob/8183d231457732f390fd78604ea6f4b110bd04a2/scheduler/pkg/kafka/gateway/client.go#L172-L198).
* Sending ack from model-gw to scheduler in a goroutine with retries.

## Checklist


scheduler/pkg/store/mesh.go
```diff
@@ -62,7 +62,9 @@ type ModelVersion struct {
 
 type ModelStatus struct {
 	State               ModelState
+	ModelGwState        ModelState
```
domsolutions, line 65: Is it correct that a model gw has the same state as a model? i.e. is it possible for a model gw to be in all these states?

```golang

const (
	ModelStateUnknown ModelState = iota
	ModelProgressing
	ModelAvailable
	ModelFailed
	ModelTerminating
	ModelTerminated
	ModelTerminateFailed
	ScheduleFailed
	ModelScaledDown
	ModelCreate
	ModelTerminate
)
```

scheduler/pkg/kafka/gateway/client_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+
+	pb "github.com/seldonio/seldon-core/apis/go/v2/mlops/scheduler"
+	"github.com/seldonio/seldon-core/components/tls/v2/pkg/tls"
+)
+
+type MockGrpcClient struct {
```
domsolutions, line 26: Could we not use `mockgen` here to generate the mocks? like we do [here](https://github.com/SeldonIO/seldon-core/blob/c2e7876952ae7c05872de36f938cca1d86c1a77e/scheduler/pkg/health-probe/health.go#L25) for example

scheduler/pkg/store/mesh.go
RobertSamoilescu: It uses a subset of those states. Also the state machine/transitions are different from the ones for the model. Those states are analogous to the ones in pipelinegw and dataflow engine, and probably unifying those in the future will make sense.

RobertSamoilescu:
> If a model gw fails to load the model and reports back to the scheduler it failed, does the scheduler attempt to retry loading? I couldn't see it does that, but might be missing it.
> 
> Say if we have 2 replicas of `model-gw` and one fails to load the model but the other succeeds, would we still report the pipeline is not ready? As the other replica should still be able to handle all the paritions and serve reqs.

There is no retry mechanism implemented yet.

In the scenario of 2 modelgw with 1 replica failing to load and the other successfully load, we would still be able to serve traffic. In general if there is one replica which managed to load successfully, we set the status to ready after receiving all the confirmation messages, successful or not. The logic is in the conflict resolutioner.

PR #6801, opened 2025-09-23
fix(model-gw): graceful shutdown of kafka consumers

domsolutions:
## Motivation

On shutdown of `model-gw` kafka consumers were not being closed. It was noticed during load tests, if a gateway shutdown, there were delays in re-balancing and msgs being replayed.

## Summary of changes

- commit latest offset before closing connection
- concurrently close each consumer 
- reduce poll timeout from 10 seconds to 1 second  on both `model-gw` and `pipeline-gw` so we don't unnecessarily delay the shutdown

## Checklist


RobertSamoilescu (approved):

scheduler/pkg/kafka/pipeline/kafkamanager.go
```diff
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	pollTimeoutMillisecs           = 10000
+	pollTimeoutMillisecs           = 1000
```
lc525, line 42: Have you noticed any reduced throughput due to the reduced poll timeout? I think 1000 _should_ be fine, but useful to know if it has any impact. With 50 workers all executing poll in a loop, we may see a small increase in CPU consumption of the pod when there are no messages to consume. But it might not be noticeable, and 10s was certainly too large.

lc525 (approved):
lgtm, just one comment/question

PR #6802, opened 2025-09-23
fix(pipeline-gw): shutdown - wait for envoy to remove route

domsolutions:
## Motivation

Under load when `pipeline-gw` shutsdown, on occasion we'd get 503s as envoy would still proxy reqs while HTTP/gRPC were gracefully shutting down and no longer accepting new connections.

## Summary of changes

- sleep for `1 second` after closing connection to scheduler which will trigger envoy cluster update

## Checklist

### DIFF
--- a/scheduler/cmd/pipelinegateway/main.go
+++ b/scheduler/cmd/pipelinegateway/main.go
@@ -65,6 +65,10 @@ const (
 	serviceTag                   = "seldon-pipelinegateway"
 )
 
+const (
+	waitForRouteToBeRemovedFromEnvoy = 1 * time.Second
+)
+
 var (
 	httpPort               int
 	grpcPort               int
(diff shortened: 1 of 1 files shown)


scheduler/cmd/pipelinegateway/main.go
```diff
@@ -65,6 +65,10 @@ const (
 	serviceTag                   = "seldon-pipelinegateway"
 )
 
+const (
+	waitForRouteToBeRemovedFromEnvoy = 1 * time.Second
```
lc525, line 69: This might not be enough with multiple envoy replicas, but it's a good point to start from. Did you by any chance run with multiple envoy replicas already?

domsolutions: Let me test that scenario, only tested with 1

PR #6811, opened 2025-09-26
chore(helm): 3GB default dataflow memory req/limit

domsolutions:
# Why

Recently been observed `dataflow-engine` can take up to  `1.5GB` when under load and also seemingly idle. 

# What

## Checklist


lc525 (approved):
lgtm

PR #6813, opened 2025-09-29
feat(tests): pipeline scalability tests

domsolutions:
## Motivation

We require full E2E tests for pipelines to ensure there's no regression and tests which scale different services to ensure reqs don't break during scaling and RPS increases in certain scaling scenarios.

## Summary of changes

- new ansible role to install `ClickHouse` to give ability to view  and interrogate OTEL trace logs via SQL
- new OTEL trace `WaitForWorker` to reveal how long a `model-gw` kafka msg has to wait for an available worker to send inference req
- multiple new tests with pipelines

## Checklist

### DIFF
--- a/scheduler/pkg/kafka/gateway/infer.go
+++ b/scheduler/pkg/kafka/gateway/infer.go
@@ -388,7 +388,7 @@ func (kc *InferKafkaHandler) RemoveModel(modelName string, cleanTopicsOnDeletion
 	delete(kc.loadedModels, modelName)
 	delete(kc.subscribedTopics, kc.topicNamer.GetModelTopicInputs(modelName))
 	if len(kc.subscribedTopics) > 0 {
-		kc.logger.WithField("topics", kc.subscribedTopics).Info("Re-subscribing to remaining topics after model deletion")
+		kc.logger.WithField("topics", kc.subscribedTopics).Debug("Re-subscribing to remaining topics after model deletion")
 		err := kc.subscribeTopics()
 		if err != nil {
 			kc.logger.WithError(err).Errorf("failed to subscribe to topics")
--- a/tests/k6/components/settings.js
+++ b/tests/k6/components/settings.js
@@ -47,10 +47,6 @@ function modelType() {
     return ["iris"]
 }
 
-function skipTeardown() {
-    return String(__ENV.SKIP_TEARDOWN).toLowerCase() === 'true';
-}
-
 function skipSetup() {
     return String(__ENV.SKIP_SETUP).toLowerCase() === 'true';
 }
@@ -481,7 +477,6 @@ export function getConfig() {
         "seldonConfigName": seldonConfigName(),
         "requestIDPrefix" : requestIDPrefix(),
         "skipSetup" : skipSetup(),
-        "skipTearDown" : skipTeardown(),
         "debug": debug(),
         "useKubeControlPlane": useKubeControlPlane(),
         "schedulerEndpoint": schedulerEndpoint(),
--- a/tests/k6/components/v2.js
+++ b/tests/k6/components/v2.js
@@ -137,10 +137,3 @@ export function setupK6(setup, forceConfig = null) {
     }
     return setup(config)
 }
-
-export function tearDownK6(config, callbackTearDown) {
-    if (config.skipTearDown) {
-        return
-    }
-    return callbackTearDown(config)
-}
--- a/tests/k6/scenarios/pipeline/1_model_pipeline.js
+++ b/tests/k6/scenarios/pipeline/1_model_pipeline.js
@@ -92,10 +92,8 @@ export default function (config) {
 }
 
 export function teardown(config) {
-    tearDownK6(config, function (config) {
-        const ctl = connectControlPlaneOps(config)
-        ctl.unloadServerFn(serverName, true, false)
-        ctl.unloadModelFn(modelName + "-1", true)
-        ctl.unloadPipelineFn(pipelineName, false)
-    })
+    const ctl = connectControlPlaneOps(config)
+    ctl.unloadServerFn(serverName, true, false)
+    ctl.unloadModelFn(modelName + "-1", true)
+    ctl.unloadPipelineFn(pipelineName, false)
 }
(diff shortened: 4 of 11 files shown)


scheduler/pkg/kafka/gateway/infer.go
```diff
@@ -388,6 +388,7 @@ func (kc *InferKafkaHandler) RemoveModel(modelName string, cleanTopicsOnDeletion
 	delete(kc.loadedModels, modelName)
 	delete(kc.subscribedTopics, kc.topicNamer.GetModelTopicInputs(modelName))
 	if len(kc.subscribedTopics) > 0 {
+		kc.logger.WithField("topics", kc.subscribedTopics).Info("Re-subscribing to remaining topics after model deletion")
```
lc525: Could we move this to `debug`? I'm thinking it would be too verbose for normal operation.

tests/k6/components/v2.js
```diff
@@ -88,17 +116,31 @@ export function disconnectV2Grpc() {
...
+export function generateDataFlowName(modelName, isPipeline) {
+    if (!isPipeline) {
         return modelName
-    } else {
-        if (suffix=="pipeline") {
-            // we add -pipeline to the model name in the case of constructing a pipeline
-            return generatePipelineName(modelName)+"."+suffix
-        } else {
-            // in this dataflow case we access the model directly
-            return modelName+"."+suffix
```
lc525, line 100: The reason for the existence of `DATAFLOW_TAG=model` (in the `kmodel` k8s overlay) is that pipeline-gw has a special mode where it can serve inference requests  targeting a model, via kafka. i.e You send an inference request to `pipeline-gw` with the `seldon-model` header set to `my_model_name.model` and `pipeline-gw` directly writes to the model input topic and waits to read from the model output topic. 

In benchmarking, this would effectively allow us to understand the overhead added by kafka streams (dataflow-engine), when comparing to an inference sent to a pipeline wrapping a single model.

I would like to keep this as an option, but I agree that DATAFLOW_TAG was a poor naming for that variable.

tests/k6/components/settings.js
```diff
@@ -221,6 +226,94 @@ function experimentNamePrefix() {
...
+}
+
+function modelGwReplicas() {
+    if (__ENV.MODEL_GW_REPLICAS) {
+        return parseInt(__ENV.MODEL_GW_REPLICAS)
+    }
+    return 1
+}
+
+function useGRPC() {
```
lc525, line 288: Previously, one would use `Number(config.inferGrpcIterations)>0` instead.

tests/k6/scenarios/pipeline/scalability/1_model_pipeline_all_svcs_scaled.js
```diff
@@ -0,0 +1,105 @@
...
+        "useKubeControlPlane" : true,
+    })
+}
+
+export default function (config) {
+    const inferPayload1 = getModelInferencePayload(modelType, 1)
+    inferHttp(config.inferHttpEndpoint, pipelineName, inferPayload1.http, true, true, config.debug, config.requestIDPrefix)
+}
+
+export function teardown(config) {
```
lc525, line 91: So, this would not be called when `--no-teardown` is passed in the overlay (as I see the it's in the submitted overlays). It would be a bit confusing to set things with both `--no-teardown` and `SKIP_TEARDOWN=false`, because teardown will end up being skipped.

Not a major issue, but just wanted to point out that the `SKIP_TEARDOWN` env var introduces some inconsistency (we should prefer one method)

tests/k6/scenarios/pipeline/scalability/1_model_pipeline_all_svcs_scaled.js
```diff
@@ -0,0 +1,105 @@
...
+        const modelServerReplicas = 5
+
+        const server = generateServer(serverName, "mlserver", modelServerReplicas, modelServerReplicas, modelServerReplicas)
+        ctl.unloadServerFn(server.object.metadata.name, true, true)
+        ctl.loadServerFn(server.yaml, server.object.metadata.name, true, true, 30)
+
+
+        const pipeline = generateMultiModelPipelineYaml(1, modelType, pipelineName, modelName, modelParams, config.modelName, modelServerReplicas, serverName)
+        pipeline.modelCRYaml.forEach(model => {
+            // we MUST unload/load models before loading server, otherwise it causes Server to scale down to 1, bug or design?
```
lc525: Server scales down to the minimum number of replicas, but at least `minReplicas` such that no server replicas above `minReplicas` are empty (no models). The scale-down action is triggered on model replica unload.

lc525 (approved):
lgtm, with some comments.

I should merge [https://github.com/SeldonIO/seldon-core/commits/v2-bench-k6-db-output/`](https://github.com/SeldonIO/seldon-core/tree/v2-bench-k6-db-output) after this, which enables the full load test raw output to be sent to clickhouse as well (i.e perf characteristics of each request made by k6).

tests/k6/components/v2.js
domsolutions: Ok let me fix this in a separate PR so it's more clear what my changes are and to unblock your pending PR

tests/k6/components/settings.js
domsolutions: But we define the iterations with the `-i` on the `k6` cmd options.. going back to your point above, might it be confusing having several options to define iterations?

tests/k6/scenarios/pipeline/1_model_pipeline.js
```diff
@@ -0,0 +1,99 @@
+import { dump as yamlDump } from "https://cdn.jsdelivr.net/npm/js-yaml@4.1.0/dist/js-yaml.mjs";
+import * as k8s from '../../components/k8s.js';
+import {connectControlPlaneOps,
+} from '../../components/utils.js'
+import {generateMultiModelPipelineYaml, getModelInferencePayload} from '../../components/model.js';
+import {connectV2Grpc, inferGrpc, inferHttp, setupK6, tearDownK6} from "../../components/v2.js";
```
MiguelAAe, line 6:
```suggestion
import {connectV2Grpc, inferGrpc, inferHttp, setupK6} from "../../components/v2.js";
```

delete tearDownK6 since no longer in `v2.js`

PR #6814, opened 2025-09-29
fix(agent): force disable auto-scaling of models on agent/scheduler

domsolutions:
## Motivation

Auto-scaling of Models does not work when there are no Models loaded on a Server and a scale up of models is required. This is potentially causing customer's issues where models scale down to 0, during quiet load, but when load increases models stays at 0. Force disabling until rectified. 

## Summary of changes

- force disabled on scheduler, will return error if client attempts to connect to model trigger scaling stream service
- force disable on agent

## Checklist


lc525 (approved):
lgtm

PR #6824, opened 2025-09-30
fix(operator): SubscribeControlPlane failure blocking loading other CRs

domsolutions:
## Motivation

When the scheduler restarts and operator remains up, operator will connect to scheduler via gRPC stream `SubscribeControlPlane` and the scheduler will request the CRs we support (servers, experiments, models, pipelines). 
There are several issues with this:

- If a Server has an invalid scaling spec, it will cause all other Servers not to be processed. It will also backoff retry forever, even though the scaling spec will remain the same until changed by customer for example.
- If there's a failure loading any of the non-Server resources, the other resources are not processed

## Summary of changes

- If all Servers fail scaling spec validation, a sentinel error is returned `ErrScalingSpec` which the operator will recognise and not backoff retry. It's up to user to correct the spec, which will trigger reconcile and notify the scheduler of the Server if it passes the scaling validation. If there's at least one Server which passed scaling check, it will be sent to the scheduler, invalid ones will be ommitted from the request. 
- Instead of requesting all resources which includes models, pipelines, and experiments, we request each one individually so the errors can be managed and isolated without affecting the other resources
- Each steam msg receved by the operator is run in its own goroutine to avoid any blocking behaviour due to errors   

## Checklist


operator/scheduler/client.go
```diff
@@ -143,36 +146,45 @@ func (s *SchedulerClient) startEventHanders(namespace string, conn *grpc.ClientC
...
+		if err != nil {
+			return fmt.Errorf("failed sending models: %w", err)
 		}
-		return err
+		return nil
 	default:
-		s.logger.Info("Unknown operation", "operation", operation)
-		return fmt.Errorf("Unknown operation %v", operation)
+		s.logger.Error(errors.New("unknown operation"), "Unsupported operation", "operation", operation)
+		// we don't want to retry unknow operations
```
lc525:
```suggestion
		// we don't want to retry unknown operations
```

operator/scheduler/server.go
```diff
@@ -23,36 +24,45 @@ import (
...
 				MaxReplicas: 0,
 			}
 		} else {
-			scalingSpec, err = v1alpha1.GetValidatedScalingSpec(server.Spec.Replicas, server.Spec.MinReplicas, server.Spec.MaxReplicas)
+			scalingSpec, err = internal.GetValidatedScalingSpec(server.Spec.Replicas, server.Spec.MinReplicas, server.Spec.MaxReplicas)
 			if err != nil {
-				return fmt.Errorf("failed to validate scaling spec: %v", err)
+				logger.Error(err, "Server failed scaling spec, omitting from server notify list", "server", server.Name)
+				errs = e.Join(fmt.Errorf("server %s failed scaling spec check: %w", server.Name, err))
+				continue
```
lc525, line 65: So, this is certainly miles better then what we currently have, and we should merge as-is.

However, I think there might be an edge case here wrt to the invalid spec (and the fact that we're not rejecting the invalid CR outright). The timing sequence is the following:

1. CR changed to invalid scaling spec
2. scheduler down / restarting
3. scheduler reconnects and requests list of servers from controller
4. controller skips server(s) with invalid scaling spec, so scheduler is not waiting for connections from them
5. however, the CR being changed to an invalid scaling spec did not modify the actual replicas / pods for that server
6. scheduler starts scheduling early because it receives the expected number of connects from agents before actually all of them connect (the agents from replicas of the server with the invalid CR still connect to the restarted scheduler).

Perhaps there's something already stopping this, but worth having a look for the next release (and validating the CRs via CEL)

lc525 (approved):
lgtm

operator/scheduler/server.go
domsolutions: Good spot, think that would be the case for a Server already created and its StatefulSet created, but not for the case where it was first created. I wonder for the next release if we just implement the CEL validation, that way we don't have to account for these edge cases and allowing for an invalid state to exist... although we'd have to be mindful of deployments that exist already in an invalid state.

PR #6828, opened 2025-10-02
fix(modelgateway): Number of partitions retrieval

MiguelAAe:
# Why
## Issues
When the config from kafka `numPartitions` is not present or isn't a string the model gateway would fail to create an infer consumer due to the casting of an integer or an empty value from the Topics config.

## Motivation

# What
## Summary of changes
* update the kafka config logic retrieval of topics  to check for either strings or integer and positive integer
* updated error messages when erroneous config is loaded
## Checklist

### DIFF
--- a/scheduler/pkg/kafka/gateway/infer.go
+++ b/scheduler/pkg/kafka/gateway/infer.go
@@ -116,7 +116,7 @@ func NewInferKafkaHandler(
 	}
 	replicationFactor, err := GetIntConfigMapValue(topicsConfigMap, replicationFactorKey, defaultReplicationFactor)
 	if err != nil {
-		return nil, fmt.Errorf("error getting replication factor: %v", err)
+		return nil, fmt.Errorf("invalid Kafka topic configuration: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


scheduler/cmd/scheduler/main.go
```diff
@@ -279,10 +278,9 @@ func main() {
 		logger.WithError(err).Fatal("Failed to load Kafka config")
 	}
 
-	numPartitions, err := strconv.Atoi(kafkaConfigMap.Topics["numPartitions"].(string))
+	numPartitions, err := kafkaConfigMap.GetNumPartitions()
```
lc525: A note that in the new setup (the PR that I'm working on) the number of partitions are no longer fetched here. However, we might be able to reuse the other changes in components/kafka if pipeline-gw/model-gw require the number of partitions.

MiguelAAe: I will delete the scheduler code and then only keep the model gateway checks

lc525: It's fine to leave as is, because [my PR](https://github.com/SeldonIO/seldon-core/pull/6833) removes the code and I'll need to rebase anyway

PR #6830, opened 2025-10-02
fix(agent/rclone): rclone OOM

domsolutions:
## Motivation

When multiple models are being loaded onto a Server, it can cause `rclone` to OOM when downloading the models. Also errors have been observed where rclone appears not to have OOM'd and we get `connection refused`

## Summary of changes

- introduced a expo backoff retry to download models, max time of  `30 seconds`
- new liveness probes which will restart rclone if unresponsive 
- set `--max-buffer-memory=64M` to mitigate OOM 
- upgrade rclone to latest `rclone:1.71.1` which includes bug fixes for `--max-buffer-memory`
- increase memory limit on rclone to `256 MB`

## Checklist


k8s/helm-charts/seldon-core-v2-setup/templates/_components-statefulsets.tpl
```diff
@@ -1316,10 +1316,20 @@ metadata:
...
+      - --rc-addr=0.0.0.0:5572
+      - --max-buffer-memory=$(MAX_BUFFER_MEMORY)
+      command:
+      - rclone
+      env:
       - name: RCLONE_LOG_LEVEL
         value: '{{ hasKey .Values.serverConfig.rclone "logLevel" | ternary .Values.serverConfig.rclone.logLevel
           .Values.logging.logLevel | upper }}'
+      - name: MAX_BUFFER_MEMORY
+        value: 16M
```
lc525: Can we increase this up a bit? I.e 64M? Just because 16M will mean files will be downloaded sequentially (only space for 1 concurrent transfer before exceeding the max buffer memory).

k8s/helm-charts/seldon-core-v2-setup/templates/_components-statefulsets.tpl
```diff
@@ -1339,12 +1339,15 @@ spec:
             path: terminate
             port: 9007
       livenessProbe:
-        failureThreshold: 2
+        exec:
```
lc525, line 1342: Where there issues with the TCP probe? the reason I'm asking is because we know for sure that's the bit that fails (we don't know if in that state this rclone command would fail or not)

domsolutions: Well the reason I want to use `rclone rc rc/noop` is it actually make a HTTP POST request to the running instance, which is what the agent expects to be able to do, so I saw it as more reflective that rclone was able to work as needed.

lc525: Right, good point. My thought here was that `rc/noop` might not exercise the `config/get` codepath that was failing. But it's very very likely that the failure came from the entire rc endpoint being down (given the response was a TCP disconnect). So let's go with the approach you've proposed.

PR #6831, opened 2025-10-02
feat(dataflow): added fullJitterBackoff ack for pipeline status

RobertSamoilescu:
# Why
## Motivation

# What
## Summary of changes
Added fullJitterBackoff to dataflow engine ack for pipeline status. In case the ack fails, the scheduler will keep the pipeline in creating/terminating state, even if some replicas have sent the ack successfully. This is trying to prevent that from happening.

## Checklist

### DIFF
--- a/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Task.kt
+++ b/scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Task.kt
@@ -9,6 +9,8 @@ the Change License after the Change Date as each is defined in accordance with t
 
 package io.seldon.dataflow.kafka
 
+import com.github.michaelbull.retry.policy.fullJitterBackoff
+import com.github.michaelbull.retry.retry
 import io.klogging.Klogger
 import io.klogging.Level
 import io.seldon.dataflow.PipelineSubscriber
@@ -25,6 +27,7 @@ abstract class Task(
     val timestamp: Long,
     private val name: String,
     val operation: PipelineOperation,
+    private val logger: Klogger,
 ) {
     abstract suspend fun run()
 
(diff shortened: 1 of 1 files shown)


domsolutions (approved):

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Task.kt
```diff
@@ -25,23 +27,32 @@ abstract class Task(
...
+                        success = success,
+                        reason = reason,
+                        timestamp = timestamp,
+                        stream = name,
+                    ),
+                )
+            }
+        } catch (e: Exception) {
+            logger.warn("Failed to send pipeline update event $operation after retries", e)
+        }
```
lc525: A nice thing about this retry library is that it has composable retry policies. So what is written above should more easily be expressed as:
```suggestion
        try {
            retry(fullJitterBackoff(100L..3_200L) + stopAtAttempts(5)) {
               pipelineSubscriber.client.pipelineUpdateEvent(
                    pipelineSubscriber.makePipelineUpdateEvent(
                        metadata = metadata,
                        operation = operation,
                        success = success,
                        reason = reason,
                        timestamp = timestamp,
                        stream = name,
                    ),
                )
            }
        } catch (e: Exception) {
            logger.warn("Failed to send pipeline update event $operation after retries", e)
        }
```

scheduler/data-flow/src/main/kotlin/io/seldon/dataflow/kafka/Task.kt
```diff
@@ -9,6 +9,8 @@ the Change License after the Change Date as each is defined in accordance with t
 
 package io.seldon.dataflow.kafka
 
+import com.github.michaelbull.retry.policy.fullJitterBackoff
```
lc525, line 12:
```suggestion
import com.github.michaelbull.retry.policy.fullJitterBackoff
import com.github.michaelbull.retry.policy.stopAtAttempts
```

PR #6833, opened 2025-10-03
fix(scheduler): Add scaling config and upgrade paths

lc525:
# Why
## Motivation
Need for a smooth upgrade path from 2.9.x to 2.10.0. Previously:
- existing SeldonConfig CRs would not smoothly work after upgrading to 2.10
- changes in SeldonConfig (for example to scalability options) would not be picked-up by the rest of the install

The smooth upgrade plan is the following:
- When using an existing SeldonConfig CR, we keep everything the same as in 2.9.1 (i.e dataflow, model-gw reading number of partitions from env vars). The scheduler considers the old numPartitions setting by fetching it via a ConfigMap from `SeldonConfig.config.scalingConfig.Pipelines.MaxShardCountMultiplier`. An empty scaling config in `SeldonConfig.config.scalingConfig` is interpreted as if `SeldonConfig.config.scalingConfig.Pipelines.MaxShardCountMultiplier == 1` (one processing replica per pipeline/model)
- Any other settings in scalingConfig can be set but have no effect in 2.10. Similarly, settings in `SeldonConfig.config.kafkaConfig.Topics` can be set but have no effect -- **this is to give customers an opportunity to upgrade their CRs in preparation for 2.11 which will use those settings**
- In particular, we will ask clients to upgrade their SeldonConfig to contain settings such as general autoscaling options, kafka number of partitions, etc, to be in sync with the values they have already set in other places (env vars for various pods). Again, modifing those will have no effect until 2.11
- In 2.11, we switch to activating those values, or considering defaults if they are missing. Customers that haven't upgraded their CRs incrementally in 2.10 will need to take manual steps to configure the upgrade to 2.11. Other customers that have already set the right options in SeldonConfig will be able to smoothly upgrade to 2.11
- In 2.10, customers will be able to set/modify `SeldonConfig.config.scalingConfig.Pipelines.MaxShardCountMultiplier` and this change will propagate automatically through our microservices (this changes the way pipelines/models are sharded over the various microservices: it controls how many replicas of pipeline-gw process a given pipeline, how many dataflow-engine replicas process a given pipeline, and how many model-gw replicas process a given model)
- In 2.11 the other settings will work the same (in terms of propagation of changes throughout the Core 2 install)

# What
## Summary of changes
- Introduce a ScalingConfig setting in SeldonConfig, the settings of which are read by the scheduler _only_
- No longer rely on `SeldonConfig.Config.KafkaConfig.Topics.*` settings for microservices (aim to port over to those in 2.11) - leave the config options in place so that customers can upgrade their SeldonConfig progressively
- Add configuration watchers to react to ScalingConfig changes.

## Checklist

### Notes for reviewers
The crux of the changes are in (it would help to review those first):
- `operator/apis/mlops/v1alpha1/seldonconfig_types.go`
- `scheduler/pkg/config/watcher.go`
- `scheduler/pkg/scaling/config/config.go`
- `scheduler/cmd/scheduler/main.go`
- `scheduler/pkg/kafka/dataflow/server.go`
- `scheduler/pkg/server/server.go`

### DIFF
--- a/scheduler/data-flow/build.gradle.kts
+++ b/scheduler/data-flow/build.gradle.kts
@@ -58,10 +58,10 @@ dependencies {
     implementation("io.kubernetes:client-java:24.0.0")
 
     // HTTP server for health probes
-    implementation("io.ktor:ktor-server-core:3.0.1")
-    implementation("io.ktor:ktor-server-netty:3.0.1")
-    implementation("io.ktor:ktor-server-content-negotiation:3.0.1")
-    implementation("io.ktor:ktor-serialization-kotlinx-json:3.0.1")
+    implementation("io.ktor:ktor-server-core:3.3.0")
+    implementation("io.ktor:ktor-server-netty:3.3.0")
+    implementation("io.ktor:ktor-server-content-negotiation:3.3.0")
+    implementation("io.ktor:ktor-serialization-kotlinx-json:3.3.0")
     implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.7.3")
     implementation("com.charleskorn.kaml:kaml:0.61.0")
 


scheduler/pkg/scaling/config/config.go
```diff
@@ -0,0 +1,142 @@
...
+		modelsCopy = &modelsDeepCopy
+	} else {
+		modelsCopy = &DefaultModelScalingConfig
+	}
+
+	if sc.Servers != nil {
+		serversDeepCopy := *sc.Servers
+		serversCopy = &serversDeepCopy
+	} else {
+		serversCopy = &DefaultServerScalingConfig
```
domsolutions: Can we maybe put a log warn here saying the default is being used? I think it'd be helpful, espeically if customer's are noticing performance issues, although not sure it makes sense having a logger attached to `ScalingConfig` so maybe it should be logged elsewhere

scheduler/pkg/config/watcher.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (cw *ConfigWatcher[T, PT]) Close() error {
+	if cw.fileWatcherDone != nil {
+		close(cw.fileWatcherDone)
+	}
+	if cw.configMapWatcherDone != nil {
+		close(cw.configMapWatcherDone)
+	}
+	if cw.watcher != nil {
+		return cw.watcher.Close()
```
domsolutions: Is it correct we return here and don't continue to close the listeners?

operator/controllers/reconcilers/seldon/runtime_reconciler.go
```diff
@@ -100,7 +98,7 @@ func validateScaleSpec(
 	var maxConsumers int32 = defaultMaxConsumers
 	if maxConsumersEnvName != "" {
 		maxConsumersEnv := getEnvVarValue(component.PodSpec, maxConsumersEnvName, "")
-		maxConsumers, err = ParseInt32(maxConsumersEnv, defaultMaxConsumers)
+		maxConsumers, err := ParseInt32(maxConsumersEnv, defaultMaxConsumers)
```
RobertSamoilescu: I think in this context, `maxConsumers` will be a new variable and it won't update the var defined before the if statement. I think you need to declare the `err` as `var err error` to avoid the compilation error if you replace `:=` with `=`.

scheduler/pkg/config/watcher.go
```diff
@@ -0,0 +1,258 @@
...
+	namespace             string
+	useConfigMapInformer  bool
+	configMapName         string
+	watcher               *fsnotify.Watcher
+	fileWatcherDone       chan struct{}
+	configMapWatcherDone  chan struct{}
+	configUpdateProcessor ConfigUpdateProcessor[T, PT]
+}
+
+func NewConfigWatcher[T any, PT util.DeepCopyable[T]](configPath string, configFileName string, namespace string, watchK8sConfigMap bool, configMapName string, clientset kubernetes.Interface, configUpdateProcessor ConfigUpdateProcessor[T, PT], logger log.FieldLogger) (*ConfigWatcher[T, PT], error) {
```
domsolutions: We could do with some unit tests for this, appreciate we're short on time so can be added at a later date.

scheduler/cmd/scheduler/main.go
```diff
@@ -278,20 +286,32 @@ func main() {
 	if err != nil {
 		logger.WithError(err).Fatal("Failed to load Kafka config")
 	}
-
-	numPartitions, err := strconv.Atoi(kafkaConfigMap.Topics["numPartitions"].(string))
+	scalingConfigHdl, err := scaling_config.NewScalingConfigHandler(scalingConfigPath, namespace, logger)
```
RobertSamoilescu, line 289: Looks like `NewScalingConfigHandler` can return nil. Shouldn't we need to account for that case. I assume the following logic won't be able to proceed if the `scalingConfigHdl` is nil.

scheduler/pkg/server/server.go
```diff
@@ -332,9 +341,60 @@ func NewSchedulerServer(
...
+
+				go func() {
+					// lock Mutex to avoid updating load balancer if a concurrent rebalance is in progress
+					c.modelEventStream.mu.Lock()
+					c.modelGwLoadBalancer.UpdatePartitions(newConfig.Pipelines.MaxShardCountMultiplier)
+					c.modelEventStream.mu.Unlock()
+					c.currentScalingConfig = newConfig
+					// rebalance all models onto available model-gw replicas according to new config
+					c.modelGwRebalance()
+				}()
```
RobertSamoilescu, line 411: is there  a risk here of out of order updates of the configs if another update arrives before the goroutine finish? I believe we need to wait for the goroutine to finish before starting handling another event. That being said, we can rebalance in parallel model-gw and pipeline-gw, but we have to wait for both of them to finish before processing another update.

scheduler/cmd/scheduler/main.go
lc525: Indeed, but if the handler is nil the error must be != nil, so things stop in the following if. Perhaps there's an argument to continue with a default, but then we're never tracking config updates and the setup will remain "frozen"

scheduler/pkg/config/watcher.go
lc525: Indeed. The config watcher is abstracted from the old implementation for AgentConfiguration, which had issues in being able to track changes in mounted ConfigMaps due to the use of symlinks. There are tests for the AgentConfiguration that can be repurposed, I'll see if I get some time now or add it briefly after the release.

scheduler/pkg/server/server.go
lc525: I'll add the wait because it's nicer. But in principle pipeline-gw and model-gw can concurrently work with different settings (the setting just influences how many "replicas" a pipeline or a model gets). And two rebalances of the same type (i.e two concurrent pipeline-gw rebalances can not proceed because of the locks.

operator/controllers/reconcilers/seldon/runtime_reconciler.go
lc525: You are obviously right, will fix. Thanks for catching this!

scheduler/pkg/config/watcher.go
lc525: Nope, will fix

scheduler/pkg/server/server.go
lc525: (offline discussion): understood the concern with out-of-order execution when processing multiple updates, you are correct. will fix

PR #6834, opened 2025-10-03
fix(operator):  unable to set 0 replicas on models/servers

domsolutions:
## Motivation

A scaling spec such as 

```yaml
replicas: 0
```

results in a validated scaling spec of 1 replica for models/servers, and in the case of models the scheduler will attempt to load the models.

## Summary of changes

- remove condition which only allows `replicas` = 0 if `minReplicas` == 0

## Checklist


lc525 (approved):
lgtm

PR #6841, opened 2025-10-06
fix(pipeline-gw): temp preStop hook 

MiguelAAe:
# Why
## Issues
When a pipeline pod is restarted, there is a race condition where the pod begins terminating before Envoy receives the updated endpoint list from Kubernetes. This results in Envoy routing traffic to a pod that is no longer accepting connections, causing intermittent 503 errors.

## Motivation
To prevent this issue, a preStop hook has been added to delay pod termination until the endpoint deregistration propagates to Envoy. This ensures Envoy stops routing traffic to the pod before it fully shuts down. This is a temporary measurement until integration with a more robust approach such as sending request to envoy xDS
# What
## Summary of changes

* PreStop hook  on pipeline gateway with a delay until xDS integration

## Checklist

### DIFF
--- a/k8s/kustomize/helm-components-sc/patch_pipelinegateway.yaml
+++ b/k8s/kustomize/helm-components-sc/patch_pipelinegateway.yaml
@@ -91,3 +91,10 @@ spec:
             value: '{{ hasKey .Values.pipelinegateway "logLevel" | ternary .Values.pipelinegateway.logLevel .Values.logging.logLevel }}'
           - name: PIPELINEGATEWAY_MAX_NUM_CONSUMERS
             value: '{{ .Values.pipelinegateway.maxNumConsumers }}'
+        lifecycle:
+          preStop:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - sleep 2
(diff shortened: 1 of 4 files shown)


k8s/kustomize/helm-components-sc/patch_pipelinegateway.yaml
```diff
@@ -91,3 +91,10 @@ spec:
             value: '{{ hasKey .Values.pipelinegateway "logLevel" | ternary .Values.pipelinegateway.logLevel .Values.logging.logLevel }}'
           - name: PIPELINEGATEWAY_MAX_NUM_CONSUMERS
             value: '{{ .Values.pipelinegateway.maxNumConsumers }}'
+        lifecycle:
```
MiguelAAe, line 94: Hi @lc525 I added the hook in this file and run the make file. However I can see the livenessprove is not in this file. Is this correct?

lc525: Instead, the liveness probe is defined in [operator/config/seldonconfigs/default.yaml](https://github.com/SeldonIO/seldon-core/blob/v2/operator/config/seldonconfigs/default.yaml). That requires running two sets of makefiles to get the change into the helm-charts and manifests

For now, I suggest we leave this pre-stop hook here (given it's temporary).

lc525 (approved):
lgtm as a temporary fix to 503s, will move to a similar strategy to draining servers in 2.11+.

PR #6849, opened 2025-10-06
fix(scheduler, dataflow): pipeline loading/unloading on pipeline-gw and dataflow engine topology

RobertSamoilescu:
# Why
## Issues

* The scheduler was sending messages to pipeline-gw replicas whenever `ModelsReady` changed. However, if the `PipelineGwStatus` was anything other than `Create` or `Delete`, the replicas would not acknowledge the message. As a result, the pipeline status never progressed.
* Sorting steps only by the sink is not always sufficient for the dataflow engine. In some cases, this led to inconsistent topologies, which caused sharding to fail when distributing the pipeline across multiple dataflow engine instances.

## Motivation

# What
## Checklist


lc525 (approved):
lgtm

PR #6853, opened 2025-10-07
fix(scheduler): Allow for pipelines with some of their statuses set to PipelineStatusUnknown

lc525:
When a new version of the scheduler (2.10) starts with the persistent store (on
PVC) from an old scheduler (<2.10), it loads the data about pipelines, and newly
added pipeline state fields are set to their default zero values.

In particular, `State.PipelineGwStatus` for all the existing pipelines in the
old cluster is set to `PipelineStatusUnknown`.

This was not considered as part of the "Running" statuses when deciding what
pipelines to consider as part of rebalancing towards pipeline-gw.


scheduler/pkg/store/pipeline/store.go
```diff
@@ -328,14 +328,19 @@ func (ps *PipelineStore) getAllRunningPipelineVersions(
...
-		switch statusSelector(pv) {
-		// we consider PipelineTerminating as running as it is still active
-		case PipelineCreate, PipelineCreating, PipelineReady, PipelineRebalancing, PipelineTerminating:
-			events = append(events, coordinator.PipelineEventMsg{
-				PipelineName:    pv.Name,
-				PipelineVersion: pv.Version,
-				UID:             pv.UID,
-			})
+		// never consider terminated pipelines, irrespective of the status selector
+		if pv.State.Status != PipelineTerminated {
```
lc525: Condition added so that we don't consider terminated pipelines when looking at `pv.State.PipelineGwStatus == PipelineStatusUnknown`

PR #6856, opened 2025-10-07
fix(scheduler): typo when setting pipeline-gw status for a pipeline 

MiguelAAe:
# Why
## Motivation

# What
## Checklist

### DIFF
--- a/scheduler/pkg/store/pipeline/store.go
+++ b/scheduler/pkg/store/pipeline/store.go
@@ -31,7 +31,7 @@ const (
 	removePipelineEventSource                  = "pipeline.store.removepipeline"
 	setStatusPipelineEventSource               = "pipeline.store.setstatus"
 	SetModelStatusPipelineEventSource          = "pipeline.store.setmodelstatus"
-	SetPipelineGwStatusPipelineEventSource     = "pipeline.store.setpipelinegwtatus"
+	SetPipelineGwStatusPipelineEventSource     = "pipeline.store.setpipelinegwstatus"
 	pipelineDbFolder                           = "pipelinedb"
 	modelEventHandlerName                      = "pipeline.store.models"
 )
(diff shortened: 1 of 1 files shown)


scheduler/pkg/store/pipeline/store.go
```diff
@@ -471,7 +471,7 @@ func (ps *PipelineStore) SetPipelineGwPipelineState(name string, versionNumber u
 	}
 	if ps.eventHub != nil {
 		for _, evt := range evts {
-			ps.eventHub.PublishPipelineEvent(setStatusPipelineEventSource, *evt)
+			ps.eventHub.PublishPipelineEvent(SetPipelineGwStatusPipelineEventSource, *evt)
```
lc525, line 474: Not so sure. I think this might actually need to be "source" (the argument passed to the function in the first place).

RobertSamoilescu, line 474: Could you please also fix the typo in the defintion: `pipeline.store.setpipelinegwtatus` -> `pipeline.store.setpipelinegwstatus`

lc525: Behaviour changes because [certain sources are ignored](https://github.com/lc525/seldon-core/blob/37aa06fbc26e43bbf109dc09ad63d027d6ee5643/scheduler/pkg/kafka/dataflow/server.go#L564)

RobertSamoilescu: Not really ... the `source` inside the `PipelineEventMsg` is different. That's a source for the eventbus which is logged when messages are sent over the channel. We can leave it as is and delete the unused constant. It won't make a difference

lc525: yep, realised in the meantime

lc525: Forgot we have a source for the event bus event and one in some events, which are passed via event.Data to the event hub.

lc525 (approved):
lgtm

PR #6882, opened 2025-10-13
fix(python-builder): [part 1] Replace `continuumio/miniconda3` with `condaforge/miniforge3` + upgrade to Python 3.12

tyndria:
# What
## Summary of changes

- Upgraded python, miniforge and poetry to support Python 3.12
- Replaced default channel image with the community-managed image.
- Remove explicit `git` installation as it's already a part of the used image

## Checklist

## Testing

- Run lint command in the updated container  ✅ 
- Run tests from `/python` folder in the updated container  ✅ 
- Pushed python package to the pypi test index from the `python-builder` container - ✅ 

## Next steps
- [ ] Push the new image (`seldonio/python-builder:0.9` tag) to the `seldonio` hub, after push CI won't fail

### DIFF
--- a/.github/workflows/python_lint.yml
+++ b/.github/workflows/python_lint.yml
@@ -2,10 +2,11 @@ name: V1 Python Lint
 
 on:
   push:
-    branches: [ master ]
+    # TODO revert before merge to master
+    branches: '*'
   pull_request:
     # TODO revert before merge to master
-    branches: [ fix/core-1-CVEs ]
+    branches: '*'
 
 jobs:
   python-lint:


python-builder/Dockerfile
```diff
@@ -21,35 +20,24 @@ RUN INSTALL_DIR=/tmp/flatc-install && \
...
 
 # Install python dependencies
-RUN pip install --upgrade pip setuptools wheel && \
-    pip install grpcio grpcio-tools flatbuffers twine mypy-protobuf
+RUN pip3 install --upgrade pip setuptools wheel && \
+    pip3 install grpcio grpcio-tools flatbuffers twine mypy-protobuf
 
 # Installs required
-RUN apt-get install -y curl
+RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
```
tyndria, line 39: I believe it's better to move to the `pipx` installation of poetry after we upgrade python and the base `condaforge/miniforge3` image, so we would get rid of this line
cc @vtaskow

python-builder/Dockerfile
```diff
@@ -21,35 +20,24 @@ RUN INSTALL_DIR=/tmp/flatc-install && \
...
-    wget https://github.com/git/git/archive/v2.30.0.zip -O git.zip && \
-    unzip git.zip && \
-    apt-get remove git -y && \
-    make prefix=/usr/local -C git-2.30.0/ all install
-
-
 # This is to install desired version of Python without updating conda version
-ENV PYTHON_VERSION "3.7.10"
-ENV CONDA_VERSION "4.7.12"
+ENV PYTHON_VERSION="3.7.10"
```
tyndria: We would need to upgrade conda and python here after tests are upgraded
cc @vtaskow

python-builder/Dockerfile
```diff
@@ -1,12 +1,11 @@
-FROM continuumio/miniconda3:4.12.0
+FROM condaforge/miniforge3:4.12.0-2
```
tyndria: We would need to upgrade this image after python tests are upgraded
cc @vtaskow

.github/workflows/python_tests.yml
```diff
@@ -2,7 +2,6 @@ name: V1 Python Tests
 
 on:
   push:
-    branches: [ master ]
```
tyndria, line 5: This is for testing purposes only (to test workflow after we push the image)

.github/workflows/python_tests.yml
```diff
@@ -2,7 +2,6 @@ name: V1 Python Tests
...
   pull_request:
     # TODO revert before merge to master
     branches: [ fix/core-1-CVEs ]
```
tyndria: For testing purposes only (revert)

PR #6885, opened 2025-10-16
fix(scheduler): not all models deployed to Servers when minReplicas on Model is set

domsolutions:
## Issues

When we have a Model with `minReplicas` set:

```yaml
kind: Model
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"mlops.seldon.io/v1alpha1","kind":"Model","metadata":{"annotations":{},"name":"iris","namespace":"seldon-mesh"},"spec":{"requirements":["sklearn","mlserver"],"storageUri":"gs://seldon-models/scv2/samples/mlserver_1.3.5/iris-sklearn"}}
  creationTimestamp: "2025-10-15T18:03:07Z"
  finalizers:
  - seldon.model.finalizer
  generation: 10
  name: iris
  namespace: seldon-mesh
  resourceVersion: "86482"
  uid: 3a8e03cb-798c-4b55-baee-08b8d76e3bda
spec:
  minReplicas: 2
  replicas: 5
  requirements:
  - sklearn
  - mlserver
  storageUri: gs://seldon-models/scv2/samples/mlserver_1.3.5/iris-sklearn

```

It will successfully deploy 5 models to the Servers. But if we then restart the Server pods, the model will be loaded > `minReplicas` but < `replicas` with this number varying each time.

## Summary of changes

- Only allow one goroutine to trigger scheduling of failed models, as if another agent has also restarted, it will call the memory store to gather the failed models, but if another agent is also restarting, the model state changes can interfere with scheduling decisions
- Also consider models as failed if they are in `Progressing` state but the desired replica number has not been reached as this is a valid scenario where we should attempt to schedule models.
- Introduced Delve to be used within Kind. This is still at experimental stage and tends to crash, will need further improvements.

## Checklist

### DIFF
--- a/scheduler/pkg/scheduler/scheduler.go
+++ b/scheduler/pkg/scheduler/scheduler.go
@@ -81,20 +81,23 @@ func (s *SimpleScheduler) Schedule(modelKey string) error {
 }
 
 func (s *SimpleScheduler) ScheduleFailedModels() ([]string, error) {
-	s.muScheduleFailed.Lock()
-	defer s.muScheduleFailed.Unlock()
-
 	if !s.synchroniser.IsReady() {
 		s.logger.Debug("Waiting for servers to connect")
 		s.synchroniser.WaitReady()
+		s.logger.Debug("Waiting for servers complete")
 	}
+
+	s.muScheduleFailed.Lock()
+	defer s.muScheduleFailed.Unlock()
 
 	failedModels, err := s.getFailedModels()
 	if err != nil {
 		return nil, err
 	}
 
-	s.logger.WithField("failed_models", failedModels).Debug("Got failed models to schedule")
+	if len(failedModels) > 0 {
+		s.logger.WithField("failed_models", failedModels).Debug("Got failed models to schedule")
+	}
 
 	var updatedModels []string
 	for _, modelName := range failedModels {
--- a/scheduler/pkg/store/memory_status.go
+++ b/scheduler/pkg/store/memory_status.go
@@ -141,8 +141,6 @@ func (m *MemoryStore) FailedScheduling(modelID string, version uint32, reason st
 				modelVersion.SetServer("")
 			}
 
-			model.versions[i] = modelVersion
-
 			m.eventHub.PublishModelEvent(
 				modelFailureEventSource,
 				coordinator.ModelEventMsg{


scheduler/pkg/agent/server.go
```diff
@@ -419,29 +420,31 @@ func (s *Server) Subscribe(request *pb.AgentSubscribeRequest, stream pb.AgentSer
...
-			return nil
-		case <-ctx.Done():
-			logger.Infof("Client replica %s:%d has disconnected", request.ServerName, request.ReplicaIdx)
-			s.mutex.Lock()
-			delete(s.agents, key)
-			s.mutex.Unlock()
-			s.removeServerReplicaImpl(request.GetServerName(), int(request.GetReplicaIdx())) // this is non-blocking beyond rescheduling models on removed server
-			return nil
-		}
+	select {
```
domsolutions, line 435: No-op, just for readability, there was no need for the loop

ansible/README.md
```diff
@@ -241,3 +241,25 @@ rm -rf ~/.cache/seldon/
 
 If you used Ansible to install Seldon Core 2 and its ecosystem into K8s cluster other than Kind you need to manually
 remove all the components.
+
```
lc525: Let's move this to README.dev.md

scheduler/pkg/scheduler/scheduler.go
```diff
@@ -80,19 +81,29 @@ func (s *SimpleScheduler) Schedule(modelKey string) error {
...
+		s.logger.Debug("Waiting for servers to connect")
+		s.synchroniser.WaitReady()
+	}
+
 	failedModels, err := s.getFailedModels()
 	if err != nil {
 		return nil, err
 	}
+
+	s.logger.WithField("failed_models", failedModels).Debug("Got failed models to schedule")
```
lc525: The message of this log is only true if `len(failedModels)`>0

scheduler/pkg/scheduler/scheduler.go
```diff
@@ -80,19 +81,29 @@ func (s *SimpleScheduler) Schedule(modelKey string) error {
...
 
 func (s *SimpleScheduler) ScheduleFailedModels() ([]string, error) {
-	s.synchroniser.WaitReady()
+	s.muScheduleFailed.Lock()
+	defer s.muScheduleFailed.Unlock()
+
+	if !s.synchroniser.IsReady() {
+		s.logger.Debug("Waiting for servers to connect")
+		s.synchroniser.WaitReady()
+	}
```
lc525: Could we switch the order here, with first waiting on the synchronizer (multiple agents can wait in parallel) and only then taking the lock? The way it is now, correctness depends on the synchronizer being signalled before `ScheduleFailedModels` is called. Otherwise, this may become a deadlock, depending on the implementation of the synchronizer and the order in which things are called (i.e. the synchronizer only becomes ready after the timeout because the other agents are waiting on the `muSchedulerFailed` lock). I believe this works right now, but is just more fragile than I would like.

scheduler/pkg/scheduler/scheduler_test.go
```diff
@@ -755,3 +764,315 @@ func TestRemoveAllVersions(t *testing.T) {
...
+//
+//	// Additional setup for replicas if needed
+//	modelStatus := store.ModelStatus{
+//		State:             state,
+//		AvailableReplicas: uint32(availableReplicas),
+//	}
+//	mv.SetModelState(modelStatus)
+//
+//	return mv
+//}
```
lc525: Can those be removed, or are they here for future tests?

scheduler/pkg/store/memory_status.go
```diff
@@ -141,6 +141,8 @@ func (m *MemoryStore) FailedScheduling(modelID string, version uint32, reason st
 				modelVersion.SetServer("")
 			}
 
+			model.versions[i] = modelVersion
```
lc525: Is this why perhaps the `ScheduleFailed` status was not being propagated when you had removed
the condition on `okWithMinReplicas`?

domsolutions: I put this is when I had removed the condition on `okWithMinReplicas` as I thought maybe `modelVersion := model.versions[i]` was making a copy instead of referring to version by pointer and that's why the state wasn't reflected as failed. But that wasn't the case, so I'll remove it.

PR #7088, opened 2026-01-13
fix scheduler test

MiguelAAe:
# Why
## Motivation

# What
## Checklist


scheduler/pkg/scheduler/scheduler_test.go
```diff
@@ -578,9 +565,12 @@ func TestScheduler(t *testing.T) {
...
 			)
 
-			mockStore := newMockStore(test.model, test.servers)
-			scheduler := NewSimpleScheduler(logger, mockStore, DefaultSchedulerConfig(mockStore), synchroniser.NewSimpleSynchroniser(time.Duration(10*time.Millisecond)), eventHub)
-			err := scheduler.Schedule(test.model.Name)
+			//mockStore := newMockStore(test.model, test.servers)
+			scheduler := NewSimpleScheduler(logger, mockModelServerAPI, DefaultSchedulerConfig(mockModelServerAPI), synchroniser.NewSimpleSynchroniser(time.Duration(10*time.Millisecond)), eventHub)
+			err := scheduler.Schedule(test.modelName)
+			//model, err := scheduler.store.GetModel(test.modelName) //todo?
+			//model.GetLastAvailableModel().Replicas //todo?
```
domsolutions: I think it's fine to not do this as it's not an integration test and the model wouldn't actually be updated right as we're just using mocks?

